Issue #5, opened 2019-09-12
App crashes on using the "Change Start Point" filter


kiozen:
**Describe the bug**
When using the "Change Start Point" filter on a track QMapShack crashes instantly

**To Reproduce**
1. Create a track with attached waypoints
2. Edit track
3. Select Filter->Miscellaneous->Change Start Point
4. Select a waypoint
5. Press button with gear to activate filter

QMapShack crashes.

**Expected behavior**
QMapShack is reordering the track

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - OS: OS X
 - QMapShack 1.13.1

**Additional context**
n/a

kiozen:
see PR #10

PR #10, opened 2019-09-13
[QMS-5] App crashes on using the "Change Start Point" filter

kiozen:
Fixes Issue #5 

Reordering seems not to work on OS X. Fixed it by splitting the track
into two parts and combining these parts in reverse order.


kiozen:
Changelog entry is missing

Issue #13, opened 2019-09-16
Add support for Garmin Edge 500


kiozen:
**Is your feature request related to a problem? Please describe.**
The Edge 500 does not take GPX as data input. It uses TCX files.

**Describe the solution you'd like**
Detect the Edge 500 via GarminDevice.xml and export projects as TCX instead of GPX

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

PR #24, opened 2019-09-24
[QMS-13] Add support for Garmin Edge 500

kiozen:
**What is the linked issue for this pull request (start with a `#`):** #13 

* Add trimmed() to all text readout as there can be leading and trailing spaces
* Decode section `FitnessCourses` and store path as TCX target path
* Read projects from TCX target path
* Detect Edge 5xx and save data as TCX
* Fix testForExternalProject() by mounting the device

* Attach a device or a USB stick with the correct GarminDevice.xml 
* TCX files on the device should be listed
* Send a project to the device -> should be listed
* Restart QMS -> project is listed on the device
* Send project again -> project on the device is replaced.

Does the code comply to the coding rules and naming conventions [Coding Guidleines](https://github.com/Maproom/qmapshack/wiki/DeveloperCodingGuideline):

- [x] yes

Is every user facing string in a tr() macro?

- [x] yes

Is the change user facing?

- [x] yes,
- [ ] no

If there are user facing changes did you add the ticket number and title into the changelog?

- [x] yes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+CMakeLists.txt.user
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 V1.XX.X
 [QMS-5] App crashes on using the "Change Start Point" filter
 [QMS-12] Unify versions of all QMapShack tools
-
+[QMS-13] Add support for Garmin Edge 500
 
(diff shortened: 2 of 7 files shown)


kiozen:
Tested with USB stick and device.

Issue #18, opened 2019-09-18
Mistake in german translation


pingurus:
**Describe the bug**
In the search help window following is written:
> 'Datum gleich' findet alles, was innerhalb von 24 Stunden des aktuellen Tages liegt.

It should be "innerhalb des angegebenen Tages". 

**To Reproduce**
1. Open Help Window for the workspace search
2. Scroll down to (see above)

sikmir:
@kiozen By the way, what about enabling some web-based localization tool for qmapshack (e.g. [Weblate](https://hosted.weblate.org/))? That would help to translators a lot.

wthaem:
Wrong translation is due to wrong English text in code:

<source>&apos;Date equals&apos; matches everything that is within 24h of the current day.</source>

The English text should be changed to something like "matches everything within the given day" (not "current" - that's different!)

pingurus:
Sorry, my fault!

kiozen:
@pingurus can you fix code and German translation? 

For all: reviewing that part one more time doesn't hurt.

kiozen:
And keep in mind that things are a bit more complicated now:

https://github.com/Maproom/qmapshack/wiki/DeveloperCommitCode

PR #25, opened 2019-09-24
[QMS-18] Improved explanation

pingurus:
**What is the linked issue for this pull request (start with a `#`):** #18 

Improved the explanation of what 'date equals'
actually does. However I'm not sure if it really
more understandable.

1. Open help window for workspace search and scroll down

- [x] yes

- [x] yes

- [x] yes,
- [ ] no

Should this also happen for this minor change?
- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,5 @@
 V1.XX.X
+[QMS-18] Improved explanation of 'Date equals'
 [QMS-5] App crashes on using the "Change Start Point" filter
 [QMS-12] Unify versions of all QMapShack tools
 
--- a/src/qmapshack/gis/search/CSearchExplanationDialog.cpp
+++ b/src/qmapshack/gis/search/CSearchExplanationDialog.cpp
@@ -54,7 +54,7 @@ CSearchExplanationDialog::CSearchExplanationDialog(QWidget *parent)
     explanation += "</ul>";
     explanation += "</li>";
     explanation += "<li>";
-    explanation += tr(  "'Date equals' matches everything that is within 24h of the current day.");
+    explanation += tr(  "'Date equals' matches everything that is within the next 24h, if time is provided with date, if not, then everything on the day queried. Be aware that that the times are converted to UTC before comparison and you local time zone is taken for the query.");
     explanation += "</li>";
     explanation += "<li>";
     explanation += tr(  "If you enter no unit the default unit (what you see when viewing the property of the item) is used.");
--- a/src/qmapshack/locale/qmapshack.ts
+++ b/src/qmapshack/locale/qmapshack.ts
@@ -1948,18 +1948,18 @@ You can modify both parameters to customize the behavior of the filter. For the 
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../gis/trk/filter/filter.cpp" line="579"/>
+        <location filename="../gis/trk/filter/filter.cpp" line="582"/>
         <source>Start Point moved to: </source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../gis/trk/filter/filter.cpp" line="629"/>
-        <location filename="../gis/trk/filter/filter.cpp" line="643"/>
+        <location filename="../gis/trk/filter/filter.cpp" line="632"/>
+        <location filename="../gis/trk/filter/filter.cpp" line="646"/>
         <source>%1 (Part %2)</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../gis/trk/filter/filter.cpp" line="770"/>
+        <location filename="../gis/trk/filter/filter.cpp" line="773"/>
         <source>Hide zero speed drift knots with a distance criteria of (%1%2) and ratio of (%3)</source>
         <translation type="unfinished"></translation>
     </message>
@@ -6605,7 +6605,7 @@ For configuration you need to know your Android device&apos;s IP address or it&a
     </message>
     <message>
         <location filename="../gis/search/CSearchExplanationDialog.cpp" line="57"/>
-        <source>&apos;Date equals&apos; matches everything that is within 24h of the current day.</source>
+        <source>&apos;Date equals&apos; matches everything that is within the next 24h, if time is provided with date, if not, then everything on the day queried. Be aware that that the times are converted to UTC before comparison and you local time zone is taken for the query.</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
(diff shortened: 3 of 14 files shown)


wthaem:
In my understanding the string

`"'Date equals' matches everything that is within 24h of the time provided, thus everything on the day queried.")`

is still misleading. "Within 24 hours" and "on the day queried" define different time intervals.

*Example:* 
* `Date = 30.08.2019 10:00`
* *Within 24 hours:* (in my understanding) (29.08.2019 10:00, 31.08.2019 10:00) (difference between queried datetime and found datetime < 24 h)
* *Day queried:* `30.08.2019 xx:xx`

Why not simply saying

`"'Date equals' means everything on the day queried" `

pingurus:
> Example:
>    Date = 30.08.2019 10:00
>    Within 24 hours: (in my understanding) (29.08.2019 10:00, 31.08.2019 10:00) (difference between queried datetime and found datetime < 24 h)
>    Day queried: 30.08.2019 xx:xx

In code, what actually happens is that everything is returned what lies in this interval: 29.08.2019 10:00 - 31.08.2019 10:00. However, if you just query for `date equals 29.08.2019`, the interval is 29.08.2019 00:00 - 31.08.2019 00:00

But we can just ommit the fact that it not only works with dates, but also with datetimes and write what you suggested.

kiozen:
Please update the change log, too. This is a user facing change.

And can you please fix the German translation, too?

wthaem:
New peoposal after this clarifying explanation:

` "'Date equals' matches everything that is within 24h, if time is provided with date, if not, then everything on the day queried.");`

pingurus:
I will take that then.

pingurus:
> And can you please fix the German translation, too?
I am not sure how to do that. If I open the qmapshack_de.ts file, on the right it shows the new explanation, however in the list of translations still the old explanation appears.

kiozen:
> > And can you please fix the German translation, too?
> > I am not sure how to do that. If I open the qmapshack_de.ts file, on the right it shows the new explanation, however in the list of translations still the old explanation appears.

Use ccmake to set `UPDATE_TRANSLATIONS` to true. Compile. This will update the language files. Next use linguist to edit the German translation.

wthaem:
There still seems to be a misunderstanding in the help string or an error in the code.

I took a waypoint with Date/time 2019-09-25T12:24:18+02:00 (see attached snapshot).

* *Search string:* `date = 25.09.2019 11:00` - waypoint found
* *Search string:* `date = 25.09.2019 13:00` - waypoint not found. With the 24-hour rule it should found.

BTW: how is the interpration in case of a track? 

WPT

pingurus:
> There still seems to be a misunderstanding in the help string or an error in the code.

This is indeed an error in the code, I didn't think of time zones when coding. It is excluded, since the UTC time of the waypoint is before the queried time.

> BTW: how is the interpration in case of a track?

The exclusivity rule applies. So it is impossible to find a multi-day track with `date equals`. Maybe we should consider changing from exclusive to inclusive in general to fix this. One could then add exclusive operators, so both modes are available.

wthaem:
Rule described above:

`'Date equals' matches everything that is within 24h, if time is provided with date ...` 

If there is a timezone or not doesn't change the picture: With or without timezone the waypoint time and the query time differ by less than 24 hours!

Thus, the phrase

*It is excluded, since the UTC time of the waypoint is before the queried time.* 

contradicts the 24 h rule and the word  *before* looks a bit strange to me because the proper query is in the given case of the form *find all data objects with datetime belonging to a given time interval* and this phrase doesn't include *before*.

WRT tracks: a track has many trackpoints with timestamps. What isn't obvious to the user is what kind of track timestamp is used when carrying out the query (*all timestamps, 1 timestamp, min/max timestamp, ...*)

pingurus:
> If there is a timezone or not doesn't change the picture: With or without timezone the waypoint time and the query time differ by less than 24 hours!

I see what you mean. However it is 24h starting from the datetime given, since entering a date only equals to entering the date with the time 00:00. So we should change it to `"'Date equals' matches everything that is within` **the next** `24h, if time is provided with date, if not, then everything on the day queried."`

> the word before looks a bit strange to me because the proper query is in the given case of the form find all data objects with datetime belonging to a given time interval and this phrase doesn't include before.

I'm not sure what you want to say, sorry.

>WRT tracks: a track has many trackpoints with timestamps. What isn't obvious to the user is what kind of track timestamp is used when carrying out the query (all timestamps, 1 timestamp, min/max timestamp, ...)

Both the start and the end timestamp currently have to be in the 24h interval. (which is the exclusivity rule)

wthaem:
Again one step clearer! Thanks! Up to now the 24 h rule could be interpreted also in the form: *from 24 h before to 24 h after the given datetime* (absolute value of difference between 2 datetimes < 24 h) . And with the new formulation the *before* is understandable to me.

Thus, use your last formulation. 

And once the topic *timezone* was raised: How are datetimes handled if now explicit timezone is given? Local timezone of the OS?

And thanks for the track clarification!

pingurus:
> And once the topic _timezone_ was raised: How are datetimes handled if now explicit timezone is given? Local timezone of the OS?

Yes, the local timezone is used.

pingurus:
My pull request is now in three commits, since I ammended a commit that I already pushed and forgot about the --force option. Should/Can I fix that by a rebase or would that mix things up even more?

kiozen:
> My pull request is now in three commits, since I ammended a commit that I already pushed and forgot about the --force option. Should/Can I fix that by a rebase or would that mix things up even more?

As long as you are on your branch it's ok to do a force push. If you are not sure you can do a `--force-with-lease`. This will warn you if you destroy changes by others with your push.

Of course you can and should fix the commit history as good as you can. For example to re-edit the last three commits you can do `git rebase -i HEAD~3` Take the instructions from git to merge commits into other commits, to re-arrange the order of commits or to rephrase wording. You can also revert the last commits to start properly from the beginning: `git reset --soft HEAD~3`. Note the `--soft` This is important. This will keep your changes in the files.

kiozen:
@wthaem are all your concerns addressed?

kiozen:
> Ooops. Now the commit history looks quite bit reduced.

And add ticket number to each commit header like:

```
[QMS-18] Change blabla

I changed foo bar.
```

This will help to find all changes related to a ticket

pingurus:
> 
> 
> Ooops. Now the commit history looks quite bit reduced.

Well, basically I did not do much :laughing: 
So what should I do?

wthaem:
I hope everything is clear now. Sorry for the delayed nswer.

Issue #35, opened 2019-10-02
Display of predicted walking time


pingurus:
**Is your feature request related to a problem? Please describe.**
Sometimes when I plan hikes and I have a specific timeframe it would be nice if QMS displayed the time needed for track.

**Describe the solution you'd like**
[Here](https://de.wikipedia.org/wiki/Marschzeitberechnung) (sorry, only german) you'll find the formula of DIN 33466 according to which the times on the hiking signs in the alps are labelled.

The calculation base is that in one hour you can cover:
* 300m of elevation gain in ascent
* 500m of elevation gain in descent
* 4km distance in flat areas.

First, the times needed for the elevation and distance are calculated seperately. Then the smaller one of the two is added to the bigger one to get the total time.

Example:

- Distance: 23km Elevation gain: 1025m Elevation loss: 1725m
- Time needed for distance: 23km/4km/h = 5,75h
- Time needed for elevation: 1025m/300m/h + 1725m/500m/h = 6,86h
- Nominal time needed: 6,86h+5,75h/2=9,74h

As a comparison we needed 9,15h for that hike, so this is a rather conservative approach, just as said in the wikipedia article.

**Describe alternatives you've considered**
Alternatively one could also use [this](https://de.wikipedia.org/wiki/Leistungskilometer) (again, only german) concept, however I prefer the first one since it outputs times.

**Additional context**
Of course, this calculation should be only triggered if the profile is set to 'foot'

kiozen:
You are aware of the speed filter?

https://github.com/Maproom/qmapshack/wiki/AdvTrkFilters#hiking

pingurus:
No, I was not.
I think we should still keep it, as it does not tamper with the actual track data. This way one can get a feeling how fast one is compared to the signs, as it is also calculated for recorded tracks.
But I understand if you think my use-case is already covered by the filter.

pingurus:
Whilst playing around with the filter, I was thinking wether it would be possible to give the user the possibility to automatically import track data into different profiles. These profiles would average the data fed into them depending on incline and distance until this point to create a profile corresponding to the users capabilities. Then, instead of the formula used by the filter or by the DIN, the predicted time and the filter results could be based on this generated profile.
However, I'm not sure if this is really needed/wanted.
So, do you think this is an overkill?
Do you think it is useful?
[...]

PR #36, opened 2019-10-02
[QMS-35] Display of predicted walking time

pingurus:
Added walking time according to DIN 33466

**What is the linked issue for this pull request (start with a `#`):** #35

I added the walking time to the getInfo method. It is only calculated when elevation data is present and the activity is foot.

1. Open any project with tracks with the activity set to foot
2. Open description or double click track

- [x] yes

- [x] yes

- [x] yes,
- [ ] no

- [x] yes

### DIFF
--- a/src/qmapshack/gis/trk/CGisItemTrk.cpp
+++ b/src/qmapshack/gis/trk/CGisItemTrk.cpp
@@ -538,6 +538,15 @@ QString CGisItemTrk::getInfo(quint32 feature) const
         str += "<br />";
     }
 
+    //nominal walking time according to DIN 33466
+    CActivityTrk::summary_t walking = activities.getSummary(CTrackData::trkpt_t::eAct20Foot);
+    qreal elevationTime = walking.ascent/300 * 3600 + walking.descent/500 * 3600;
+    qreal distanceTime = walking.distance/4000 * 3600;
+    qreal nominalTime = qMax(elevationTime, distanceTime) + 0.5 * qMin(elevationTime, distanceTime);
+    IUnit::self().seconds2time(nominalTime, val1, unit1);
+    str += tr("Est. Walk. Time: %1%2").arg(val1).arg(unit1);
+
+    str += "<br/>";
     str += "<br/>";
 
     if(timeIsValid && timeStart.isValid())
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,5 @@
 V1.XX.X
+[QMS-35] Add nominal walking time to track info
 [QMS-27] Error in workspace search for attributes
 [QMS-18] Improved explanation of 'Date equals'
 [QMS-5] App crashes on using the "Change Start Point" filter
--- a/src/qmapshack/gis/trk/CActivityTrk.h
+++ b/src/qmapshack/gis/trk/CActivityTrk.h
@@ -181,7 +181,9 @@ public:
         return activityRanges;
     }
 
-
+    const summary_t getSummary(trkact_t act) const{
+        return activitySummary.value(act);
+    }
 private:
     friend class CGisItemTrk;
     CActivityTrk(CGisItemTrk * trk);


src/qmapshack/gis/trk/CGisItemTrk.cpp
```diff
@@ -506,6 +506,17 @@ QString CGisItemTrk::getInfo(quint32 feature) const
             IUnit::self().meter2elevation(totalDescent, val2, unit2);
 
             str += tr(", %1%2%3, %4%5%6").arg(QChar(0x2197)).arg(val1).arg(unit1).arg(QChar(0x2198)).arg(val2).arg(unit2);
+
+            if(actCnt == 1 && acts.contains(CTrackData::trkpt_t::eAct20Foot))
+            {
+                str += "<br />";
+                //ominal walking time according to DIN 33466
+                qreal elevationTime = totalAscent/300 * 60 + totalDescent/500 * 60;
+                qreal distanceTime = totalDistance/4000 * 60;
```
kiozen: An hour has 3600 seconds ;)

src/qmapshack/gis/trk/CGisItemTrk.cpp
```diff
@@ -506,6 +506,17 @@ QString CGisItemTrk::getInfo(quint32 feature) const
             IUnit::self().meter2elevation(totalDescent, val2, unit2);
 
             str += tr(", %1%2%3, %4%5%6").arg(QChar(0x2197)).arg(val1).arg(unit1).arg(QChar(0x2198)).arg(val2).arg(unit2);
+
+            if(actCnt == 1 && acts.contains(CTrackData::trkpt_t::eAct20Foot))
+            {
+                str += "<br />";
+                //ominal walking time according to DIN 33466
```
kiozen: typo `ominal`

src/qmapshack/gis/trk/CGisItemTrk.cpp
```diff
@@ -506,6 +506,17 @@ QString CGisItemTrk::getInfo(quint32 feature) const
             IUnit::self().meter2elevation(totalDescent, val2, unit2);
 
             str += tr(", %1%2%3, %4%5%6").arg(QChar(0x2197)).arg(val1).arg(unit1).arg(QChar(0x2198)).arg(val2).arg(unit2);
+
+            if(actCnt == 1 && acts.contains(CTrackData::trkpt_t::eAct20Foot))
```
kiozen: This shouldn't be limited to tracks with a single activity. There is also a track statistic per activity.

src/qmapshack/gis/trk/CGisItemTrk.cpp
```diff
@@ -506,6 +506,17 @@ QString CGisItemTrk::getInfo(quint32 feature) const
...
+            {
+                str += "<br />";
+                //ominal walking time according to DIN 33466
+                qreal elevationTime = totalAscent/300 * 60 + totalDescent/500 * 60;
+                qreal distanceTime = totalDistance/4000 * 60;
+                qreal nominalTime = qMax(elevationTime, distanceTime) + 0.5 * qMin(elevationTime, distanceTime);
+                IUnit::self().seconds2time(nominalTime, val1, unit1);
+                str += tr("Nom. Time: %1%2").arg(val1).arg(unit1);
```
kiozen: I doubt that any user will understand what `Nom. Time` is about.

src/qmapshack/gis/trk/CGisItemTrk.cpp
pingurus: :man_facepalming:

src/qmapshack/gis/trk/CGisItemTrk.cpp
pingurus: Yes, I doubt that too.
However, I wanted to shorten it, so it wouldn't take up that much space in the info box.
Any other suggestion?

src/qmapshack/gis/trk/CGisItemTrk.cpp
pingurus: So how should I handle it so it is only calculated for the walking part?

kiozen: Add a method to `CActivityTrk` to get the `summary_t` structure of an activity.

src/qmapshack/gis/trk/CGisItemTrk.cpp
kiozen: `est. walk. time` An move it below the total time and moving time.

src/qmapshack/gis/trk/CGisItemTrk.cpp
```diff
@@ -506,6 +506,9 @@ QString CGisItemTrk::getInfo(quint32 feature) const
             IUnit::self().meter2elevation(totalDescent, val2, unit2);
 
             str += tr(", %1%2%3, %4%5%6").arg(QChar(0x2197)).arg(val1).arg(unit1).arg(QChar(0x2198)).arg(val2).arg(unit2);
+
+
+
```
kiozen: Has there been an offer for empty lines? :wink:

src/qmapshack/gis/trk/CGisItemTrk.cpp
```diff
@@ -538,6 +541,15 @@ QString CGisItemTrk::getInfo(quint32 feature) const
         str += "<br />";
     }
 
+    //nominal walking time according to DIN 33466
+    CActivityTrk::summary_t walking = activities.getSummary(CTrackData::trkpt_t::eAct20Foot);
+    qreal elevationTime = walking.ascent/300 * 3600 + walking.descent/500 * 3600;
+    qreal distanceTime = walking.distance/4000 * 3600;
+    qreal nominalTime = qMax(elevationTime, distanceTime) + 0.5 * qMin(elevationTime, distanceTime);
+    IUnit::self().seconds2time(nominalTime, val1, unit1);
+    str += tr("est. walk. time: %1%2").arg(val1).arg(unit1);
```
kiozen: Sorry for the fuzz, but the rest of the strings is with leading capital, too. Didn't consider this when suggesting the string. So it's `Est. Walk. Time.` or `Est. walk. time` whatever you feel more natural.

kiozen:
Use 
```sh
git rebase -i HEAD~2
```
to clean up your commit history. You can use `git log` or `tig` to check if your log history fits into the usual pattern.

mitxel-m:
Hi all

I have been testing the PR. Pingurus work is OK, all works as indicated in the PR, but I see a real problem with the DIN method itself.

It's obvious that the estimated time is inflated, but no problem with this. One may think that it is to have a wide security margin and I will not argue about it.

The important thing is that the method of calculation is not consistent for all cases. Take a track and divide it into two parts, the sum of the times of the parts does not match with the time of the whole track.

In many cases the error is acceptable, but in some others it is quite considerable.

**Use case.**

DIN-test1

Take the whole track and check the estimation, then divide it at a B waypoint, and we get: 

*Data copied from QMS as shown. Remark that distances are rounded by QMS*

|Track|Dist. |↗|↘| Est. Walk. Time: |
|---------------- |------- |------ |-------| ----------|
| Total Track ABC | 19.4km | ↗620m | ↘275m |  06:09:10h |
| First part  AB  |  2.0km | ↗405m | ↘0m   |  01:36:00h |
| Second part BC: | 17.4km | ↗215m | ↘275m |  04:59:46h |


The sum of the parts is increased about 26 min. Doing it separately takes more time than doing it in one trip . This doesn't make sense. :-/

I assume that DIN calculation method is valid as a thumb rule easy to use when you are already in the mountains. In fact many of us (me included) have used a similar old school rule on some occasion, perhaps with other parameters, but similar method. You get the feeling that works well because the margin is very wide, and you always arrive before the calculated, and the same logic applies to signposts.

But times changes,  :-) and now we have a better tool for planning. One of the best things we have at QMS is the time filter for hiking. (.Maybe it could be improved), From my experience, when you adjust the parameters well it fits to real time within reasonable margins. And of course the result does not vary if you cut the track. The same calculation is reproducible under the same conditions, so if you print the project summary you get a nice roadbook with all partial times and distances calculated consistently. 

You can even enter in this filter the values of flat speed, ascent and descent recommended by DIN, DAV, OASV, SAC or whatever you want and the result will be consistent throughout the track. :-)

I am aware that the proposed DIN calculation don't touch the track register, but take in mind that the time showed with the track general data can be a calculated time if you are plannig a route.  

Honestly, I do not find a good idea to show the DIN rough estimation just next to the time that you've carefully calculated with the time filter...Thay are very different values, and I'm afraid that introducing that element  will lead to confusion,

I appreciate Pingurus' work, but in this case I think the approach is not good.

pingurus:
I see the issue @mitxel-m has with this PR, maybe we should have a discussion back at the issue about how the estimation value should be calculated, since I do believe that it is useful to have it displayed without changing the actual track data.
Besides that, I think we should wait until the enhanced activity system is implemented, since that will affect the time filters and could make this more useful when implemented in a general way for all activities.

Issue #37, opened 2019-10-02
Inconsistent use of time zones


pingurus:
**Describe the bug**
In my case QMS was set to UTC time zone in the settings, however the filter tab used my local time zone. So, when I manually adjusted the beginning of a track, what I saw in the track info did not match to what I set.

**To Reproduce**
1. Set time zone to whatever zone is not yours
2. Open track
3. Go to Filter>Timestamps>Change Time
4. Set some time
5. The time shown to the left in the info is not what you set

**Expected behavior**
I expect it to be either the same or the user being able to choose the time zone in the filter.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/36167579/66052792-788a5280-e531-11e9-89c4-1afc7f75b0ac.png)

**Desktop (please complete the following information):**
 - OS: Windows 8.1
 - QMapShack Version 1.13.2

kiozen:
should I assign it to you?

pingurus:
If it is a quick fix for you I'd prefer you do it, otherwise I'll have a look.

pingurus:
The error arises since in [CFilterNewDate.cpp](https://github.com/Maproom/qmapshack/blob/dev/src/qmapshack/gis/trk/filter/CFilterNewDate.cpp), the manipulation of DateTimes is through QDateTime and not through IUnit.

I can fix that, however I am unsure about two points:
* how do I get the time zone as a string from `getTimeZoneSetup`, as that returns a QByteArray?
* is it really the best way to get the time in the correct time zone by getting the DateTime from the DateTimeEdit, converting it to a string and runnning it through IUnit?

kiozen:
Feed the QByteArray into a QString. I don't recall, but there was a reason to use QByteArray. Maybe this isn't a problem anymore. After all this code is quite old. 

If you find a more elegant way do it. This code has been written in many after work hours. And probably not every decision was a good one :wink:

kiozen:
@pingurus please test pull request #72

PR #44, opened 2019-10-09
[QMS-37] Inconsistent use of time zones

pingurus:
**What is the linked issue for this pull request (start with a `#`):** #37

I made the newDate filter use IUnit time zones
instead of QDateTime local time zones

1. Open track filters
2. Set new starting time of track

- [x] yes

- [x] yes

- [x] yes,
- [ ] no

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,5 @@
 V1.XX.X
+[QMS-37] Fix inconsistent use of time zones
 [QMS-27] Error in workspace search for attributes
 [QMS-18] Improved explanation of 'Date equals'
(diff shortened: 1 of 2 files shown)


pingurus:
> Now play with the time settings UTC, local time and automatic. This should result into different timezone strings. ("UTC", "CEST" and something for the US). It doesn't. And without that patch it's not working either wink Guess this has to be solved on long winter evening.

I wasn't aware, that that was what the `automatic` option did. Yes, like that that will be complicated to figuere out.

kiozen:
Is it ok to close that PR for now?

pingurus:
Sure.

Issue #43, opened 2019-10-09
Individual filter for each project


mitxel-m:
The new search on workspace feature is excellent...Thanks to pingurus and all the team ;-)

What about adding a filter for projects with the same procedure?

**Use case:**

You want to filter tracks below 1500 m from your 'Skitour' folder, you can use the search on workspace: elevation < 1500 m, all items from all projects out of the search criteria disappear. 

Ok you have found your tracks. Now you want to work with this filtered set, but at the same time you need to see on the map other items like your Summit wp collection, or other tracks from other projects/folders. For now the workaround is to copy the filtered set to a new temp project and then clear the filter on workspace. This is not efficient. 

 It would be nice to have a 'Filter Project' feature that would work only for the related Project (in addition to actual Search on workspace)

**How would it work**

Right click on a Project - select  'Filter...'

![menu](https://i.ibb.co/bHPbgC5/Filter-menu.jpg)

A dialog or a box appears to type the filter

![box2](https://i.ibb.co/hYVR7wt/search-box2.jpg)

Close or accept 

Now you have filtered items from this project besides all the elements from other projects. 

...and the Project appears marked with a Filter icon in the workspace list:

![icon](https://i.ibb.co/cyHvB35/Filter-icon-prj.jpg)

**How to clear the filter**

Right click on a Project - select  'Filter...'- click on the X 

![box2](https://i.ibb.co/hYVR7wt/search-box2.jpg)

(another approach: in the menu add an action 'Clear Filter', faster but enlarges the menu... I doubt which is better)

**Other considerations**
- The Project Filter is exclusive for the related Project, so you could apply at the same time a filter for every Project with different criteria. 
- You could apply at the same time the Workspace filter and a Project Filter, in this case you would get from the filtered project items matching both criteria.

kiozen:
Sounds a bit like another step to abusing the workspace as database replacement. The **idea is not** to have all your data in the workspace and to filter out the stuff you don't need. The **idea is** to have only the data you are working on in the workspace.

pingurus:
> Sounds a bit like another step to abusing the workspace as database replacement. The idea is not to have all your data in the workspace and to filter out the stuff you don't need. The idea is to have only the data you are working on in the workspace.

[...]

PR #45, opened 2019-10-09
[QMS-43] Individual filter for each project

pingurus:
**What is the linked issue for this pull request (start with a `#`):** #43

I added the option to make individual filters for each project

1. Right click a project and click `Filter Project`

- [x] yes

- [x] yes

- [x] yes,
- [ ] no

- [x] yes


kiozen:
> @kiozen , do you have an idea why the LineEdit appears at the bottom?
> And an idea how to achieve that the LineEdit is not removed on reorderig? Even when saving the QTreeWidgetItem the LineEdit is placed on and placing it back on top of the children the LineEdit is removed. However then, interestingly the QTreeWidgetItem appears at the top, as it already should have on insertion.

The child items of each project are sorted. They are grouped by type in the order as defined by GPX. And they are sorted by name or timestamp as configured.  Probably the easiest way to solve this is to remove the filter item before ordering/sorting and to add it again when done. Or you integrate it into the ordering.

pingurus:
I did integrate it into ordering. The QTreeWidgetItem that the CSearchLineEdit is placed on, is handled as I wanted it. But the CSearchLineEdit , that is placed by the CGisListWks onto that QTreeWidgetItem is lost. Not sure if it is clearer like that. I can put the code back in place, so you can see what I mean.

mitxel-m:
A new piece of feedback

Filter works separately for every project: OK

Some things to fix (maybe you are aware yet)

**Context menu**
- In the context menu the "Filter..." button keeps always the last action that was checked/unchecked, regardless whether the project has a filter or not.
 
  This leads to a strange behaviour, to reproduce it:
 
1. Check "Filter project" for project_A, and do a query. The filter box is shown
2. Go to Project_B and do right click, the filter button appears checked althought it has no filter. 

filtercheck2

3. Press the button and it turns to unchecked.
4. Come back to Project_A, it has a filter but the button in context menu is unchecked.

filtercheck

5. Check it and a duplicated filter box appears
 
 duplicated-filterline
 
the only way to close both them is doing the inverse way.
 
So the context menu do not look if the project has filter or not.,  

**Redraw canvas and project tree**

- when the query is typed the project tree is redrwawn but it has no effect in the canvas until you drag or zoom the map ( for worksapace search it redraws automatically)

- When you edit the project and the project tree is refrshed ( e.g. add/remove an item)  the filter is cleraed and the line edit box is hidden. Dont sure if this one has to be fixed, or  if it's better that way.

**For both lineedit boxes Filter Project and Workspace search**

- click on the X do nothing
- the string tooltip for ""?" does not match
- the string for "Case sensitive" does not match

settings

HTH ;-)

src/qmapshack/gis/search/CSearchLineEdit.cpp
```diff
@@ -0,0 +1,162 @@
...
+    actionHelp = new QAction(QIcon(":/icons/32x32/CSrcUnknown.png"), tr("Clear Filter"), this);
+    actionSetupFilter = new QAction(QIcon(":/icons/32x32/Apply.png"), tr("Setup Filter"), this);
+    actionError = new QAction(QIcon(":/icons/32x32/Attention.png"), tr("Error parsing search"), this);
+    actionAutoProperty = new QAction(QIcon(":/icons/32x32/Hint.png"), tr("Auto selected property"), this);
+
+    actionNameOnly = new QAction(tr("Name Only"), this);
+    actionNameOnly->setCheckable(true);
+    actionCompleteText = new QAction(tr("Complete Text"), this);
+    actionCompleteText->setCheckable(true);
+    actionCaseSensitive = new QAction(tr("Clear Filter"), this);
```
mitxel-m: the tr string does not match, in current version is "Case Sensitive" 
```suggestion
    actionCaseSensitive = new QAction(tr("Case Sensitive"), this);

**Comment by pingurus** [comment]
> In the context menu the "Filter..." button keeps always the last action that was checked/unchecked, regardless whether the project has a filter or not.

> when the query is typed the project tree is redrwawn but it has no effect in the canvas until you drag or zoom the map ( for worksapace search it redraws automatically)

Ah, thank you! I didn't see those yet.

> When you edit the project and the project tree is refrshed ( e.g. add/remove an item) the filter is cleraed and the line edit box is hidden. Dont sure if this one has to be fixed, or if it's better that way.

That is a temporary fix because of the stuff I asked @kiozen above. I would like to keep the filter.

> click on the X do nothing
    the string tooltip for ""?" does not match
    the string for "Case sensitive" does not match

I will check this, since it is obviously some error introduced by my refactoring.

File: src/qmapshack/gis/CGisListWks.cpp [hunk 2]
```diff
@@ -2081,6 +2083,39 @@ void CGisListWks::slotSyncDevWks()
...
+                hasSearch=true;
+                break;
+            }
+        }
+        if(!hasSearch)
+        {
+            QTreeWidgetItem* searchItem = new QTreeWidgetItem(project, {tr("Search:")});
+            CSearchLineEdit* edit = new CSearchLineEdit(this, project, searchItem);
+            project->insertChild(0, searchItem);
+            project->setSortingFolder(project->getSortingFolder());
```
pingurus: I don't really like this line of code. Any better idea of triggering a sorting to bring the search to the top?

src/qmapshack/gis/prj/IGisProject.cpp
```diff
@@ -1141,6 +1147,22 @@ void IGisProject::sortItems()
...
+    if(widget != nullptr)
+    {
+        for(QTreeWidgetItem * searchItem : others)
+        {
+            if(searchItem->text(0) == tr("Search:"))
+            {
+                CSearchLineEdit* edit = new CSearchLineEdit(widget, this, searchItem);
+                edit->setText(projectSearch.getSearchText());
+                //eColumnName is 1
+                widget->setItemWidget(searchItem, 1, edit);
```
pingurus: Same with this. Especially the fact that I can't use the enum, since that would require a circular inclusion

mitxel-m:
I have only been able to do a quick test of the last commit ( now i have to leave), and from my user point of view now everything works very well and is intuitive. 

;-)

kiozen:
I am quite busy right now. Don't know if I manage to review this week.

src/qmapshack/gis/CGisListWks.cpp
kiozen: Hm, this does not match the way items are added to a project at all. First each item is derived from QTreeWidgetItem (or IGisItem if it is a GIS item). In the case of the filter I would expect something like CProjectFilterItem derived from QTreeWidgetItem. The code to add CSearchLineEdit will be in CProjectFilterItem. To check if there is a filter item you don't have to check for tr("Search:") which might be subject to a change. You dynamic cast to CProjectFilterItem and check for a nullptr. 

And you tell IGisProject to enable or disable the filter item. Let IGisProject handle the necessary code. In IGisProject you have access to all methods. Event to protected sortItems() without doing that nasty hack you don't like,

src/qmapshack/gis/prj/IGisProject.cpp
kiozen: Why don't you handle the filter item the same way as all other item? I don't really see why this shouldn't be possible.

src/qmapshack/gis/CGisListWks.cpp
pingurus: Ah of course :man_facepalming:
Thanks!

pingurus:
I refactored the code, however I am not sure how to proceed further. To implement the information transfer I used signals and slots, which need the classes to be derived from `QObject`. `IGisProject` is already derived from `QTreeWidgetItem` and this causes an ambiguity for the call to `parent()`.
@kiozen do you have some best practices to handle this?

kiozen:
BTW files are missing :smile: 

Inheriting QObject together with QTreeWidgetItem is usually a bad idea and should be avoided. That shouldn't be a problem. You should be able to connect everything between the filter widget and the CGisWorkspace. And you can pass the pointer to the project item with your signals. If necessary.

kiozen:
Just asking because I think about a release: Will you be able to work on this any time soon. Or should we skip it for next release?

pingurus:
> 
> 
> Just asking because I think about a release: Will you be able to work on this any time soon. Or should we skip it for next release?

I might be able to work on it on the weekend, but no guarantee that I will get it done.

pingurus:
> BTW files are missing :smile: 

In the meantime the virtual machine I was develpoing on became unusable, but luckily I managed to locate the files after some time :relieved: :laughing:

kiozen:
No rush. There is always a next release. Let's keep it out of the next one. After all this is a fun project and no commercial project with a tight timetable for product increments.

pingurus:
@kiozen , could you please take a look why the compilation of 'qmapshack_autogen/mocs_compilation.cpp.o' fails? Adding Q_DECLARE_METATYPE doesn't work since IGisProject doesn't have a default constructor. 

Also, how do I remove the prj/CProjectFilterItem files? I tried deleting them in the browser, then pulling and rebasing, but they appeared again. The ones I use are those at search/CProjectFilterItem

src/qmapshack/gis/search/CSearchLineEdit.cpp
```diff
@@ -0,0 +1,184 @@
...
+    connect(actionCompleteText, &QAction::triggered, this, &CSearchLineEdit::slotSearchCompleteText);
+    connect(actionCaseSensitive, &QAction::triggered, this, &CSearchLineEdit::slotCaseSensitive);
+    connect(actionHelp, &QAction::triggered, this, &CSearchLineEdit::slotSearchHelp);
+    connect(actionClearFilter, &QAction::triggered, this, &CSearchLineEdit::slotClearFilter);
+    connect(this, &CSearchLineEdit::textChanged, this, &CSearchLineEdit::slotCreateSearch);
+}
+
+CSearchLineEdit::CSearchLineEdit(QWidget *parent, IGisProject* project, CSearch* search) : CSearchLineEdit(parent)
+{
+    connectedProject=project;
```
kiozen:
```
CSearchLineEdit::CSearchLineEdit(....
    : CSearchLineEdit(parent)
    , connectedProject(project)
```

src/qmapshack/gis/search/CSearchLineEdit.cpp
```diff
@@ -0,0 +1,184 @@
...
+}
+
+void CSearchLineEdit::slotSearchHelp()
+{
+    if(explanationDlg == nullptr)
+    {
+        explanationDlg = new CSearchExplanationDialog(this);
+    }
+
+    if(!explanationDlg->isVisible())
```
kiozen: Consider:
```
explanationDlg->setVisible(true);
```

src/qmapshack/gis/CGisListWks.cpp
```diff
@@ -112,6 +112,7 @@ CGisListWks::CGisListWks(QWidget *parent)
     actionGroupSort->setExclusive(true);
     actionSortByTime    = addSortAction(this, actionGroupSort, "://icons/32x32/Time.png", tr("Sort by Time"), IGisProject::eSortFolderTime);
     actionSortByName    = addSortAction(this, actionGroupSort, "://icons/32x32/SortName.png", tr("Sort by Name"), IGisProject::eSortFolderName);
+    actionFilterProject = addAction(QIcon("://icons/32x32/Filter.png"), tr("Filter Project"), this, SLOT(slotAddProjectFilter()));
```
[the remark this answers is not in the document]
pingurus: I had that in one of the previous versions, however that was irritating, because it didn't adjust to the corresponding project.
Is there a way to achieve that?

src/qmapshack/gis/CGisListWks.cpp
```diff
@@ -2081,6 +2083,31 @@ void CGisListWks::slotSyncDevWks()
...
+        {
+            continue;
+        }
+        CProjectFilterItem* projectFilter = project->filterProject(true);
+        connect(projectFilter->getLineEdit(), &CSearchLineEdit::searchChanged, this, &CGisListWks::slotUpdateProjectFilter);
+        connect(projectFilter->getLineEdit(), &CSearchLineEdit::searchCleared, this, &CGisListWks::slotRemoveProjectFilter);
+    }
+}
+
+void CGisListWks::slotUpdateProjectFilter(CSearch search, IGisProject* project)
```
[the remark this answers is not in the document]
pingurus: I think that didn't work because of circular inclusions, but I will look into it.

src/qmapshack/gis/CGisListWks.h
```diff
@@ -21,6 +21,7 @@
 #define CGISLISTWKS_H
 
 #include "gis/prj/IGisProject.h"
+#include "gis/search/CSearchLineEdit.h"
```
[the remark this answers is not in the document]
pingurus: Ah, damn I didn't think about that possibility yet :man_facepalming: :man_facepalming:

src/qmapshack/gis/prj/IGisProject.cpp
```diff
@@ -1141,6 +1151,11 @@ void IGisProject::sortItems()
     }
 
     addChildren(items);
+    if(projectFilter != nullptr)
+    {
+        projectFilter->showLineEdit(&projectSearch);
+    }
+    applyFilters();
```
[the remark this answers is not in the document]
pingurus: I will look into it

src/qmapshack/gis/CGisListWks.cpp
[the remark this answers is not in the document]
kiozen: Don't you have already a method to add a `CProjectFilterItem` and to remove it again? And if you do not use the 'X' button in the search line to  remove it but to clear the line edit's content things should be pretty straight forward.

src/qmapshack/gis/CGisListWks.cpp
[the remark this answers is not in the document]
kiozen: As a signal is a direct call of the callback (slot) it shouldn't make any difference. As long as you do not use signals/slots across threads.

src/qmapshack/gis/prj/IGisProject.cpp
```diff
@@ -1215,3 +1245,26 @@ void IGisProject::gainUserFocus(bool yes)
...
+            insertChild(0, projectFilter);
+        }
+        //Set expanded anyways to show that search exists
+        projectFilter->showLineEdit();
+        setExpanded(true);
+    }
+    else
+    {
+        removeChild(projectFilter);
+        delete projectFilter;
```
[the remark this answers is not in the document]
pingurus: As far as I can see, this is not possible since CProjectFilterItem is not derived from QObject and according to qt's documentation you can use delete to a QTreeWidgetItem anytime.

src/qmapshack/gis/search/CProjectFilterItem.h
```diff
@@ -0,0 +1,41 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+**********************************************************************************************/
+#ifndef CPROJECTFILTERITEM_H
+#define CPROJECTFILTERITEM_H
+
+#include <gis/search/CSearchLineEdit.h>
```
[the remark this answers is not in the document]
pingurus: This is not possible when using QPointer, right? (Just to make sure it complained because of something else)

src/qmapshack/gis/search/CSearchLineEdit.cpp
pingurus: This is not possible, since an initialiser for a delegated constructor must appear alone

src/qmapshack/gis/prj/IGisProject.cpp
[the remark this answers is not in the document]
pingurus: This works for me, maybe it was solved by straightening out the signals

src/qmapshack/gis/search/CProjectFilterItem.h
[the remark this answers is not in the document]
pingurus: Nevermind, in the end I didn't use QPointer.

src/qmapshack/gis/search/CSearchLineEdit.h
```diff
@@ -0,0 +1,67 @@
...
+#include <QMenu>
+class IGisProject;
+
+class CSearchLineEdit : public QLineEdit
+{
+    Q_OBJECT
+public:
+    CSearchLineEdit(QWidget* parent);
+    CSearchLineEdit(QWidget* parent, IGisProject* project, CSearch *search=nullptr);
+    ~CSearchLineEdit()
```
kiozen: `virtual ~CSearchLineEdit() = default`

pingurus:
Are you sure you have the current version? This does not happen for me.

src/qmapshack/gis/search/CSearchLineEdit.h
```diff
@@ -0,0 +1,67 @@
...
+
+**********************************************************************************************/
+#ifndef CSEARCHLINEEDIT_H
+#define CSEARCHLINEEDIT_H
+
+#include "CSearchExplanationDialog.h"
+#include "gis/search/CSearch.h"
+#include <canvas/CCanvas.h>
+#include <QLineEdit>
+#include <QMenu>
```
kiozen: Please check if all these headers are need and/or can be replaced by forward declarations.

src/qmapshack/gis/search/CSearchLineEdit.h
pingurus: Can't I just ommit it to produce the same result?

kiozen:
> Are you sure you have the current version? This does not happen for me.

Just pulled again. No changes. Crash is still there. Looks like a race. But haven't investigated details.

src/qmapshack/gis/prj/IGisProject.cpp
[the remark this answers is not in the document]
kiozen: right, it's a `QTreeWidgetItem`. 

But the root problem is still existing: `CSearchLineEdit::slotClearFilter()` calls `connectedProject->filterProject(false);`  The will delete the `QTreeWidgetItem` that owns the `CSearchLineEdit`. From my understanding this will delete the `CSearchLineEdit` wile it is still in `slotClearFilter`. It depends a bit how this is coded in Qt, but nevertheless we shouldn't expect or count on anything. This is risky code. On Linux such stuff usually works, on Windows it often fails.

And from my point of view it's not necessary. Pressing the `X` in the filter shouldn't remove it. It just should clear it.

kiozen:
Still crashing. See attached files for a fix.

[CProjectFilterItem.zip](https://github.com/Maproom/qmapshack/files/3857217/CProjectFilterItem.zip)

src/qmapshack/gis/CGisListWks.cpp
```diff
@@ -2081,6 +2096,22 @@ void CGisListWks::slotSyncDevWks()
...
+    if(project != nullptr)
+    {
+        if(actionFilterProject->isChecked())
+        {
+            project->filterProject(true);
+        }
+        else
+        {
+            project->filterProject(false);
+        }
```
kiozen:
```
project->filterProject(actionFilterProject->isChecked())
```

src/qmapshack/gis/CGisListWks.cpp
```diff
@@ -1270,6 +1275,16 @@ void CGisListWks::slotContextMenu(const QPoint& point)
...
 
+                    actionFilterProject->setEnabled(true);
+                    if(project->getProjectFilterItem() == nullptr)
+                    {
+                        actionFilterProject->setChecked(false);
+                    }
+                    else
+                    {
+                        actionFilterProject->setChecked(true);
+                    }
```
kiozen:
```
actionFilterProject->setChecked(project->getProjectFilterItem() != nullptr);

```

src/qmapshack/gis/prj/IGisProject.cpp
```diff
@@ -1141,6 +1145,11 @@ void IGisProject::sortItems()
     }
 
     addChildren(items);
+    if(projectFilter != nullptr)
+    {
+        projectFilter->showLineEdit(&projectSearch);
+    }
```
kiozen: Why do we have to destroy and instantiate the `CSearchLineEdit` widget on each sorting action?

pingurus:
Thanks!

src/qmapshack/gis/prj/IGisProject.cpp
pingurus: `CSearchLineEdit` is placed onto the `CProjectFilterItem`. However, when sorting the project, every child is removed and qt automatically destroys the widget, since the `QTreeWidgetItem` that it is placed on isn't part of the `QTreeWidget` anymore. Thus we have to recreate it after every sorting.

src/qmapshack/gis/search/CSearchLineEdit.cpp
```diff
@@ -0,0 +1,202 @@
...
+    {
+        CSearch::setCaseSensitivity(Qt::CaseInsensitive);
+    }
+    //Trigger reprocessing of search
+    slotCreateSearch(text());
+}
+
+void CSearchLineEdit::slotCreateSearch(const QString& str)
+{
+    actionClearFilter->setIcon(str.isEmpty() ? QIcon("://icons/32x32/Filter.png") : QIcon("://icons/32x32/Cancel.png"));
```
kiozen: I think we have to set 
```
        CCanvasCursorLock cursorLock(Qt::WaitCursor, __func__);
        QMutexLocker lock(&IGisItem::mutexItems);
```

somewhere in this method. But I am not sure where to place it. It shouldn't interfere with

```
void CGisWorkspace::slotSearch(const CSearch& currentSearch)
```

kiozen: A bit more background information:

The draw thread is iterating over the workspace tree widget to draw all items. If that happens it's a bad idea to alter the tree widget or to access items. That's why each operation on the workspace tree widget should be blocked by the `IGisItem` mutex. 

Showing a wait cursor is just good practice.

pingurus: What about placing it into `IGisProject::applyFilters()`, where the actual manipulation happens?

PR #61, opened 2019-11-01
do not try to examine devices with no path

moench-tegeder:
On FreeBSD, bsdisks cannot (yet) return details on nvme devices.
This wouldn't be much of a problem, since we're searching for
removable devices here - but when we try to get the driveIface for
that device-with-NULL-path, the dbus library triggers an abort(),
which crashes the whole qmapshack. So we catch that nullptr and
return. This is portable across all relevant platforms, as a NULL
drive_object.path() would cause that abort() about anywhere; and
it is safe (no leakage) as all objects created so far in slotDeviceAdded()
only have local scope.

**What is the linked issue for this pull request (start with a `#`):** QMS-#

I added...

1. Open whatever
2. Click here

- [ ] yes

- [ ] yes

- [ ] yes,
- [ ] no


kiozen:
Sorry I have to decline this pull request as it completely ignores the process of contributing code.

* It lacks a ticket
* It ignores the questions of the commit template (e.g. asking for a ticket number)

https://github.com/Maproom/qmapshack/wiki/DeveloperCommitCode

Issue #79, opened 2019-11-14
Possibility to 'star' Waypoints


pingurus:
**Is your feature request related to a problem? Please describe.**
When planning geocaching trips, I always have Geocaches I want to look for in any case and some that I don't specifically want to find. I plan my route with those I specifically want to find and then I import other geocaches aswell, since they are on my way. Now when I replan my route, I don't know anymore which are those that I specifically want to find and which I can remove.

**Describe the solution you'd like**
One can add a start to a project, to mark it as the active one. The same way, I'd like to mark any waypoint as an important one.

**Describe alternatives you've considered**
An alternative would be adding offline notes to the saved geocaches, and show in the workspace tree if a geocache has an offline note. This is more complicated (probably) and the use-case is not that broad. However, one could also implement both, since the offline note can be useful in other cases aswell.

**Additional context**

kiozen:
Tagging a project as active project is much more than just adding a star. It is a mode and all new items are automatically added to the active project. Therefore we shouldn't use anything that is looking equal to that in the GUI.

Adding notes that are not part of the GPX definitions and are only QMapShack internal for all items is more the way to go. This information can be shown in the info text. And be subject to filtering.

pingurus:
If I see it correctly, the geocaches in QMS don't use the `wpt.cmt` field of the `CGisItemWpt`.
So my idea would be, that we add a tab to the geocache edit view to edit the comment. Then we can show an icon next to every `IGisItem` that has a comment, so we achieve what I wished for.

kiozen:
But the chances are real that another user asks for the same feature for tracks, routes, areas. That why I think we should add it in a more general way to `IGisItem`

pingurus:
I just checked, we do have `virtual const QString& getComment() const = 0;` in `IGisItem.h`. My naive approach would be to enable the comment for the geocaches and display an icon whenever `gisItem->getComment()!=""`. This would be minimal effort and usable for any `IGisItem`.

kiozen:
What you want to have is a way to rate items internally. `comment` and `description` are part of the GPX definition. Usually they are used for other stuff. If you add rating GPX can't really handle it. 

You can add rating information to GPX fields. But to make QMapShack aware of the rating it would need a defined syntax and/or escape sequence that might upset devices or other software.

That's why I would suggest to make an QMapShack internal field/structure (int: rating, string: keywords) for rating times.

pingurus:
Ah, now I see what you want to do! :smile:
[...]

PR #92, opened 2020-01-06
[QMS-79] Possibility to 'star' Waypoint

pingurus:
**What is the linked issue for this pull request (start with a `#`):** QMS-#79

I added the possibility to rate and tag GisItems.

1. Open some project
2. Right-Click a GisItem and click on "Set Tags"
3. Fill fields and/or move slider
4. Filter or order by rating or keywords

- [x] yes

- [x] yes

- [x] yes,
- [ ] no


pingurus:
- The Rating column takes much space on the first start moving the name column out of the visible widget space. Very irritating.
    - Sure, I can fix that.
- I don't like the Ratingcolumn in general. It uses too much real estate of workspace list. There must be a better solution e.g. appending the name.
    - I don't really think that appending to a name is a nice way showing it's rating. In what form would you append it? "rating: 2" would use much more space. "(2)" can be confusing just as "(**)" can. Also, in which column should the tag symbol then come? A field can only have one icon and all current columns already have an icon at some point. Also, I don't think it is using much space, but that depends on screen size, I guess.
- The slider in the setup dialog isn't really nice. We can use the star icons from the caches to do a combobox or create our own slider widget.
    - Sure
- Any reason why using semicolons and no comma? I think must users would consider commas more natural.
    - No reason, for me this was more natural. I don't mind changing it

kiozen:
>       * I don't really think that appending to a name is a nice way showing it's rating. In what form would you append it? "rating: 2" would use much more space. "(2)" can be confusing just as "(**)" can. Also, in which column should the tag symbol then come? A field can only have one icon and all current columns already have an icon at some point. Also, I don't think it is using much space, but that depends on screen size, I guess.

Yes I know, the size problem sucks. Maybe using two lines for items with rating/tags is a feasible solution. First line rating/tags, second name. Worth a try?

pingurus:
> Maybe using two lines for items with rating/tags is a feasible solution.

But that is then using vertical space instead of horizontal. In my opinion that is worse, since it causes additional scrolling whereas the other doesn't bother the user (except when having a small screen). I could let the user hide the current column solution for those that don't have the horizontal space.

kiozen:
Well, you will have to die one death. At least it's easer to gain vertical space by arranging the docker widgets without loosing map real estate.

Another option would be to drop this information from the treewidget and just rely on the item's information text.

pingurus:
So what do you think of a toggle box in the view menu that enables or hides the extra column? Those that disable it have to rely on the info text then.

kiozen:
It should be the workspace menu or workspace setup. 

Another option would be to dynamically create the item icon with additional information. In that case the icon size should be 32x16 if that is possible.

pingurus:
>It should be the workspace menu or workspace setup

True.

> Another option would be to dynamically create the item icon with additional information. In that case the icon size should be 32x16 if that is possible.

Wouldn't that be too expensive for large workspaces (fixable if you cache the icons) and many different waypoint symbols (no idea how the dynamic creation performs)?

pingurus:
What do you think of this solution for the column?
I will fix the semicolon and the rating issues later.

pingurus:
@kiozen what do you think of the current solution for the slider? If you like the concept I will improve the positioning of the symbols.

kiozen:
Still doesn't look very nice. I would go for a combobox or radio buttons.

pingurus:
Not sure a combobox or radio buttons would look any better.
How about a vertical slider to the left with stars 'attached' to each tick? Those stars could then be placed in a horizontal line to look better

kiozen:
I still hate this Qt slider, sorry. You could write your own slider widget with the star icons. Just show 5 empty stars and depending on a mouse click or click and move you can select the rating in stars or half-stars. That would be kind of nice.

But a combobox is fine, too.

kiozen:
ok, much better. You might want to play a bit with the button settings to avoid the button press style. 

The rating column still expands to much the first time it is shown.

If you edit the rating and the tags several times, the behaviour is a bit peculiar. Tags are duplicated or removed. I haven't figured out scheme behind it, yet.

src/qmapshack/gis/qms/serialization.cpp
```diff
@@ -718,6 +731,7 @@ QDataStream& CGisItemWpt::operator<<(QDataStream& stream)
     setIcon();
     setText   (CGisListWks::eColumnName, getName());
     setToolTip(CGisListWks::eColumnName, getInfo(IGisItem::eFeatureShowName));
+    IGisItem::updateDecoration(eMarkNone, eMarkNone);
```
kiozen: Why?

src/qmapshack/gis/trk/CGisItemTrk.cpp
```diff
@@ -630,6 +630,17 @@ QString CGisItemTrk::getInfo(quint32 feature) const
         }
     }
 
+    if(getRating() > 0)
```
kiozen: Can't we do this for all items in IGisItem? Saves us duplicating the code 4 times.

src/qmapshack/gis/CGisItemRate.cpp
```diff
@@ -0,0 +1,81 @@
...
+    rating(rating)
+{
+    setupUi(this);
+    keywordPlainTextEdit->setPlainText(QStringList(commonKeywords.toList()).join(", "));
+
+    pushButtonR1->setIcon(QPixmap(rating < 1 ? "://icons/cache/32x32/star_empty.png" :  "://icons/cache/32x32/star.png"));
+    pushButtonR2->setIcon(QPixmap(rating < 2 ? "://icons/cache/32x32/star_empty.png" :  "://icons/cache/32x32/star.png"));
+    pushButtonR3->setIcon(QPixmap(rating < 3 ? "://icons/cache/32x32/star_empty.png" :  "://icons/cache/32x32/star.png"));
+    pushButtonR4->setIcon(QPixmap(rating < 4 ? "://icons/cache/32x32/star_empty.png" :  "://icons/cache/32x32/star.png"));
+    pushButtonR5->setIcon(QPixmap(rating < 5 ? "://icons/cache/32x32/star_empty.png" :  "://icons/cache/32x32/star.png"));
```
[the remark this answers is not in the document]
pingurus: No, calling `ratingLabelClicked (rating)` will cause rating to decrement.

src/qmapshack/gis/CGisItemRate.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+bool IGisItemRate::getRatingChanged()
+{
+    return ratingChanged;
+}
+
+QSet<QString> IGisItemRate::getAddedKeywords()
+{
+    QSet<QString> keywords = keywordPlainTextEdit->toPlainText().split(QRegExp("\w*,s\w*"), QString::SkipEmptyParts).toSet();
```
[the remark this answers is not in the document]
pingurus: Not `\\w`?

src/qmapshack/gis/IGisItem.cpp
```diff
@@ -1037,3 +1056,37 @@ IGisItem * IGisItem::newGisItem(quint32 type, quint64 id, QSqlDatabase& db, IGis
...
+
+qreal IGisItem::getRating() const
+{
+    return rating;
+}
+
+void IGisItem::setRating(qreal rating)
+{
+    this->rating=rating;
+    changed("changed rating", "://icons/32x32/Tag.png");
```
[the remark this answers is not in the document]
pingurus: What would be the correct way to mark it changed without adding it to the history?

src/qmapshack/gis/trk/CGisItemTrk.cpp
pingurus: Yes, good idea.

src/qmapshack/gis/IGisItem.cpp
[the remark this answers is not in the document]
kiozen: use `updateHistory()`

src/qmapshack/gis/CGisItemRate.cpp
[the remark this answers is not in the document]
kiozen: you are right.

src/qmapshack/gis/CGisItemRate.cpp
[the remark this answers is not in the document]
kiozen: Then move this block of push button setIcon() calls in their own private method that its shared between ctor and `ratingLabelClicked()`

src/qmapshack/gis/qms/serialization.cpp
pingurus: It causes the symbols for the tags to appear. I should move it into the `if(version){}`.
Or do you want it to be somewhere else?

kiozen: Then we have duplicated code. Probably all set... methods are covered by updateDecoration, too.

pingurus:
> The rating column still expands to much the first time it is shown.

How can I reproduce this? Hiding and unhiding works fine for me

kiozen:
simply switch between your version and official version.

pingurus:
Is the Version number of QMS saved in the config? I'm trying to figure out a way how to decide when the header should be restored and when not.

kiozen:
no it's not. Better start with a sane default.

pingurus:
Should be working now. Please test if the workaround is device specific.

kiozen:
`Rating` column is still way to large if activated the first time.

pingurus:
Just tested again, it works for me.
Compiled and ran the `dev` branch. Exited normally. Compiled and ran the `QMS-79` branch. Lines of code to ensure that the rating column is shortened do what they should and the result looks good aswell.
Could you please test for me what `treeWks->columnWidth(CGisListWks::eColumnName)` returns on your machine, when you start `QMS-79` for the first time?

kiozen:
It's 100. I doubt this is a good approach.

pingurus:
So what abour leaving it as is? It's a one-time inconvenience for the user.
Is there no way to find out in which version QMS was last opened?

kiozen:
No we can't leave it like it is. It too confusing and I fear it will result in quite a bunch of bug reports, complains and bad publicity.

pingurus:
Yes, true point.
So how about starting to write the last QMS version it was opened with into the config and if it is not present (as now) we don't load the header from the config?

kiozen:
I even doubt that hiding the column is a good idea. Imho it would be better let the user actively add that column. And remove it again if not wanted.

pingurus:
What do you mean by hiding? The option to toggle from the workspace setup wether the column is shown or not?
Might it be an option to hide the column right in the beginning and only showing it when the user activel enables it? Then one could always set the Width of the column to the default value whenever this option is toggled and thus fixing the problem

kiozen:
That might be a solution.

kiozen:
Current version in repository does not compile.

pingurus:
Sorry, forgot to commit

kiozen:
Changing the number of tags, does not update the number in the workspace tree widget.

src/qmapshack/gis/IGisItem.h
```diff
@@ -592,6 +593,14 @@ class IGisItem : public QTreeWidgetItem
 
     virtual const searchValue_t getValueByKeyword(searchProperty_e keyword) = 0;
 
+    qreal getRating() const;
+    void setRating(qreal rating);
+    const QSet<QString>& getKeywords() const;
+    void addKeywords(const QSet<QString> &otherKeywords);
+    void removeKeywords(const QSet<QString> &otherKeywords);
+    void clearKeywords();
```
kiozen: Where is `clearKeywords()` needed?

kiozen:
And please clean-up your commit history.

pingurus:
> 
> 
> And please clean-up your commit history.

I can do so of course, but I must say I don't really see the point if you squash the commits anyway. Or do I get something wrong?

kiozen:
A clean commit history saves me time when merging the stuff because I do not have to create a commit message out of all the cluttered commit messages. 

A clean commit history helps you to sort in your patches into thematic commits. And to document each commit. Right now it's just a mess of several commits, with mostly the same commit header. 

A clean commit history helps with large PR as the reviewer can focus on single thematic commits. Right now the only view that makes sense is the complete change. But it would help to see what files are affected by the tags dialog. What files are affected by store/restore workspace. What files are affected by serializing the stuff. What files are affected by search and sort. With a clean history the reviewer can just focus in the changes done by the thematic commit. 

A clean commit history is the advantage of GIT.  But of course it requires a bit more discipline.

pingurus:
> 
> 
> A clean commit history saves me time when merging the stuff because I do not have to create a commit message out of all the cluttered commit messages.
> 
> A clean commit history helps you to sort in your patches into thematic commits. And to document each commit. Right now it's just a mess of several commits, with mostly the same commit header.
> 
> A clean commit history helps with large PR as the reviewer can focus on single thematic commits. Right now the only view that makes sense is the complete change. But it would help to see what files are affected by the tags dialog. What files are affected by store/restore workspace. What files are affected by serializing the stuff. What files are affected by search and sort. With a clean history the reviewer can just focus in the changes done by the thematic commit.
> 
> A clean commit history is the advantage of GIT. But of course it requires a bit more discipline.

Is it better now? Please don't make me tear apart those commits and repack them so they are more specific to single improvements. 
I will take care next time to have all the stuff separated. Can I add the requested changes to the single commit where I've put them until now?

kiozen:
It's better now.

You can do more commits and sort them in. See `git rebase -i HEAD~<number of commits>` This will open an interactive rebase for the last <number of commits> Now you can reorder commits and use `f` to merge them.

pingurus:
I'm sory that I didn't sort those changes in, but for example renaming the column would have been logical from the beginning on, but in some places it only appeared later. That's why I've simply amended the latest commit.

kiozen:
In theory you could soft reset all commits and do clean grouped commits. If you like challenges you can try (backup!) If not it's ok to amend the last commit of this PR :wink:

kiozen:
Hm, simple smoke test is failing.

1. Workspace is not restored. I set rating column visible but after a restart it's not visible. The property in the config file reads `treeWks\tagsHidden=false`

2. And the tags algorithm for common tags does not work properly. I play around wit combinations of aaa, bbbb, and so on. I set mutual exclusive tags and some common tags. I group edit items, I edit single items. I always end up in inconsistent tags. Some times a tag is added twice. Often common tags are not shown. This is done with several app restarts in between. The behavior is too erratically to be understood by the user.

kiozen:
3. After changing the rating of an item with tags and rating, no tags icon is shown, however tags are listed in the item info. Calling the edit dialog once again tags are missing. Restarting QMS the tags are completely gone.

pingurus:
> After changing the rating of an item with tags and rating, no tags icon is shown, however tags are listed in the item info. Calling the edit dialog once again tags are missing. Restarting QMS the tags are completely gone.

Not sure what you mean by this.
Are the other problems now fixed for you too?

pingurus:
Should I move the links below the rating and the tags? I think that would look better.

kiozen:
I took the freedom to add two patches:

1. Adding the tags/rating option to the on-screen option dialogs
2. Optimizing the code (unrelated to your changes)

pingurus:
> Should I move the links below the rating and the tags? I think that would look better.

Other small changes, that I've been thinking of:

- If the user selectes multiple items, should the rating really show the average or rather nothing? Question is what is more inuitive for how it works (as long as you don't touch the stars, the rating won't be changed for any of the items)
- Should the tags be in alphabetical order? Or by time added? Currently it is random.

kiozen:
No stars for multiple items.

Alphanumeric order. See 

```
static bool sortByName(const QString& item1, const QString& item2)
```
how it's done.

pingurus:
I left the links where they were, it's maybe a bit more logically grouped like that.
Any more changes?

kiozen:
I amended the last commit with a bit of code cleanup.

src/qmapshack/gis/CGisWorkspace.cpp
```diff
@@ -1341,3 +1345,63 @@ bool CGisWorkspace::findPolylineCloseBy(const QPointF& pt1, const QPointF& pt2,
...
+void CGisWorkspace::setTagsHidden(bool hidden)
+{
+    treeWks->setColumnHidden(CGisListWks::eColumnRating, hidden);
+}
+
+bool CGisWorkspace::firstTimeWithTags()
+{
+    treeWks->setColumnWidth(CGisListWks::eColumnRating, 100);
+    return true;
+}
```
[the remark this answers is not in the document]
pingurus: I thought cfg.value() only returns (and thus executes) the second parameter if the key is not found. But I already have a different solution in mind so it's no problem.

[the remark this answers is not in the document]
kiozen: nope, this is not Python :smile:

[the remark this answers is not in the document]
pingurus: True true :laughing:

pingurus:
Anything else left to do?
Is it fine that I made a new commit for this?

kiozen:
Sorry, it's still buggy. I used your version and hide the rating column. I switched to current dev version and now the name column has disappeared. For a normal user it's impossible to restore it.

Please keep in mind that these things have to be kind of backward compatible and please test that stuff before you commit. There will be users that try a new version and switch back to the former one for whatever reason. At least that step has to work without editing the config file or the Windows Registry.

kiozen:
Another smoke test fails: Hide the rating column and restart. Rating column is visible again. Config says `treeWks\tagsHidden=true`

pingurus:
Yes, I'm sorry. I definitely shold test my code better :laughing:

pingurus:
I fixed the behaviour you found, however I couldn't come up with a solution to avoid that when switching back and forth again the rating column is as wide as the name column was (and thus pushing the name column out of view).
Problem is, that the settings I set are preserved when opening with the old version and thus I can't detect that the column width needs to be fixed.

src/qmapshack/gis/CGisWorkspace.cpp
```diff
@@ -51,6 +52,13 @@
 #include <QtWidgets>
 #include <QtXml>
 
+enum tags_hidden_e
+{
+    eTagsHiddenTrue,
+    eTagsHiddenFalse,
+    eTagsHiddenUnknown
+};
+
```
kiozen: Why is this outside of the class? Any good reason?

pingurus: I can't remeber why I didn't palce it into CGisWorkspace, that I decided not to place it into CGisItemRate because it has nothing to do with that.
So probably the answer is no.

kiozen: Then place it into the class.

kiozen:
with last commit Rating column is too small every time I start QMS. Size changes are not restored.

Honestly, can't we get rid of the column completely?  It's just causing problems since weeks.

pingurus:
> Honestly, can't we get rid of the column completely?

Yeah, I don't know. I actually really want that column, so you can find the tagged waypoints quickly without having to filter. Can you please try with the latest version?

src/qmapshack/gis/CGisWorkspace.cpp
```diff
@@ -63,6 +64,18 @@ CGisWorkspace::CGisWorkspace(QMenu *menuProject, QWidget *parent)
 
     SETTINGS;
     treeWks->header()->restoreState(cfg.value("Workspace/treeWks/state", treeWks->header()->saveState()).toByteArray());
+
+    tags_hidden_e tagsHidden = (tags_hidden_e)cfg.value("Workspace/treeWks/tagsHidden", eTagsHiddenUnknown).toInt();
+    // Limit the column width to half the table width to avoid the name column to be pushed out the window.
+    if(tagsHidden == eTagsHiddenUnknown || treeWks->columnWidth(CGisListWks::eColumnRating) > 0.4 * this->width())
+    {
+        treeWks->setColumnWidth(CGisListWks::eColumnRating, 0.2 * this->width());
+    }
```
kiozen: Can you explain why you compare to 0.4 but limit to 0.2?

pingurus: I compare to 0.4, since that is approx the size at which the name gets pushed out. However it should default to a small value (0.2 seems good on my screen and since it is a ratio I hope it is on other screens too).

The difference in the two values should make sure the problem you described with the version before does not occur.

Issue #116, opened 2020-03-12
Filter edit field lost


wthaem:
**Describe the bug**
* Select entry `Filter project` in a project context menu.
* In older QMS versions an edit field with some decoration is shown below the project name in the workspace, in the latest commit (star waypoints) this field is missing and thus filtering doesn't work anymore. Compare attached image which shows the 2 different layouts.

**To Reproduce**

see description of bug

**Expected behavior**

Old behavior should be re-established

**Screenshots**
![Filter](https://user-images.githubusercontent.com/55309218/76507924-e2e2df80-644d-11ea-8c38-52b1a332f893.jpg)

**Desktop (please complete the following information):**
 - OS: Windows10
 - QMapShack latest commit (star waypoints)

**Additional context**

None

pingurus:
Ah :man_facepalming: :man_facepalming: 
I didn't think of that. It's a quick fix though.

PR #117, opened 2020-03-12
[QMS-116] Filter edit field lost

pingurus:
**What is the linked issue for this pull request (start with a `#`):** QMS-#116

* Fixed the cause of losing the field
* Add missing descriptions

1. Open a project
2. Filter the project
3. The filter field should be there regardless of the rating column being shown or not.

- [x] yes

- [x] yes

- [x] yes,
- [ ] no

### DIFF
--- a/src/qmapshack/gis/search/CSearch.cpp
+++ b/src/qmapshack/gis/search/CSearch.cpp
@@ -507,6 +507,8 @@ QMap<searchProperty_e, QString> CSearch::initSearchPropertyMeaningMap()
     map.insert(eSearchPropertyGeneralDate, tr("searches the Date"));
     map.insert(eSearchPropertyGeneralComment, tr("searches the Comment"));
     map.insert(eSearchPropertyGeneralDescription, tr("searches the Description"));
+    map.insert(eSearchPropertyGeneralKeywords, tr("searches the Keywords"));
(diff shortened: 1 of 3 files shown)


src/qmapshack/gis/search/CSearch.cpp
```diff
@@ -507,6 +507,8 @@ QMap<searchProperty_e, QString> CSearch::initSearchPropertyMeaningMap()
     map.insert(eSearchPropertyGeneralDate, tr("searches the Date"));
     map.insert(eSearchPropertyGeneralComment, tr("searches the Comment"));
     map.insert(eSearchPropertyGeneralDescription, tr("searches the Description"));
+    map.insert(eSearchPropertyGeneralKeywords, tr("searches the Keywords set"));
```
kiozen: "Set Tags"
"Keywords"  
The wording should be more unified.

pingurus: In my opinion tags contains both keywords and ratings.

kiozen: ah, ok. Good to know for translation. 

but change it to "searches the Keywords". We know it's a set but the user does not.

pingurus: Ah, the "set" shouldn't mean the set as a union of elements but as the verb. However I agree that it should go.

Issue #119, opened 2020-03-14
Better handling of disabled and archived geocaches


pingurus:
**Is your feature request related to a problem? Please describe.**
Up until now there is no way to filter for disabled or archived geocaches and also they're not visibly different besides the status description in the header of the detailed view.

**Describe the solution you'd like**
Two things came to my mind, feel free to add:
* Grey out the icon for disabled and archived geocaches
* Enable filtering for the status

**Describe alternatives you've considered**
None

**Additional context**
None

PR #121, opened 2020-03-14
[QMS-119] Better handling of disabled and archived geocaches

pingurus:
**What is the linked issue for this pull request (start with a `#`):** QMS-#119

I added the possibility to filter for the status of a geocache and greyed out geocaches that are disabled.

1. Open a file with active, disabled and archived geocaches
2. Filter for "status with archived" or similar
3. See that the icon is grey

- [x] yes

- [x] yes

- [x] yes,
- [ ] no

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,6 +15,7 @@ V1.XX.X
 [QMS-111] Geocaching logging page - only 1 server used
 [QMS-116] Filter edit field lost
 [QMS-118] Too much whitespace in extended search help window
+[QMS-119] Better handling of disabled and archived geocaches
 
 V1.14.0
 [QMS-5] App crashes on using the "Change Start Point" filter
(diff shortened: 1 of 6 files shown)


kiozen:
Instead of adding another bunch of icons that need to be maintained wouldn't it be better to gray scale these icons on-the-fly?

pingurus:
> 
> 
> Instead of adding another bunch of icons that need to be maintained wouldn't it be better to gray scale these icons on-the-fly?

I tried to do so, but I couldn't make those keep the transparency. Thus I decided that it would save my time if I generate them with a tool that works on the first try.

kiozen:
This mght not be the most beautiful code -  in fact it's a five minute hack - but in the core you have to save the alpha channel as mask and apply it to the gray scaled image, if that does not preserve the alpha channel by itself.
`    
    QPixmap pixmap("://icons/geocaching/icons/earth.png");
    QBitmap alpha = pixmap.createHeuristicMask();
    QImage image = pixmap.toImage().convertToFormat(QImage::Format_Grayscale8);
    QPixmap pixmap_gray = QPixmap::fromImage(image.convertToFormat(QImage::Format_ARGB32));
    pixmap_gray.setMask(alpha);
    pixmap_gray.save("test.png");
`

pingurus:
Thanks!

Issue #62, opened 2019-11-01
Enhance handling of PROJ.4 files in Windows


wthaem:
**Is your feature request related to a problem? Please describe.**

Currently, the Windows QMS version can be compiled with different Proj.4 versions. But the 2 files

* `msvc_64\copyfiles.bat` and
* `msvc_64\QMapShack_Installer.nsi` 

require a fixed Proj.4 version. Any change of the Proj.4 version requires a change of these files.

**Describe the solution you'd like**

Replace lines with a fixed version (similar to proj_6_x.dll) in both files with a wildcard version (proj_6_*.dll). This will make both scripts independent of a fixed Proj.4 version and doesn't need any change in the documentation. (Idea taken from huirad).

**Describe alternatives you've considered**

None.

**Additional context**

None.

kiozen:
Btw, as this is another ticket that totally ignores the template, please do so in the future. I will have to close these tickets as they are a bad example and encourage others to do the same. This will end up in the same chaotic ticket style as we had on BitBucket. That's why I will enforce it.

wthaem:
Sorry for this. Edited content to match template form.

Issue #122, opened 2020-03-16
Wrong search path for Qt help files?


wthaem:
**Describe the bug**
(Not noticed earlier - sorry)
In the QMS Windows version the QMS files are stored in the directory `msvc_64/Files`. But the qmapshack.exe can be started from a different directory, say `cwd`, using the full path to call it.

Currently, QMS tries to find its Qt help files relative to `cwd`. This file distribution can't be handled by the QMS installation because `cwd` is not known.

**To Reproduce**
1. Save Qt help files in `msvc_64/Files/doc/...`
1. Start QMS from outside `msvc_64/files`and press F1 - result: empty help window.

**Expected behavior**
QMS should search its help files relative to `msvc_64/files` ( the location of qmapshack.exe).

**Screenshots**

not necessary
**Desktop (please complete the following information):**
 - OS: Windows 10
 - QMapShack latest commit

**Additional context**

none

pingurus:
On my Ubuntu 19.10 VM the help window is empty too, regardless of wether I open it via F1 or the question mark and then help (Started via Qt Debugger).

kiozen:
did you do `make install`? The help files must be copied into the correct path. And this will be done by the install target.

The initial ticket is expected. @huirad has to fix the code for Windows.

pingurus:
> did you do make install? The help files must be copied into the correct path. And this will be done by the install target.

No, I didn't.

kiozen:
try it :smile: It's the same with translation. You have to install to get latest update.

pingurus:
I did do
```
make
sudo make install
qmapshack
```
Same result, also if I start it from the programs screen 🤷‍♂

huirad:
Heeey, no hurry!

kiozen:
Haha, now that you respond to the topic I could assign you. Your fault :imp: :stuck_out_tongue:

kiozen:
@pingurus search the debug log for `search help at:` and cross check the path.

pingurus:
Where do I find the debug log?

kiozen:
start qmapshack with paramter `-d` from console.

pingurus:
```
2020-03-19 9:14:14.242 [debug] search help at: "/usr/local/share/doc/HTML/QMSHelp.qhc"
2020-03-19 9:14:14.315 [debug] QUrl("qthelp://qms/doc/doc/html/DocMain.html")
[...]

PR #130, opened 2020-03-19
[QMS-122] Windows: support for offline help

huirad:
Adapt Windows installer for offline help support

**What is the linked issue for this pull request (start with a `#`):** QMS-#122

I adapted the Windows installer scripts;
- add Qt5Help.dll
- add the offline documenation files

1. Create a Windows Installer
2. Press F1 or call for help via the menu

- [X] N/A

- [X] N/A

- [X] yes,
- [ ] no

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,7 +19,7 @@ V1.XX.X
 [QMS-119] Better handling of disabled and archived geocaches
 [QMS-125] Missing icon for not available geocache
 [QMS-126] Wrong results searching for geocache status
-
+[QMS-122] Windows: support for offline help
 
 V1.14.0
 [QMS-5] App crashes on using the "Change Start Point" filter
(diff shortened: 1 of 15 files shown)


wthaem:
Would there be a possibility to resolve issue #62?

huirad:
Hei Wolfgang,

Concerning your assumption that a code change is needed.
It tested the change and it works.
Both when I start QMS form the "Files" directory created by copyfiles.bat and when I start it as set up by the installer. I see no point in investigating theoritical cases.

Concerning your request about #62:
- cross posting does not make sense as the issues are not related
- even in these times I have a full-time job - QMS is only a spare time activity
If you want #62: please do it yourself. Thanks for your understanding.

wthaem:
> Both when I start QMS form the "Files" directory created by copyfiles.bat and when I start it as set up by the installer.

That's correct. But it describes at the same time the restriction I spoke about.

huirad:
@wthaem: Now I understood what you mean with "But the qmapshack.exe can be started from a different directory, say cwd, using the full path to call it."
Oh man, who can even think about such crazy use cases?
I will have a look, but don't expect the change too fast. I have many other things to do with higher priority.
Take care and stay healthy!

Issue #68, opened 2019-11-03
Feature Request: Route Optimization


WST1811:
**Is your feature request related to a problem? Please describe.**
I got a lot of Waypoints (20 to 25 or more) and want a Route/Track with the fastest Way between Start and End with touch all Waypoints. Now I have to sort the waypoints in the order in which they should optimally be approached.

**Describe the solution you'd like**
Here an option would be nice to have an automatic calculation from waypoint 1 over the other points to the last waypoint without having to specify the best order.

**Describe alternatives you've considered**

**Additional context**
Is this better feature request then #67?

pingurus:
The problem you are trying to solve is also called the Traveling Salesman Problem. Since QMapShack doesn't do the routing by it's own, we would need a router that is able to solve this problem. 

If you find a router, that is able to solve the TSP and has an API that QMS can use, I am sure the feature can somehow be fitted into QMS. However someone needs to implement this. If you can code C++ the fastest would be if you do it yourself. Since this is nothing I will use in near future, I must say I'm not too keen to implement it. However, if you provide a good preparation it might well be, that I will implement that feature sooner or later, since I do think that it might be a useful feature.

WST1811:
Thanks for answer. I don't know if this is what you need and you can use the API: 
https://openrouteservice.org/dev/#/api-docs/optimization/post
https://github.com/VROOM-Project/vroom/blob/master/docs/API.md

And/or this: https://www.graphhopper.com/blog/2019/05/17/getting-started-with-the-optimization-api-traveling-salesman-problem/

PR #141, opened 2020-04-02
[QMS-68] Route Optimization

pingurus:
**What is the linked issue for this pull request (start with a `#`):** QMS-#68

Since the TSP somehow fascinated me, I decided to write my own algorithm to optimize routes using the routers already available in QMS.

I adapted the routers, so they also return the cost of the route calculated using `calcRoute`. This enables the algorithm to actually optimize on what the router is trying to minimize.

The only option left to speed up that comes to my mind would be to intercept the `calcRoute` calls for the points the user actually uses, however I didn't find a good possibility to do so.

1. You have to have a router installed that is capable of fast routing.
2. Draw a route
3. Press the button 'optimize'
4. Wait (Brouter can take ~10-20 times longer than routino, with my heavily assymetric mountainbiking profile optimizing a route of 18 Points and 40km initial length takes ~30sec, shortest path with routino takes ~1 sec, shortest path with brouter takes ~20sec)
5. See the hopefully improved result.
6. If you are not happy with the result, press optimize again, the result should improve and come faster than the first time.

- [x] yes

- [x] yes

- [x] yes,
- [ ] no

### DIFF
--- a/src/qmapshack/gis/rte/router/CRouterBRouter.cpp
+++ b/src/qmapshack/gis/rte/router/CRouterBRouter.cpp
@@ -323,7 +323,7 @@ QNetworkRequest CRouterBRouter::getRequest(const QVector<QPointF> &routePoints, 
     return QNetworkRequest(url);
 }
 
-int CRouterBRouter::calcRoute(const QPointF& p1, const QPointF& p2, QPolygonF& coords)
+int CRouterBRouter::calcRoute(const QPointF& p1, const QPointF& p2, QPolygonF& coords, qreal *costs)
 {
     if(!hasFastRouting())
     {
@@ -335,10 +335,10 @@ int CRouterBRouter::calcRoute(const QPointF& p1, const QPointF& p2, QPolygonF& c
     QList<IGisItem*> nogos;
     CGisWorkspace::self().getNogoAreas(nogos);
 
-    return synchronousRequest(points, nogos, coords);
+    return synchronousRequest(points, nogos, coords, costs);
 }
 
-int CRouterBRouter::synchronousRequest(const QVector<QPointF> &points, const QList<IGisItem *> &nogos, QPolygonF &coords)
+int CRouterBRouter::synchronousRequest(const QVector<QPointF> &points, const QList<IGisItem *> &nogos, QPolygonF &coords, qreal* costs = nullptr)
 {
     if (!mutex.tryLock())
     {
@@ -377,43 +377,63 @@ int CRouterBRouter::synchronousRequest(const QVector<QPointF> &points, const QLi
         {
             throw reply->errorString();
         }
-        else
+        slotClearError();
+
+        const QByteArray &res = reply->readAll();
+
+        if(res.isEmpty())
         {
-            slotClearError();
+            throw tr("response is empty");
+        }
 
-            const QByteArray &res = reply->readAll();
+        QDomDocument xml;
+        xml.setContent(res);
+        const QDomElement &xmlGpx = xml.documentElement();
 
-            if(res.isEmpty())
+        if(xmlGpx.isNull() || xmlGpx.tagName() != "gpx")
+        {
+            throw QString(res);
+        }
+        setup->parseBRouterVersion(xmlGpx.attribute("creator"));
+
+        // read the shape
+        const QDomNodeList &xmlLatLng = xmlGpx.firstChildElement("trk")
+                                        .firstChildElement("trkseg")
+                                        .elementsByTagName("trkpt");
+        for(int n = 0; n < xmlLatLng.size(); n++)
+        {
+            const QDomElement &elem   = xmlLatLng.item(n).toElement();
+            coords << QPointF();
+            QPointF &point = coords.last();
+            point.setX(elem.attribute("lon").toFloat()*DEG_TO_RAD);
+            point.setY(elem.attribute("lat").toFloat()*DEG_TO_RAD);
+        }
+
+        //find costs of route (copied and adapted from CGisItemRte::setResultFromBrouter)
+        if(costs != nullptr)
+        {
+            const QDomNodeList &nodes = xml.childNodes();
+            for (int i = 0; i < nodes.count(); i++)
             {
-                throw tr("response is empty");
-            }
-            else
-            {
-                QDomDocument xml;
-                xml.setContent(res);
-
-                const QDomElement &xmlGpx = xml.documentElement();
-                if(xmlGpx.isNull() || xmlGpx.tagName() != "gpx")
+                const QDomNode &node = nodes.at(i);
+                if (!node.isComment())
                 {
-                    throw QString(res);
+                    continue;
                 }
-                else
+                const QString &commentTxt = node.toComment().data();
+                // ' track-length = 180864 filtered ascend = 428 plain-ascend = -172 cost=270249 '
+                const QRegExp rxAscDes("(\\s*track-length\\s*=\\s*)(-?\\d+)(\\s*)(filtered ascend\\s*=\\s*-?\\d+)(\\s*)(plain-ascend\\s*=\\s*-?\\d+)(\\s*)(cost\\s*=\\s*)(-?\\d+)(\\s*)");
+                int pos = rxAscDes.indexIn(commentTxt);
+                if (pos > -1)
                 {
-                    setup->parseBRouterVersion(xmlGpx.attribute("creator"));
-
-                    // read the shape
-                    const QDomNodeList &xmlLatLng = xmlGpx.firstChildElement("trk")
-                                                    .firstChildElement("trkseg")
-                                                    .elementsByTagName("trkpt");
-                    for(int n = 0; n < xmlLatLng.size(); n++)
+                    bool ok;
+                    *costs = rxAscDes.cap(9).toDouble(&ok);
+                    if(!ok)
                     {
-                        const QDomElement &elem   = xmlLatLng.item(n).toElement();
-                        coords << QPointF();
-                        QPointF &point = coords.last();
-                        point.setX(elem.attribute("lon").toFloat()*DEG_TO_RAD);
-                        point.setY(elem.attribute("lat").toFloat()*DEG_TO_RAD);
+                        *costs = -1;
                     }
                 }
+                break;
             }
         }
     }
--- a/src/qmapshack/gis/rte/router/CRouterOptimization.cpp
+++ b/src/qmapshack/gis/rte/router/CRouterOptimization.cpp
@@ -1,0 +1,320 @@
+/**********************************************************************************************
+    Copyright (C) 2020 Henri Hornburg hrnbg@t-online.de
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+**********************************************************************************************/
+
+#include "CRouterOptimization.h"
+#include <gis/rte/router/CRouterSetup.h>
+#include <GeoMath.h>
+
+CRouterOptimization::CRouterOptimization()
+{
+    routerOptions = CRouterSetup::self().getOptions();
+}
+
+int CRouterOptimization::optimize(SGisLine &line)
+{
+    checkRouter();
+    if(!CRouterSetup::self().hasFastRouting())
+    {
+        return -1;
+    }
+    if(line.length() < 4)
+    {
+        return 0; //There is nothing to optimize
+    }
+
+    //Optimize using air distance and known distances, since this is much faster than routing, especially brouter
+    SGisLine newAirdistanceOrder;
+    SGisLine oldAirdistanceOrder = line;
+    qreal gain = createNextBestOrder(oldAirdistanceOrder, newAirdistanceOrder);
+    while(gain < 0)
+    {
+        oldAirdistanceOrder = newAirdistanceOrder;
+        gain = createNextBestOrder(oldAirdistanceOrder, newAirdistanceOrder);
+    }
+    //Do routing and calculate cost for order found
+    qreal airdistanceOrderCosts = getRealRouteCosts(oldAirdistanceOrder);
+
+    //Do routing and calculate costs of the order the user supplied
+    //cancel the route calculation when the cost of the airdistance order is reached
+    qreal givenOrderCosts = getRealRouteCosts(line, airdistanceOrderCosts);
+    if(givenOrderCosts < 0 && airdistanceOrderCosts < 0)
+    {
+        return -1;
+    }
+
+    //determine starting order for optimization
+    qreal bestCosts = NOINT;
+    if(givenOrderCosts > 0 && givenOrderCosts < airdistanceOrderCosts)
+    {
+        bestCosts = givenOrderCosts;
+    }
+    else
+    {
+        bestCosts = airdistanceOrderCosts;
+        //the old order is the "optimal" one, since the new one didn't have a gain<0
+        line = oldAirdistanceOrder;
+        fillSubPts(line);
+    }
+
+    SGisLine lastWorkingOrder = line;
+    qreal lastWorkingOrderCosts = bestCosts;
+    int numOfRestarts = 0;
+    // The number of needed starting permutations is somewhat arbitrary,
+    // but you'd likely need more to find the global optimum if there are more possibilities
+    while(numOfRestarts < line.length())
+    {
+        SGisLine newWorkingOrder;
+        qreal bestInsertionGain = createNextBestOrder(lastWorkingOrder, newWorkingOrder);
+
+        if(bestInsertionGain < 0)
+        {
+            qreal newWorkingOrderCosts = getRealRouteCosts(newWorkingOrder, lastWorkingOrderCosts);
+
+            if(newWorkingOrderCosts > 0 && newWorkingOrderCosts < lastWorkingOrderCosts)
+            {
+                lastWorkingOrder = newWorkingOrder;
+                lastWorkingOrderCosts = newWorkingOrderCosts;
+
+                if(newWorkingOrderCosts < bestCosts)
+                {
+                    bestCosts = newWorkingOrderCosts;
+                    line = newWorkingOrder;
+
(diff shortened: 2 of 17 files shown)


pingurus:
One idea to speed up the Brouter part that came to my mind would be to request a geojson from brouter for the whole route at once and then add up the costs that are returned streetwise for each segment. However that would be loads of work and I am not sure to what extent it would speed the whole thing up (The difference for calculating the whole route for one of my benchmarks with 17 segments at once vs. all by themselves was about factor 3), since the gain will be significantly smaller when the difference is between one and two needed requests.

pingurus:
> 
> 
> One idea to speed up the Brouter part that came to my mind would be to request a geojson from brouter for the whole route at once and then add up the costs that are returned streetwise for each segment. However that would be loads of work and I am not sure to what extent it would speed the whole thing up (The difference for calculating the whole route for one of my benchmarks with 17 segments at once vs. all by themselves was about factor 3), since the gain will be significantly smaller when the difference is between one and two needed requests.

Spent much time trying to do it, the code grew too long so I dismissed the idea.

I found a stupid mistake which speeds up the finding of the route significantly.

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -413,6 +413,33 @@ int CRouterBRouter::synchronousRequest(const QVector<QPointF> &points, const QLi
...
+                                    *costs = rxAscDes.cap(9).toDouble(&ok);
+                                    if(!ok)
+                                    {
+                                        *costs = -1;
+                                    }
+                                }
+                                break;
+                            }
+                        }
+                    }
```
kiozen, line 442: Way too many nested code. Try to avoid that many hierarchies by escape logic or subfunctions.

kiozen:
Remove the WIP in the title once you are done and ready for a review.

kiozen:
Damnit closing a PR by accident the second time today. Sorry.

pingurus:
@kiozen When the route to optimize is a new route, the messages set via `canvas->reportStatus(key.item, tr("<b>Started Optimization.</b><br/>"));` appear before the route metrics, if it is a route that is edited, it appears after the route metrics. Is there any way to prevent this?

kiozen:
see 'void CCanvas::reportStatus(const QString& key, const QString& msg)'. Status messages are sorted by their keys.

pingurus:
> 
> 
> see 'void CCanvas::reportStatus(const QString& key, const QString& msg)'. Status messages are sorted by their keys.

Thanks!

src/qmapshack/gis/rte/router/CRouterOptimization.cpp
```diff
@@ -0,0 +1,322 @@
...
+    {
+        return -1;
+    }
+    if(line.length() < 4)
+    {
+        return 0; //There is nothing to optimize
+    }
+
+    CProgressDialog progress(CProgressDialog::tr("Optimizing route"), 0, line.length(), nullptr);
+    progress.setWindowFlag(Qt::Tool, true);
```
pingurus, line 60: @kiozen, is there a way to keep the CProgressDialog in front? This doesn't do it.

kiozen: No as it is a non-modal dialog. (or was it modal?) Anyways CProgressDialog sucks big time.

src/qmapshack/helpers/CProgressDialog.cpp
```diff
@@ -30,14 +30,16 @@ QStack<CProgressDialog*> CProgressDialog::stackSelf;
 CProgressDialog::CProgressDialog(const QString text, int min, int max, QWidget *parent)
     : QDialog(parent)
 {
-    if(!stackSelf.isEmpty())
```
pingurus, line 33: @kiozen what is the neccessity of this stack? When removing it displaying the progress works fine except the minor drawback (not sure if it really is a drawback) that there are progresses appearing above each other.

kiozen: The stack was introduced as everything else irritated users.

pingurus: I'm not sure wh it irritates users if there is the optimization progress bar and sometimes a routing progress appears above it. I must say I prefer if a program shows a bit too much that it is doing something. I understand that you don't want to have 100 progresses above each other on the screen, but since closing one screen will reactivate the underlyng one it doesn't change anything for me.

What I think the reason is that the bar is not properly shown is that a new bar immediately disables the old upper one but only appears one second later. So if there are routing progress bars appearing every 0.5s no progress bar will be shown.

Do you know if Progress bars are created from different threads and/or on the heap so construction order is not destruction order?

kiozen: In the beginning I used QProgressBar and we had several progress bars on the screen. On some window managers the blocked ones where overlapping the active ones blocking the active cancel buttons. I guess that is the reason for the stack. If I recall right :)

The progress bar is a design flaw. Qt has a cooperative multitasking. For GUI related stuff there can be only one thread. The main thread. The rule is to have no method that takes very long. By that the event loop is operated frequently and everything works nicely in a serialized way without blocking. 

For lengthy operations you are in trouble. Either you execute them in a thread. In that case you have to use events or scheduled signals to trigger feedback in the GUI. Additionally you have to take much care to serialize data access.

Or you block Qt. In that case you need the feedback of the progress bar. But to show the progress dialog and to process mouse events (click cancel) you need to process the event loop. And that is the moment where shit hits the fan. If the dialog would have it's own eco system everything would be fine. But it relies on the main event loop. Thus calling the event loop introduces re-entrancy. Something you do not take into account when writing software with Qt as everything is serialized. 

Now if you have several progress dialogs that are not independent, e.g. overall routing progress is interrupted by finding alternatives cancelling the overall progress dialog does not necessarily cancel the other ones. Therefore showing only one is the way to go.

This complete progress bar system is crap. But I did not find a better solution.

pingurus: I now moved the pausing of the old top progress bar to the moment the new one appears and the optimization one stills shows nicely and we only have one bar at once.

I changed the modality of the progress bar to application modal instead of window modal, as I don't want any user input on the element except of the cancel button. Are there use-cases in which the modality is only to stay on-top (window modal with nullptr parent) or the where user input shpuld only be blocked only for the parent and grandparent windows (window modal with valid parent)? I assume you can't exclude it, but if you know that it is the case then I'll find a solution in which the behaviour of the old code stays the same regarding modality instead of looking through the code to check every single use :wink:

kiozen: Oh, good catch. If the progress bar dialog is visible no other input except the cancel button should be possible.

pingurus:
@kiozen @wthaemelt Could you please test?

kiozen:
Compiles. I will have a closer look when time permits.

wthaem:
Tested it with Routino and BRouter and 6 points. Results ok.

What is the stop criterion in your algorithm? Is there a way to stop long-running optimizations with large numbers of routing points and is then the best known route shown?

There is one weak point: The "Optimize" button needs the route edit mode. It is located next to the "O,A,V,T" buttons. If I select say "O", then I select a certain type of routing (when editing the route!). But the optimization doesn't care about this setting. It is using the settings in the routing tab (BRouter/Routino settings). This is a bit confusing.

Therefore, I think a tooltip should be added to the new button informing that the "external" routing engine and not the "O,V,T" methods are used for calculations.

And yet another small issue: The route to be optimized obviously must be a closed one (start = end). The algorithm doesn't work (=didn't work for me) with an "open" route (start != end). Would it be possible to make the optimize button inactive in the case of an open route? And mention the closed route requirement in the tooltip?

pingurus:
> What is the stop criterion in your algorithm? Is there a way to stop long-running optimizations with large numbers of routing points and is then the best known route shown?

The inner stop-criterion is the local optimum when trying to move a single point in the order of the waypoints. The outer stop-criterion is that it performed n 2-opt steps with n being the number of points in the route. When you press cancel on the optimization progress bar the optimization should stop and set the currently best route found.

> And yet another small issue: The route to be optimized obviously must be a closed one (start = end). The algorithm doesn't work (=didn't work for me) with an "open" route (start != end). Would it be possible to make the optimize button inactive in the case of an open route? And mention the closed route requirement in the tooltip?

Interesting, I tested with an open route and it worked fine. What didn't work for you? Was the route returned bad or did it fail in some other way? Keep in mind that the start and the end points are fixed and you thus need to have at least 4 waypoints to optimize anything.

> There is one weak point: The "Optimize" button needs the route edit mode. It is located next to the "O,A,V,T" buttons. If I select say "O", then I select a certain type of routing (when editing the route!). But the optimization doesn't care about this setting. It is using the settings in the routing tab (BRouter/Routino settings). This is a bit confusing.

> Therefore, I think a tooltip should be added to the new button informing that the "external" routing engine and not the "O,V,T" methods are used for calculations.

Good point.

wthaem:
> Interesting, I tested with an open route and it worked fine. What didn't work for you? Was the route returned bad or did it fail in some other way? Keep in mind that the start and the end points are fixed and you thus need to have at least 4 waypoints to optimize anything.

Most likely the example I tested wasn't the best one. Now I took another example of an open route and optimization worked fine. My examples are quite small. Thus, optimization is very quick and no progress bar comes to the GUI.

wthaem:
To make the background of the optimize button clearer add somehow "TSP" to the button label.
Proposals: Solve/Optimize (as) TSP

src/qmapshack/gis/rte/router/CRouterOptimization.cpp
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+    return 0;
+}
+
+void CRouterOptimization::checkRouter()
+{
+    if(routerOptions != CRouterSetup::self().getOptions())
+    {
+        routingCache = QMap<QString, QMap<QString, routing_cache_item_t> >();
```
kiozen, line 328: `routingCache = {};` much easier to read

kiozen: or even `routingCache.clear();` which is more verbose about the intent.

kiozen: Let's optimize:
```
void CRouterOptimization::checkRouter()
{
    const QString& options = CRouterSetup::self().getOptions();
    if(routerOptions != options)
    {
        routingCache.clear();
        routerOptions = options;
    }
}
```

src/qmapshack/gis/rte/router/CRouterOptimization.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+qreal CRouterOptimization::bestKnownDistance(const IGisLine::point_t& start, const IGisLine::point_t& end)
+{
+    //10 digits after the decimal point in exponential format should be by far enough
+    QString start_key=QString::number(start.coord.x(), 'e', 10) + QString::number(start.coord.y(), 'e', 10);
+    QString end_key=QString::number(end.coord.x(), 'e', 10) + QString::number(end.coord.y(), 'e', 10);
+
+    if(!routingCache.contains(start_key))
+    {
+        routingCache[start_key] = QMap<QString, routing_cache_item_t>();
```
kiozen, line 250: `routingCache[start_key] = {};`

src/qmapshack/gis/rte/router/CRouterOptimization.cpp
```diff
@@ -0,0 +1,331 @@
...
+}
+
+const CRouterOptimization::routing_cache_item_t* CRouterOptimization::getRoute(const QPointF& start, const QPointF& end)
+{
+    QString start_key=QString::number(start.x(), 'e', 10) + QString::number(start.y(), 'e', 10);
+    QString end_key=QString::number(end.x(), 'e', 10) + QString::number(end.y(), 'e', 10);
+
+    if(!routingCache.contains(start_key))
+    {
+        routingCache[start_key] = QMap<QString, routing_cache_item_t>();
```
kiozen, line 282: `routingCache[start_key] = {};`

src/qmapshack/gis/rte/router/CRouterOptimization.cpp
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    if(!routingCache[start_key].contains(end_key))
+    {
+        routing_cache_item_t cacheItem;
+        int response = CRouterSetup::self().calcRoute(start, end, cacheItem.route, &cacheItem.costs);
```
kiozen, line 288: Why not pass a reference to cacheItem?

src/qmapshack/gis/rte/router/CRouterOptimization.cpp
```diff
@@ -0,0 +1,331 @@
...
+{
+    for(int i = 0; i < line.length()-1; i++)
+    {
+        line[i].subpts.clear();
+        const routing_cache_item_t* route= getRoute(line[i].coord, line[i+1].coord);
+        if(route == nullptr)
+        {
+            return -1;
+        }
+        for(QPointF point : route->route)
```
kiozen, line 316: `const QPointF& point` avoid copies wherever possible.

src/qmapshack/gis/rte/router/CRouterRoutino.cpp
```diff
@@ -444,6 +444,20 @@ int CRouterRoutino::calcRoute(const QPointF& p1, const QPointF& p2, QPolygonF& c
...
+                    {
+                        // ROUTINO_ROUTE_SHORTEST
+                        *costs = next->dist;
+                    }
+                    else if(comboMode->currentIndex() == 1)
+                    {
+                        // ROUTINO_ROUTE_QUICKEST
+                        // This works, since CRouteOptimization adapts it's weights according to the data it gets
+                        *costs = next->time;
+                    }
```
kiozen, line 459: Please handle the `else` path, too.

src/qmapshack/mouse/line/IMouseEditLine.cpp
```diff
@@ -350,6 +353,38 @@ void IMouseEditLine::slotTrackRouting()
...
+    }
+    catch(const QString &msg)
+    {
+        response = -1;
+        lineOp->showRoutingErrorMessage(msg);
+    }
+
+    if(response == -1)
+    {
+        canvas->reportStatus("Optimization", tr("<b>Optimization failed.</b><br/><b>Note:</b> The selected router must be able to route on-the-fly. Offline routers usually can do, online routers can't.<br/>"));
```
kiozen, line 374: For the sanity of the translators we should avoid html tags in the translation macros. Makes their life much easier. Have a look at `CMouseRangeTrk.cpp` line 45 ff. 

Please adjust all other strings, too.

src/qmapshack/gis/rte/router/CRouterOptimization.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+**********************************************************************************************/
+
+#include "CRouterOptimization.h"
+#include <gis/rte/router/CRouterSetup.h>
+#include <GeoMath.h>
+#include <helpers/CProgressDialog.h>
+#include <QDebug>
+#include <QElapsedTimer>
+#include <QRandomGenerator>
```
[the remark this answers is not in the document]
pingurus: Already gone.

src/qmapshack/gis/rte/router/CRouterOptimization.cpp
pingurus: I wanted to avoid to duplicate the code in `calcRoute` for the passing `routing_cache_item_t` directly.

kiozen: ok

pingurus:
> To make the background of the optimize button clearer add somehow "TSP" to the button label.
> Proposals: Solve/Optimize (as) TSP

I am not sure how well the abbreviation TSP is known to the general public and googling tsp yields mostly results for the "thrift savings plan". I think it's more useful if the button is explained in the wiki.

wthaem:
The tooltip of the button should/could decode the abbreviation - my favorite approach (and this fits into kiozen's philosophy to use tooltips quite extensively and only then the Wiki).

My issue is simply that "optimize" alone can be misleading. 

The object under consideration is a route. And "optimizing a route" means for me first of all: Let the given routing points be fixed and find better paths between the routing points using one or the other optimziation criterion (surface quality, length, ...). But that is exactly, what the routing engine should have done when creating the route.

But here it means something different and not obvious: re-order the routing points to get a better route.

Maybe this label could do: "Reorder points". The essential and interesting part "optimization" is lost here - but the label must be short.

kiozen:
I think `optimize` is fine. But yes make a detailed tooltip. Like 

This will optimize your route by reordering the route points. This is usually known as solving the "Traveling Salesman Problem".

pingurus:
How about:

> Optimizes the route by reordering the points using the router specified in the "Routing" tab. 
> This is known as solving the Traveling Salesman Problem, however start and end are kept fixed here.

wthaem:
My variant:

Optimize the route by reordering the route points (except start and end points that are kept fixed) using the router specified in the "Routing" tab (solve so-called Traveling Salesman Problem).

src/qmapshack/mouse/line/IMouseEditLine.cpp
```diff
@@ -350,6 +353,41 @@ void IMouseEditLine::slotTrackRouting()
...
+    if(response == -1)
+    {
+        canvas->reportStatus("Optimization", QString("<b>%1</b><br/><b>%2</b> %3<br/>")
+                             .arg(tr("Optimization failed."))
+                             .arg(tr("Note:"))
+                             .arg(tr("The selected router must be able to route on-the-fly. Offline routers usually can do, online routers can't.")));
+    }
+    else
+    {
+        canvas->reportStatus("Optimization", tr("<b>%1</b><br/>").arg(tr("Optimization successful.")));
```
kiozen, line 381: no tr macro on the first part with the html tags

src/qmapshack/mouse/line/IScrOptEditLine.ui
```diff
@@ -304,6 +304,16 @@
        </layout>
       </widget>
      </item>
+     <item>
+      <widget class="QPushButton" name="pushOptimize">
+       <property name="toolTip">
+        <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Optimizes the route by reordering the points using the router specified in the &amp;quot;Routing&amp;quot; tab. &lt;/p&gt;&lt;p&gt;This is known as solving the Traveling Salesman Problem, however start and end are kept fixed here&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
kiozen, line 310: You took the wrong editor. There is one for html and one for plain text. Please use the one for plain text.

If you want extended formatting create a string in the cpp file and add it as tooltip.

kiozen:
If everything is done and everyone content I would merge the PR.

pingurus:
For me everything is fine.

wthaem:
Go ahead!

Issue #134, opened 2020-03-26
Rework track range selection from scratch


kiozen:
**Is your feature request related to a problem? Please describe.**
The range selection can be used to hide/show/copy track point and to partially apply activities. Especial for the last task, applying activities, the current range selection can be quite cumbersome. Especially for looping tracks it is quite hard to hit the wanted point in the track. 

**Describe the solution you'd like**
On range selection an on-screen dialog should open at the bottom of the canvas. The track profile and the table of track points should be shown. There should be buttons to move the start and the end of a selected range. Point by point on a single click. Repetitive when holding down the button.

There are also buttons for the usual range selection operations. This tool won't close until the user releases it by a right click on the canvas. Therefore several operation on several ranges can be done in a row.

**Describe alternatives you've considered**
Alternatives could be placing it into the track edit dialog. -> already too crowded. 

**Additional context**
That's how the new on-screen tool might look like:

![Screenshot_20200326_153744](https://user-images.githubusercontent.com/54329441/77660938-4145ad00-6f7a-11ea-8523-b6773401588e.png)

PR #144, opened 2020-04-05
[QMS-134] Rework track range selection from scratch

kiozen:
**What is the linked issue for this pull request (start with a `#`):** QMS-#134

I added a complete new on-screen tool for range selection on a canvas. Selecting a range in a plot is not affected by this pull request.

Features:
* You can select the range on the canvas
* You can select the range in the plot with the range tool
* You can use buttons to move the start and and of a range
* You can also pick up again start and end of a range on the canvas.
* You can start a new range at the start and the end of selected range for seamless range selection
* You can delete track points (not just hide them)

Activate range selection for a track on the canvas. Refer to tool tips for help.

- [x] yes

- [x] yes

- [x] yes,
- [ ] no


kiozen:
@wthaemelt & @pingurus This is still work in progress but you might want to start test and review.

pingurus:
Nice feature!

Things that came to my mind when playing around:

* Make the vertical bar of this butten thicker: 
grafik
* When moving the mouse to the on screen display section it happens, that the table jumps to a point you've been passing over. A button to view the start or the end of the selection in the table (just what happens if you press back and then forward)
* On my screen the position is not completely visible, even though there is enough space on screen to make the table wider
* When a track has no elevation data, the NOINT (Or whatever the values is set to if there is no elevation) is shown. A '-' just as for slope etc. would be better
* If I am in 'select range' for a track and then select another track by double clicking another track, the 'select range' overlay gets grayed out with the other track being marked as selected on the map. I need to right-click into the map to remove the window to continue without the unusable overlay. It would be nice if it would be auto-closing or opening the range tool for the other track.
* Would it be somehow possible to extend the selected range without the buttons after one did the initial selection? For example if one clicks the start ot endpoint of the range, the range selection could be picked up from that point.

kiozen:
* I a not happy with the icons either.
* Don't think that a button is necessary. The list simply follows the focus.
* Increased width
* Fixed bug for NOINT. Nice catch.
* Reset mouse to normal when double clicking an item. Nice catch.
* You can extend the range already. Just place the mouse close enough to the ends until they get highlighted and click.

wthaem:
(Tested without the latest commit!)

* helpful tool, but in the current state still with some confusing features!
* last part of position column cut-off - depends on the setup of map font - the bigger the font, the more is cut-off. If range info part is displayed, then position field not visible at all. My conflict: I like relatively a large font for POI names on the map. This leads to a (too) large font in the range window point list. 

    (From pingurus): On my screen the position is not completely visible, even though there is enough space on screen to make the table wider
* list of trackpoints appears only after moving the mouse to the range window
* (From pingurus): Make the vertical bar of this butten thicker - WT: yes, please
* Control part with 6 arrows: a comment/label for each 3-arrow-group giving the meaning of the group would be nice **or** avoid equal icons with differing meaning.
* (from pingurus): If I am in 'select range' for a track and then select another track by double clicking another track, the 'select range' overlay gets grayed out with the other track being marked as selected on the map. I need to right-click into the map to remove the window to continue without the unusable overlay. It would be nice if it would be auto-closing or opening the range tool for the other track.

  This is a confusing situation: The icon on the map shows "range selection" - but range can't be selected in map for the selected track. Range window is grey and shows only the controls and nothing else. A right click on the map brings the cursor back to the default one, the range window is closed. A click on the map (not on the track!) removes the selection of the track. Selecting the track again and choosing the range icon re-opens the grey range window. There is no obvious way to stop the (hidden) still active former range selection. If you don't remember the previously handled track you are lost. And even if you do, the "stop range" in the track edit window doesn't work - range wasn't selected in the edit window (similar situation in edit window with 3 graphs: select range in one graph, try to stop range in other graph - doesn't work, stop range must be done their, where it was started).
* "Jumping maps": While selecting the range the map jumps permanently to a new position (centers the last selected range point?). Part of the track is covered by the range window, whereas at the start my zoom was so that I had a nice look at my whole track.
* Something is going wrong with the action icons "reset, hide, show". Clicking them, they get grey. After a while (moving the mouse?) they get their color back.
* There seems to be a serious timing or refresh problem: Even for a very short track it takes ages to see the complete and colored range window after e.g. a "hide points" action. You get and see the grey range window, wait a while and then think: Ok - now I got stuck. But after a considerable while colors come back.
* (from pingurus): When a track has no elevation data, the NOINT (Or whatever the values is set to if there is no elevation) is shown. 

  WT: This is different to the behavior of the point list in the track edit window - there elevation is empty!

pingurus:
grafik
Maybe this kind of icon (sorry for the paint art :laughing: ) would be better instead of the following icon (I think it somehow better depicts what the button does):
grafik

pingurus:
Looks great now!
I don't know if you tried to fix it, but the positions is still not completely shown.

kiozen:
I let the widget expand a bit more.If that is not sufficient you have to find a solution yourself and send a patch.

pingurus:
How about moving the CScrOptToolRange 100px to the left and allowing it to expand another 100px? In that case the scale would also be shown completely on more screens.

pingurus:
When toying around with the map size to see if I get to see the whole  coordinate column I ended up with this kind of screen (to reproduce make the left sidebar really wide, open the on screen options and then make them normal again, it won't change that the ScrOpt is behind the sidebar):
grafik
The position of the ScrOpt seems to be fixed relative to the left side of the map and not be calculated when the map size changes.

On the other hand, if I make the map the only view on my screen, it takes much space filling it with nothing but white:
grafik

Sure, these things are more or less an aesthetical problem, but I still think it might be worth looking for a solution.

wthaem:
* Much better now
* Instead of (already good) "start/end" I'd prefer something like "(move )towards start/end". Enough space is there.
* It is hard to distinguish the active and the inactive state of the arrow icons. The black color is very close to the grey color and the green below the arrows is only a tiny spot, too.
* Click "Hide" action icon and leave the mouse pointer on the icon. All action icons remain grey. Move mouse to point list - icons still grey. Move mouse to profile graph - icons get their colors back. Is this the wanted behavior?
* Goto fullscreen with open range window and then back - some lines of the window get lost (not always!).

LostLines

* What should happen with the range window in the case that the track is in the edit mode and a range should be selected in this case? There is no range window. Only the old 2 icons for "Delete ..." and "Calculate ...".
* Some "Beifang":

Beifang

kiozen:
> Click "Hide" action icon and leave the mouse pointer on the icon. All action icons remain grey. Move mouse to point list - icons still grey. Move mouse to profile graph - icons get their colors back. Is this the wanted behavior?

Sure. Moving it to the graph you selected a single track point. All buttons that apply if a single trackpoint is selected are enabled. Works as intended.

> What should happen with the range window in the case that the track is in the edit mode and a range should be selected in this case? There is no range window. Only the old 2 icons for "Delete ..." and "Calculate ...".

Edit mode has nothing to do with range selection. These are two completely different use-cases with different requirements.

> Some "Beifang":

yes I know. Has been like this all the time. Never found a good solution and it does not really hurt.

> It is hard to distinguish the active and the inactive state of the arrow icons. The black color is very close to the grey color and the green below the arrows is only a tiny spot, too.

Icons can be done by everyone. Feel free to provide a better icon set. That said so far no one accepted that offer :wink: 


I made a lot of tiny changes  the last days. The feature is now more or less in a good shape. Please test.

pingurus:
Is it intended that when I start a range selection in the "Edit Details" screen and then switch to the map there are no on-screen options? The selected range is marked on the map.

When doing it the other way around (Open range selection on the map and then switching over to the "Edit Details") the option "stop range" in the context menu of the diagrams is disabled.

wthaem:
Thanks for the explanations!

And attached  is my offer for slightly improved icons (thicker arrows, one instead of two colors in circles).

[range_icons_svg.zip](https://github.com/Maproom/qmapshack/files/4462249/range_icons_svg.zip)

kiozen:
Thanks for the help. But the new icons are not really convicing:

Screenshot_20200410_194940

The different thickness of the arrows is irritating. And the bullets without border bleed. Maybe using dark green as bullet color would help. And the reset icon has the old pattern.

You should be able to use bash as shell n Windows 10. If it works you can try the makeicon script with an svg file as parameter. That will place the png in the subpaths. By that you can easily verify the final result.

wthaem:
Next attempt (updated reset icon, green ===> dark green, thicker curved arrows). Don't believe that I can do better.

Checked in QMS. Bash script worked like a charm!

[range_icons_svg1.zip](https://github.com/Maproom/qmapshack/files/4462979/range_icons_svg1.zip)

kiozen:
Reworked the icons again. Sticked to the lighter green but applied a frame. Unified all elements.

kiozen:
> Is it intended that when I start a range selection in the "Edit Details" screen and then switch to the map there are no on-screen options? The selected range is marked on the map.

> When doing it the other way around (Open range selection on the map and then switching over to the "Edit Details") the option "stop range" in the context menu of the diagrams is disabled.

@pingurus  Yes this is intended behavior. It's either Track Details Dialog or the Range Selection on the map. To let the user use both in parallel leads to a heap of corner cases no one want's to handle.


I just noticed the track documentation is hopelessly outdated. 

https://github.com/Maproom/qmapshack/wiki/DocGisItemsTrk2

@wthaem If you feel like it you are welcome to update it. But it's quite a bunch of work.

wthaem:
> I just noticed the track documentation is hopelessly outdated.
> 
> https://github.com/Maproom/qmapshack/wiki/DocGisItemsTrk2
> 
> @wthaem If you feel like it you are welcome to update it. But it's quite a bunch of work.

And AdvTrkGeneral is outdated as well.

I'll try a revision of these pages.

BTW: Thanks for the "Jumping maps" button!

Issue #132, opened 2020-03-22
DEMs: Make the scale range control cut off not only rendering but also loading or requesting the data.


mitxel-m:
**Is your feature request related to a problem? Please describe.**

- For maps: You can control visibility with the scale range controls. Consequently, QMS does not load or request the map out of that range.

- For DEMS: you can control the visibility of hillshading, slopes, and elevation limit with the scale range control. However, even if you limit a scale range, QMS still loads/requests the elevation data for all the canvas area.

This leads to slow down the loading of the map when zooming out, and to avoid it the only remedy is to disable the DEM. Then, after scrolling the map and zooming in, re-activate it, which is tedious and illogical.

**Describe the solution you'd like**

Make the scale range controls work identically for Maps and DEMs.
More specifically, for DEMs make the scale range controls cut off not only rendering but also loading or requesting the data.

**Describe alternatives you've considered**
I honestly haven't found any

**Additional context**

One might want to cut the hillshading but still see the elevation data. However, this makes no sense if you make a very outer zoom with a very dense DEM, because at a distant zoom level you don't have enough precision with the mouse to know what point you are pointing at.

It makes more sense to work as with maps, loading two DEMS with different resolution, one on top of the other, and control which one will be used with the scale range controls. This is really useful to hillshade garmin maps, but for now you are requesting all the data. 

For elevation profiles and status line, QMS gives us the data of the first one it finds, so the logical thing is to place the densest one on top, then another one more coarse.
 
But to make this work smoothly it is essential to limit the data load with the scale range slider.

mitxel-m:
**A bit more of context**

For now, the only effect of the scale range is on the display of hillsade, slopes and elevation limit. Even if we are outside the scale range, QMS always loads/request elevation data, regardless of whether the scale range is enabled or not.

This is good for drawing tracks and waypoints, but has the downside that moving and zooming out on the map slows everything down.

Another problem is that the DEM data request is not visible, this leads some users to think that QMS is very slow loading maps, not realizing that the problem is that they have activated a too dense DEM for the zoom level they are using.

**Alternatives to handle this in the current version:**

- Deactivate/Reactivate the DEM according to the zoom level, which is a bit boring, and then we have already lost the advantage we had to get elevation data while drawing a track.

[...]

PR #155, opened 2020-04-17
[QMS-132] Do not load DEM data out of scale

kiozen:
**What is the linked issue for this pull request (start with a `#`):** QMS-#132

Avoid to call GDALDataset::RasterIO() if current scale is out of  valid scale. To achieve that the out of scale information derived during CDemVRT::draw() has to be stored and used in getElevationAt() getSlopeAt().

* Load  DEM
* Set the scale bounds
* Out of the scale bounds no elevation data should be shown in the status bar

- [x] yes

- [x] yes

- [x] yes,
- [ ] no

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,10 +1,12 @@
 V1.XX.X
 [QMS-68] Route Optimization
+[QMS-132] DEMs: Make the scale range control cut off not only rendering but also loading or requesting the data.
 [QMS-134] Rework track range selection from scratch
 [QMS-135] Improved French translation
 [QMS-140] Enhancements for OS X release
 [QMS-148] Misplaced track info window
 [QMS-151] Missing tooltip in route toolbar
+[QMS-153] Improved French translation
 
 V1.14.1
 [QMS-37] Inconsistent use of time zones
--- a/src/qmapshack/canvas/CCanvas.cpp
+++ b/src/qmapshack/canvas/CCanvas.cpp
@@ -648,8 +648,8 @@ void CCanvas::mouseMoveEvent(QMouseEvent * e)
 {
     QPointF pos = e->pos();
     map->convertPx2Rad(pos);
-    qreal ele = dem->getElevationAt(pos);
-    qreal slope = dem->getSlopeAt(pos);
+    qreal ele = dem->getElevationAt(pos, true);
+    qreal slope = dem->getSlopeAt(pos, true);
     emit sigMousePosition(pos * RAD_TO_DEG, ele, slope);
 
     mouse->mouseMoveEvent(e);
--- a/src/qmapshack/dem/CDemDraw.cpp
+++ b/src/qmapshack/dem/CDemDraw.cpp
@@ -268,7 +268,7 @@ void CDemDraw::restoreActiveMapsList(const QStringList& keys, QSettings& cfg)
 }
 
 
-qreal CDemDraw::getElevationAt(const QPointF& pos)
+qreal CDemDraw::getElevationAt(const QPointF& pos, bool checkScale)
 {
     qreal ele = NOFLOAT;
     if(CDemItem::mutexActiveDems.tryLock())
@@ -287,7 +287,7 @@ qreal CDemDraw::getElevationAt(const QPointF& pos)
                     break;
                 }
 
-                ele = item->demfile->getElevationAt(pos);
+                ele = item->demfile->getElevationAt(pos, checkScale);
                 if(ele != NOFLOAT)
                 {
                     break;
@@ -299,7 +299,7 @@ qreal CDemDraw::getElevationAt(const QPointF& pos)
     return ele;
 }
 
-qreal CDemDraw::getSlopeAt(const QPointF& pos)
+qreal CDemDraw::getSlopeAt(const QPointF& pos, bool checkScale)
 {
     qreal slope = NOFLOAT;
     if(CDemItem::mutexActiveDems.tryLock())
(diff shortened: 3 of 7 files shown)


kiozen:
@mitxel_m please test.

mitxel-m:
**I have tested it and it works fine, as expected**

 
#### What have I tested 
I have tested with three simultaneous DEMs, one on top of the other, and applying scale ranges, as shown in the image:
 
DEM_scale_range

 
1. On top: ESP_IGN-MDT05.wcs  /  online DEM 5x5m
2. Then: ESP_IGN-MDT25.vrt       / Local DEM 25x25m - Tiled Tiff format
3. Then: WORLD_SRTM90.wcs    / Online coarse DEM 90x90m
 
### After testing:
 
It works fine and much faster than before.
Obviously the user has to organize well these scale ranges, but it's the same as for the maps, so it's coherent and easy to understand.
 
One point that could still be improved is how this affects the drawing of tracks, and  DEM plots.
 
With this change the track drawing takes the elevation data from the DEM that it finds in the scale range you are drawing. To be more precise in the scale range you are when you finish the track drawing. That is, if you draw at outer zooms, it will take worse data than if you draw at closer zoom, 
 
For the drawing tracks and dem plots it would be better to always take  the data from the first dem on top,regardless the sacale ranges
 
If it is possible to code,  it  would be  a simple and  easy  rule to understand by the users, (I don't know if it's that easy to code) .

In fact it is just the same behavior we have followed for now. And a request for all the points that usually integrate a track is not an excessive request, even for an online DEM. Therefore it should not slow down the workflow
 
When drawing tracks it makes sense because we always would like the best elevation data. For DEM plots it would be to avoid strange plots, like the following one:
 

 DEM_plot

In this image,with the current PR: the track has been drawn in a close zoom and h, thus it has taken the data of the finest DEM 5x5m. We zoom out ( to a country level) in the map view and  then we click on the track profile. At this point with this PR the DEM plot (red line) takes the data from the SRTM90 (90x90m) (obviously much smoother). This could be confusing. It could be solved by applying the rule of always taking the data for tracks and DEM plots from the first dem on top for tracks and plots.
 
The third element affected would be the status line, that shows  elevation data at the pointer position . I dont find this so important as others, but may be the same rule coud be applied .  Honestly I am not  sure  if it is  possible  to only request the data for the pointer position ( like for track points) deppending on the DEM format,   The actual behavoir is  that  we are requesting data for the whole canvas, that in some cases implies requesting complete lines of the DEM ( even out of the canvas),  or  blocks if the DEM is an optimized tiled DEM.

### Summary:
 
1-  The current PR, that cuts off the data request according to the scale ranges,   works fine and optimizes the map rendering. IMHO a must have
 
2-  For the drawing tracks, wp,  and dem plots, if possible :taking always the data from the first dem on top,regardless the sacale ranges, would be excellent.
 
3- For the status line  I am not so sure which could be the best. If possible follow the rule as for tracks and DEM plots.

kiozen:
Hmmmm... Let's see what we had:

* Drawing DEM stuff(hill shade, slopes etc.) was limited by the scale all time.
* Reading the elevation or slope at a position was not limited by the scale.

What we have now:

* Drawing DEM stuff(hill shade, slopes etc.) is still limited by the scale.
* Reading the elevation or slope at a position is limited by the scale.

The elevation/slop is read at various places. Creating a new waypoint, drawing a track, plotting track data, etc. It seems to me that for all these use cases the user wants the best resolution. Right? So it would be the top most DEM file in the list.

What's left? As far as I can see just the status bar. Could it be that this is the only thing that slows down something? And should not select the top most DEM file always. Just the top most one with a matching scale?

mitxel-m:
>  The elevation/slop is read at various places. Creating a new waypoint, drawing a track, plotting track data, etc. It seems to me that for all these use cases the user wants the best resolution. Right? So it would be the top most DEM file in the list.

Right. 
In the image above it would be  `ESP_IGN-MDT05.wcs`  

> What's left? As far as I can see just the status bar. Could it be that this is the only thing that slows down something? 

Yes, you can observe how as you move the cursor over the map you get the  elevation data. However, when you move to another area further away there is a small pause and not only the cursor position, but a whole block of data has to be loaded around the cursor before it reacts.  This is almost unappreciable when the resolution of the DEM is adequate for the zoom scale, but is annoying when there is a high resolution DEM on an outer zoom level, because too much data is requested and everything slows down

> And should not select the top most DEM file always. Just the top most one with a matching scale?

That would be the fastest. No doubt. For me it is Ok.
 
It doesn't make sense to aim for 5x5m accuracy when one pixel on the screen is equivalent to 100x100m. In fact it is the same for coords, but I'm not sure if all users would understand this. 

May be a warning on the status line would be adequate to remind the user. For example, prefix the elevation data with (!) when the data is not read from the top most DEM file in the list. Something like this:

`N45.12345º E0.12345º Elevation: (!) 1234m  Slope: 27.1º`

kiozen:
I limited scale check just to be used by the status bar. All other requests use the topmost active DEM file. Please test again.

mitxel-m:
__Tested succesfully__

- Drawing tracks, wp, profiles, ... always take valid data from the most top DEM, even out of the scale range ->OK
- Status bar shows elevation data for the first DEM within the scale range. It doesn't show any out of the scale range. ->OK
- Hillsade, slope, and elevation limit, is only rendered within the sacale range ->Ok 

#### Use of contiguos DEMs.
 
Use case for DEMs with different SRS and/or overlapping at country borders. View image below

contiguos_DEMs

Top most DEM (pink tinted) follows country borders, so there is a gap with no data. This DEM file has nodata value well defined. When drawing a track at this gap QMS takes the elevation data from the next DEM with valid data, even if this second one is out of the scale range.  -> OK!! 

#### Finally:
I find all this very consistent :-)

kiozen:
Ok, fine. Thanks for testing.

Issue #164, opened 2020-04-23
Workspace filter is not applied to newly loaded projects


pingurus:
**Describe the bug**
The Workspace filter is not applied to a project that was loaded to the workspace after the filter was typed in.

**To Reproduce**
1. Type a filter to the workspace
2. Load a project that contains some items that should be excluded
3. Check and see that the items that should be excluded are still there

**Expected behavior**
The filter that applies to the workspace should always apply to all projects.

**Screenshots**
No need

**Tracebacks**
No need

**Desktop (please complete the following information):**
 - OS: Windows 8.1
 - QMapShack Version 1.14.1

**Additional context**
If I'm already at it, I will check if I can find some way that the filter is only invoked if some time has passed after the last letter has been typed. Right now it is a bit annoying to filter a list of 1000 Geocaches. (I know @kiozen that you don't think that one should have loaded that many items to the workspace, but there are 1000 Geocaches in my moving radius and I don't want to go to the website all the time to download another small portion :wink:)

kiozen:
:laughing: well yes pocket queries....

PR #167, opened 2020-04-24
[QMS-164] Workspace filter is not applied to newly loaded projects

pingurus:
**What is the linked issue for this pull request (start with a `#`):** QMS-#164

* Fixed non-filtering
* Added timout to start filtering

1. Apply a filter
2. Load file with many Items
3. See that filter is applied
4. Change the filter
5. See that you don't hang after typing single characters if you are typing normally

- [x] yes

- [x] yes

- [x] yes,
- [ ] no

### DIFF
--- a/src/qmapshack/gis/search/CSearchLineEdit.cpp
+++ b/src/qmapshack/gis/search/CSearchLineEdit.cpp
@@ -21,12 +21,19 @@
 #include "gis/search/CSearch.h"
 #include <canvas/CCanvas.h>
 #include <QMenu>
+#include <QTimer>
 
 CSearchExplanationDialog* CSearchLineEdit::explanationDlg = nullptr;
 
 CSearchLineEdit::CSearchLineEdit(QWidget *parent)
     : QLineEdit (parent)
 {
+    searchCreationTimer = new QTimer(this);
+    searchCreationTimer->setSingleShot(true);
+    connect(searchCreationTimer, &QTimer::timeout, this, [this] {
+        slotCreateSearch(text());
+    });
+
     actionClearFilter = new QAction(QIcon(":/icons/32x32/Filter.png"), tr("Clear Filter"), this);
     actionHelp = new QAction(QIcon(":/icons/32x32/CSrcUnknown.png"), tr("Open Help Window"), this);
     actionSetupFilter = new QAction(QIcon(":/icons/32x32/Apply.png"), tr("Setup Filter"), this);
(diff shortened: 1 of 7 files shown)


kiozen:
I tested and it's not working as expected. 

* Loaded a GPX with tracks.
* Set global filter `distance <6km`
* Load a project from database with tracks
* All tracks are shown. And all tracks are longer than 6km

pingurus:
> I tested and it's not working as expected.

How are database projects loaded into the workspace? Obviously not via loadGisProject, but I couldn't find how. For me only those projects don't work.

src/qmapshack/gis/search/CSearchLineEdit.cpp
```diff
@@ -196,3 +202,9 @@ void CSearchLineEdit::slotCreateSearch(const QString& str)
         emit sigWorkspaceSearchChanged(currentSearch);
     }
 }
+
+void CSearchLineEdit::slotTriggerCreateSearch(const QString &str)
+{
+    //One second is pretty long, but it may allow the user a second of thinking before the lineEdit blocks for filtering
+    searchCreationTimer.start(1000);
```
kiozen, line 209: wouldn't this be another candidate for a lambda?

src/qmapshack/gis/search/CSearchLineEdit.h
```diff
@@ -57,6 +59,7 @@ private slots:
     IGisProject * connectedProject = nullptr;
     QTreeWidgetItem * searchItem = nullptr;
 
+    QTimer searchCreationTimer;
```
kiozen, line 62: The usual Qt-ish way is to define a pointer. Allocate the timer with new in the ctor. This  class as parent.

src/qmapshack/gis/search/CSearchLineEdit.h
```diff
@@ -19,6 +19,7 @@
 #define CSEARCHLINEEDIT_H
 
 #include <QLineEdit>
+#include <QTimer>
```
kiozen, line 22: `class QTimer;`

Rule: Never add more headers into a header than barely needed. Avoid header hell.

kiozen:
Done with everything? If yes I merge.

pingurus:
Should I mention in the commit message that I added the new comparator "is"? Maybe update it in the wiki?

I there are any more things I can add, tell me. If not, I think this is everything for now.

kiozen:
Well it would be nicer not to sneak in that stuff. A new ticket and PR would be appropriate.  But as it is really just a minor change it's ok for me.

If there is something in the Wiki please update the wiki.

Issue #172, opened 2020-04-29
Select action "for all" in Database Conflict


pingurus:
**Is your feature request related to a problem? Please describe.**
Some times when I save something to my database, this screen appears:
![grafik](https://user-images.githubusercontent.com/36167579/80570897-eb8e7580-89fb-11ea-80df-1e3879df5132.png)

It is annoying to have to click your choice for every single element that has been changed.

**Describe the solution you'd like**
Just as there is the option in the other conflict dialog to apply selection to all elements, it would be nice to have it here as well.

**Describe alternatives you've considered**
None

**Give a reason why someone else should spend time on that request**
Because everyone has to spend time sitting and clicking once the problem arises. If you are unlucky you can't even abort, since the progress is hidden behind the instantly reappearing selection window.

**Additional context**
What is the difference between the two conflict pop-ups? I tried to find it in the wiki, but couldn't find anything.
I am using a local database only.

PR #181, opened 2020-05-15
[QMS-172] Select action "for all" in Database Conflict

pingurus:
**What is the linked issue for this pull request (start with a `#`):** QMS-#172

* Added seperate dialog class for database conflict with checkbox
* catched error of cloning geocaches

1. Copy two different versions of a geocache into two different database folders
2. Save workspace
3. Select that you want to replace the geocache
4. See in the appearing database conflict that you can choose action for all and that if you provoke multiple errors you will not be asked again.

- [x] yes

- [x] yes

- [x] yes,
- [ ] no

### DIFF
--- a/src/qmapshack/gis/db/CDBProject.cpp
+++ b/src/qmapshack/gis/db/CDBProject.cpp
@@ -20,6 +20,7 @@
 #include "gis/CGisDatabase.h"
 #include "gis/CGisWorkspace.h"
 #include "gis/db/CDBProject.h"
+#include "gis/db/CResolveDatabaseConflict.h"
 #include "gis/db/CSelectSaveAction.h"
 #include "gis/db/IDB.h"
 #include "gis/db/macros.h"
@@ -297,28 +298,8 @@ int CDBProject::checkForAction2(IGisItem * item, quint64 &itemId, QString& hashI
             "your version and take the one from the database"
             ).arg(item->getNameEx()).arg(user).arg(date);
 
-        QMessageBox msgBox(QMessageBox::Question, tr("Conflict with database..."), msg, QMessageBox::NoButton, CMainWindow::self().getBestWidgetForParent());
-        QAbstractButton* pButClone  = msgBox.addButton(tr("Clone && Save"), QMessageBox::YesRole);
-        QAbstractButton* pButForce  = msgBox.addButton(tr("Force Save"),    QMessageBox::ApplyRole);
-        QAbstractButton* pButUpdate = msgBox.addButton(tr("Take remote"),   QMessageBox::DestructiveRole);
-        msgBox.addButton(QMessageBox::Abort);
-
-        CProgressDialog::setAllVisible(false);
-        msgBox.exec();
-        CProgressDialog::setAllVisible(true);
-
-        if(msgBox.clickedButton() == pButClone)
-        {
-            action = eActionClone;
-        }
-        else if(msgBox.clickedButton() == pButForce)
-        {
-            action = eActionUpdate;
-        }
-        else if(msgBox.clickedButton() == pButUpdate)
-        {
-            action = eActionReload;
-        }
+        CResolveDatabaseConflict dialog (msg, item, action2ForAll, CMainWindow::self().getBestWidgetForParent());
+        action = dialog.getAction();
     }
     else
     {
@@ -675,6 +656,8 @@ bool CDBProject::save(int lastResult)
             CProgressDialog::setAllVisible(true);
         }
     }
+    //Reset choice possible choice the user made for all items
+    action2ForAll = eActionNone;
 
     // serialize metadata of project
     QByteArray data;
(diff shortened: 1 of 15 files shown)


src/qmapshack/gis/db/CDBProject.h
```diff
@@ -159,6 +159,7 @@ class CDBProject : public IGisProject
     };
 
     Qt::CheckState checkState = Qt::Unchecked;
+    friend class CDatabaseConflict;
```
kiozen, line 162: Why friend? Just because `enum action_e` is private? Move the enum to public section.

pingurus:
> And I am not that happy with the statics. This leads to weird behavior. If I repeat the action for the very same project the choice of the first attempt is implicitely used. If I switch between projects I get asked each time. There must be a better solution for code and UX.

What I had before was that there occured a manual reset of the default action from code at the end of the iteration over the project in CDBProject. Whilst this resolves the issue of repeating the action for the project, I think it is a bit error prone having to remember to reset the default action if the code ever changes.

Any suggestions how to improve this?

pingurus:
Two ideas:
* save the maximum index of an item in the current project, if it reduces, the action set for all is reset. I am not sure though if we can assume that the items come in the correct order.
* A parameter in the constructor that says wether the action should be reset. Needs more code in CDBProject (well, two lines per usage), but that will no be forgotten if it is required in the constructor. Makes the usage also more flexible.

Your opinion @kiozen ? I am not exactly sure what you meant by 

> There must be a better solution for code

pingurus:
Third option:
* Create the CResolveDatabase outside the loop, and rerun the dialog via getAction(msg, item). The static variables could then be omitted.

pingurus:
I now opted for a similar solution to the one that is used in `checkForAction1`.
The dialog is a bit ugly, since the label does not wrap even though the property is set. Did I forget about something in that respect?

kiozen:
I would prefer to have this in a cleaned up way. Starting with:
```
bool CDBProject::save(CSelectSaveAction::result_e action1ForAll, action_e action2ForAll)
````
By that the private member `action2ForAll` shouldn't be necessary and you do not need that peculiar reset.

**Comment by pingurus** [comment]
Is there a specific reason why there is a `CSelectSaveAction::result_e` and a `CSelectCopyAction::result_e` with exactly the same contents that are being used implicitly synonymously in the current code? Shouldn't those be unified?

[CGisSummaryDropZone.cpp L120](https://github.com/Maproom/qmapshack/blob/af850d954aeb615c23c42ccb8faf2ac4e60202ed/src/qmapshack/gis/summary/CGisSummaryDropZone.cpp#L120) <-> [CDBProject L620](https://github.com/Maproom/qmapshack/blob/af850d954aeb615c23c42ccb8faf2ac4e60202ed/src/qmapshack/gis/db/CDBProject.cpp#L620) <-> [CDBProject L510](https://github.com/Maproom/qmapshack/blob/af850d954aeb615c23c42ccb8faf2ac4e60202ed/src/qmapshack/gis/db/CDBProject.cpp#L510)

I am asking, since I cleaned up to have the `bool CDBProject::save(CSelectSaveAction::result_e action1ForAll, action_e action2ForAll)` and stumbled over several enums being used as ints and specifying the type of those led me there.

**Comment by kiozen** [comment]
A very good question! I guess I can't really answer it precisely. I could think of the following reasons:

1. I duplicated it and forgot to clean up.
2. I thought it's better to have an enumeration  for each of these tasks. By that I can extend these enumerations without taking care of dependencies somewhere else.
3. These enumerations evolved independently at simply match by accident and so far no one noticed.

Point 2 is quite likely because it would match my usual reasoning. But it's always good practice to challenge my decision. So if it doesn't make any sense change it.

**Comment by pingurus** [comment]
I left the two enums seperate, as that was the only point where the were used as the same. I split them up there, since the values is fixed anyways (and if it wouldn't be, the user should see both dialogs seperately).
I hope I cleaned up everything.

File: src/qmapshack/gis/db/CDBProject.cpp [hunk 2]
```diff
@@ -675,6 +656,8 @@ bool CDBProject::save(int lastResult)
             CProgressDialog::setAllVisible(true);
         }
     }
+    //Reset choice possible choice the user made for all items
+    action2ForAll = eActionNone;
```
kiozen, line 660: do we need that reset? I think it has no effect.

src/qmapshack/gis/db/CResolveDatabaseConflict.h
```diff
@@ -0,0 +1,41 @@
...
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+**********************************************************************************************/
+
+#ifndef CDATABASECONFLICT_H
+#define CDATABASECONFLICT_H
+
+#include "gis/db/CDBProject.h"
+#include "ui_IDatabaseConflict.h"
```
kiozen, line 23: wrong header file

src/qmapshack/gis/db/CResolveDatabaseConflict.h
```diff
@@ -0,0 +1,41 @@
...
+#include <QDialog>
+
+class CResolveDatabaseConflict : public QDialog, private Ui_IDatabaseConflict
```
kiozen, line 26: Ui::IResolveDatabaseConflict

PR #198, opened 2020-06-01
Fix build with Qt 5.15

rayslava:
Build with Qt 5.15 causes the following error:
```
  qmapshack-V_1.14.0/src/qmapshack/helpers/CDraw.cpp:210:18: error: aggregate ‘QPainterPath bubblePath’ has incomplete type and cannot be defined
```

### DIFF
--- a/src/qmapshack/helpers/CDraw.cpp
+++ b/src/qmapshack/helpers/CDraw.cpp
@@ -24,6 +24,7 @@
 #include <QImage>
 #include <QPointF>
 #include <QtMath>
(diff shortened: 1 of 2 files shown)


kiozen:
Even if it is a very small PR stick to the rules and use the provided template

https://github.com/Maproom/qmapshack/wiki/DocFaqProjectSite

You can find the content of the template in `.github/PULL_REQUEST_TEMPLATE.md`. Copy it over and fill it.

Issue #262, opened 2020-11-10
Wrong timestamp when loading FIT


norbusan:
**Important Notice:**

**Do not delete any of the sections. Read what is requested.**
**_Replace(!)_ the instructions by your _answer(!)_. Bug**
**reports that are not properly filed might be closed without**
**further notice.**

**Choose a good title**

What is a good title?
The title will be listed in the changelog when the issue is fixed. Do
you think someone else would understand what's fixed or done from your
title? Does it really name the real topic of the issue?

[X ] Did you choose a good title?

**Describe the bug**

The timezone information saved into GPX tracks is inconsistent, and in particular does not contain the correct timezone in the time stamps of the track points.

Looking at the GPX file I see
```
 <metadata>
  <name>2020.11.08 Karikomi-Ike</name>
  <time>2020-11-09T04:50:42.000Z</time>
 </metadata>
```
which is correct (I guess I downloaded it yesterday at that time).

But the track points are incorrect:
```
   <trkpt lon="136.71547961" lat="36.06665066">
    <ele>931</ele>
    <time>2020-11-08T13:31:44.000Z</time>
```
since the time part is the time in JST, but the time zone extensions is `Z` which is `GMT` and not `+0900`.

**What have you done to circle down the problem?**

It is the first time that I realize this.

**To Reproduce**

1. Select a track that is not in GMT timezone
2. Right click and Save as GPX
3. Import the GPX track in other applications, or read the text file manually

**Expected behavior**

Correct timezone information saved.

**Screenshots**

This screenshot shows that the track starts at 13:31 JST (+9)
![ksnip_20201110-120614](https://user-images.githubusercontent.com/1735589/98622586-36932900-234d-11eb-81d4-b754ebe8ed3e.png)

**Attachments**

This is the GPX track I am speaking about.
[2020.11.08 Karikomi-Ike.txt](https://github.com/Maproom/qmapshack/files/5514598/2020.11.08.Karikomi-Ike.txt)

**Tracebacks**

No crashes

**Desktop (please complete the following information):**

 - OS: Linux (Debian)
 - QMapShack Version 1.15.0
 - Window Manager KDE/Plasma

**Additional context**

The track was downloaded from a Garmin Fenix device.

kiozen:
Hard to tell anything as you did not provide the original track file before it was loaded into QMapShack.

norbusan:
Will get it again directly from the watch and attach it.

norbusan:
[...]

PR #263, opened 2020-11-16
[QMS-262] Wrong timestamp when loading FIT

kiozen:
**What is the linked issue for this pull request (start with a `#`):** QMS-#262

The helper function 
```
static QDateTime toDateTime(quint32 timestamp)
```
did not return the UTC timestamp but the local one. Used converting method.

* Load the FIT file from the ticket.
* Copy the track into a gpx project
* Save the gpx project
* Convert the track with gpsbabel
* Compare timestamps in between both GPX
* Load both GPX into QMapShack
* All three tracks (FIT, QMS GPX, gpsbabel GPX) must start with the same timestamp

- [x] yes

- [x] yes


norbusan:
Confirmed that this fixes the issue. Thanks a lot!

Issue #221, opened 2020-07-01
Feature request: Plugin API


gerion0:
*Important Notice:**

**Do not delete any of the sections. Answer them. If you really**
**think they do not apply to you give a short statement. Feature**
**requests that are not properly filed might be closed without**
**further notice.**

**Choose a good title**

What is a good title?
The title will be listed in the changelog when the issue is fixed. Do
you think someone else would understand what's fixed or done from your
title? Does it really name the real topic of the issue?

[x] Did you choose a good title?

**Is your feature request related to a problem? Please describe.**

Some time ago a planned a tour for longer bike trip, and wanted to know all camping sites along the route. This is codable with the Overpass API and/or Python in a straight forward way. However, I was obviously bound to exporting the data from qmapshack as gpx, parse that again and then work with it and reimporting it later on, while I would like to remain in qmapshack.

**Describe the solution you'd like**

I want to achieve the above steps directly within qmapshack. In my opinion the simplest thing for that is a plugin API.

**Describe alternatives you've considered**

The above described way is already an alternative: Write a lot of small scripts in Python etc. and do everything with GPX Im-/Export.

The other way would be to directly edit the source code of qmapshack and maintain a fork for this specific problem or bring it upstream. Nevertheless, writing C++ is hard for such a small thing.

**Give a reason why someone else should spend time on that request**

- Maybe you like the idea.
- A lot of programs get really nice extensions that have a Plugin API (Firefox/Blender/Gajim/...).
- It could lead to more developers that are willing to work within qmapshack.
- It could have some synergy effects, when people would begin to share their code.
- It is a generic solution that enables the solving a lot of very specific problems.

**Additional context**

Typically plugins are in a script language like Lua or Python. Especially for GIS programs also a direct connection to the Overpass API would be nice.

kiozen:
https://github.com/Maproom/qmapshack/wiki/DocFaqProjectSite

gerion0:
> https://github.com/Maproom/qmapshack/wiki/DocFaqProjectSite

Fair point, I have edited my post.

pingurus:
I like the idea, but I think that it needs a good concept. 

The easiest way would probably be to automatically create python wrappers around the non-UI gis classes (i.e. with SWIG). Even though you still would not be able to work all the time inside qmapshack, you could easily access and edit your data without having to import/export it. (At least this is what I think after looking up stuff for 30mins, as I did not create wrappers before)

On a side note: If implemented, this might also cover #206

[...]

PR #286, opened 2020-12-28
WIP:  [QMS-221] Feature request: Plugin API 

pingurus:
_**by your answer**_

**What is the linked issue for this pull request (start with a `#`):** QMS-#221

I added support for python scripting by using [PythonQt](https://github.com/MeVisLab/pythonqt).
For now this PR is only to gather feedback. (Only CGisWorkspace and IGisProject are included in the interpreter for now and no constructors)

1. Setup PythonQt
  I did this using Ubuntu 20.04, thus the steps for your platform may differ. Also, as I am no expert in CMake, I don't know, if the CMakeLists.txt files work on other platforms. My steps were the following:
    ```
    sudo apt install qtbase5-private-dev
    
    git clone git://github.com/commontk/PythonQt.git
    cd PythonQt
    git checkout -b patched-10 origin/patched-10
    cd ..
    mkdir build_PythonQt
    cd PythonQt
    cmake ../PytonQt
    make
    ```

2. Build QMapShack (hopefully) and see console appearing
3. For now an example to see the interpreter working would be:
    ```
    proj = Workspace.selectProject(False)
    proj.getName()
    ```

- [ ] yes

- [ ] yes


pingurus:
@kiozen @gerion0 What do you think of my approach?
I am thinking that it would be useful to build some kind of plugin manager which would be much more user friendly than having to deal with the command line. Do you have any concepts in mind?

kiozen:
To be honest I am not really a fan of this. Adding plugin support to an application is surely a nice challenge. I have done this myself a couple of times. However the benefit is always limited. It adds a lot of cluecode and crude logic to the code base that has to be maintained. If you change the code you have a good chance to break plugins. 

And for what? A handful of users might make some use out of it. The average user has no benefit from it, because they do not know how to code and they are already unable to cope with all the functionality provided by QMS. 

I do not see the point why users that are able to code should not extend QMapShack by the normal way. They have to work their way into the code and the API the one or the other way. If they share their code all users benefit from it. The new function can be discussed and integrated into the GUI nicely. When accepted the function will be documented. 

When done as scripting code or plugin the frustration will be on all sides. The authors of plugin get frustrated when they reach the limits if the plugin API. I get frustrated when I get requests to change the API for the favor of plugins I do not even know. Users get frustrated when using some plugins and a new version is braking them.

Besides that adding Qt/Python as dependency is not fun for the OS X and Windows package maintainers. Now they have to bundle Python and the Qt bindings. Given the fact that non of the maintainers is a real expert for the task and they already struggle with the current library zoo a lot. This will add another heap of trouble for them. 

In other words this is a nice PoC but I am not really happy to get that kind of experimental code with a lot of additional dependencies into the mainstream as it has no benefit for the average users.

gerion0:
> @kiozen @gerion0 What do you think of my approach?
> I am thinking that it would be useful to build some kind of plugin manager which would be much more user friendly than having to deal with the command line. Do you have any concepts in mind?

A command line is nice for development of plugins but what I had in mind are rather some hooks somewhere in the GUI. From a usage perspective that would be:

1. Open some kind of plugin manager.
2. Install a specific plugin (this consists of several Python files).
3. The plugin is usable from the menu bar (Plugins->MyPlugin) can open additional windows and can do actions that are doable as user in qmapshack as well.

On the coding site it could be something like this (for my camping POIs search proposal):
```python
import qmapshack

class CampingPOIs(qmapshack.Plugin):
    def register(self):
        # code that registers the code in the menu (or context menu).

    def execute(self):
        track_name, radius = draw_window_for_user() # open some user window, that request from the user for which track she wants the camping POIs and in what radius they should be
        track_data = qmapshack.get_track(track_name)
        pois = calculate_POIs(track_data, radius)
        for poi in pois:
            qmapshack.load_new_poi(poi)
```

> I do not see the point why users that are able to code should not extend QMapShack by the normal way. They have to work their way into the code and the API the one or the other way. If they share their code all users benefit from it. The new function can be discussed and integrated into the GUI nicely. When accepted the function will be documented.

For me, the main advantages are:
1. Python coding (or another scripting language) is a lot faster than coding in C++. Maybe this is my own bias but I know both languages and also have programmed the same thing in both languages. Python was always much faster in terms of development time (of course not in runtime).
2. For me it was much easier to look into (gpxpy)[https://pypi.org/project/gpxpy/] and quickly hack together what I wanted than even finding out how to compile qmapshack. A plugin infrastructure can bridge here very well. You can simply hack your functionality together and use it in qmapshack (and don't have to think about deploying, coding guidelines, compiling, multi os functionality, ...).
3. Plugins don't need to be good/bugfree/well designed/.... For programs that have plugins (Firefox/Blender/Kodi...), it is normal for me that some plugins simply don't work, crash sometimes, are not compatible with the current version anymore and so on. But that is not a problem. Nevertheless, they help for sharing the code. If someone has an idea, she can write it down (if capable of Python) and gets familiar with qmapshack that way and plublish it afterwards.
4. Yout as qmapshack maintainer don't need to decide, whether a plugin is useful or need. It can simply exist (some kind of playground). It it prooves useful, then it can be integrated into the core.
5. A plugin has a very narrow scope. The plugin maintainer/developer only has to maintain this specific plugin.


> When done as scripting code or plugin the frustration will be on all sides. The authors of plugin get frustrated when they reach the limits if the plugin API. I get frustrated when I get requests to change the API for the favor of plugins I do not even know. Users get frustrated when using some plugins and a new version is braking them.

Maybe you also get a lot of new functionality that is currently completely out of scope.

My specific use case was a possibility to calculate camping sites along a track. Would you be willing to integrate this function? It is fairly specific, most users don't need it, it leads to additional maintainance costs for little usage.

Of course I am not the one who decides this and also not the one who has to maintain it afterwards. I have seen a lot of other software projects that has plugins and saw it always as a benefit. I filed the issue because I encountered the situation that I needed a function that was not part of qmapshack. I tried to look at the source code to extend qmapshack (e.g. the gpx parsing code) but gave up because I didn't want to spend that much time for it. Then I hacked it together in Python (~1 hour, never used gpxpy before), which fulfils my needs but was not well written etc.. A plugin would have connected both sides.

kiozen:
Somehow your reasoning convince me even more that plugins are a hacky solution that require quite some infrastructure on the project without a real benefit for the project or the community. Just benefits for some single individuals.  Considering the small group of QMapShack users this will be just a hand full. 

"My specific use case was a possibility to calculate camping sites along a track. Would you be willing to integrate this function? "

I think many users would be interested in finding POIs along a track/route. That would be a feature. There is something similar already "Search Web for position" (right click on the map). That was requested by a user with an incredible well defined feature request. The user did a quite intensive search for services. He figured out all the urls and if it is feasible to use them in a common way. Together we developed a nice GUI and implemented it. And now it's of everyone's benefit. 


On the technical side: 
Adding stuff to the menu: Usually everything that is in a menu is available as a tool button, too. The tool button bar is configurable. Adding stuff dynamically will be tricky.

Adding item related functions: The problem of GIS software is that there a many items on a map. Each item can be considered as individual document and on the first sight completely unrelated to other items. Therefore the GUI has to solve the focus-follows-mind problem when it comes to applying functions to items. QMapShack solves this by forcing the user to either pick an item from the workspace tree and apply the function via context menu. Or by selecting the item on the map and selecting the function from the bubble. Therefore adding something to the menu bar will not be the right way if it comes to functions operating on items. It has to be added to the context menu and the tool button bar of the bubble.

Exposing the data model: Especially tracks are very complex. There is a lot of information attached to the track points, we have segmentation etc. If you expose this as a read only simplified track it's ok. If you expose this as read/writable object for all kinds of operations things get really complicated. As the GPX format is not capable holding all data that is available in the internal data model a read/writable exchange is tricky and limited.

It might be simply to hack something in a Python script. In fact I do this quite often. However it's much more of a challenge to get something into a full featured GUI application with a complexity of QMapShack that is targeting average users. And according to the Web QMapShack is already way to complex for most of them.

pingurus:
I think both of your stances are perfectly valid, so I'll just drop in what came to my mind while reading your comments:

@kiozen:

> I do not see the point why users that are able to code should not extend QMapShack by the normal way. They have to work their way into the code and the API the one or the other way. If they share their code all users benefit from it. The new function can be discussed and integrated into the GUI nicely. When accepted the function will be documented.

I do think that there are use cases in which you hack something together for your specific use case where you don't want to make the effort to formalise it into something that can be integrated into QMS. As a plugin it still can be shared for inspiration. 

> When done as scripting code or plugin the frustration will be on all sides. The authors of plugin get frustrated when they reach the limits if the plugin API. I get frustrated when I get requests to change the API for the favor of plugins I do not even know. Users get frustrated when using some plugins and a new version is braking them.

The point when the author reaches the limit of the API is where they should transfer the code to C++, as the Python API will be nothing more than a wrapper (at least if we take what I started now). If it is clearly documented somewhere that the API is and will be only an auto-generated wrapper there is no reason for discussion. (Yes, I know that it is easy to say for me since you have to bother with it, but you enforced the issue templates in a similar way)

> Besides that adding Qt/Python as dependency is not fun for the OS X and Windows package maintainers. Now they have to bundle Python and the Qt bindings. Given the fact that non of the maintainers is a real expert for the task and they already struggle with the current library zoo a lot. This will add another heap of trouble for them.

That is a good point that I did not consider until now.

@gerion0:

 > I tried to look at the source code to extend qmapshack (e.g. the gpx parsing code) but gave up because I didn't want to spend that much time for it. Then I hacked it together in Python (~1 hour, never used gpxpy before), which fulfils my needs but was not well written etc.. A plugin would have connected both sides.

Don't expect a well documented API for the plugins, my code really just exposes the public member functions of all the QMS classes to python. I think that otherwise it would be too much effort to maintain. On the other hand that makes it even easier to afterwards implement that piece of code in C++. 

@kiozen:

>  And now it's of everyone's benefit.

Plugins can be to everyone's benefit aswell, especially if we provide a central repository where we pool working plugins. I understand if you don't want to review even more code, but it'd be fine for me to help you out with looking over those plugins.

> Adding stuff to the menu: Usually everything that is in a menu is available as a tool button, too. The tool button bar is configurable. Adding stuff dynamically will be tricky.

The (bubble?) button and (context?) menu entry could simply be "send to plugin" where a dropdown or similar pops up where you can select the plugin that you want to send this item to.

> Exposing the data model: Especially tracks are very complex. There is a lot of information attached to the track points, we have segmentation etc. If you expose this as a read only simplified track it's ok. If you expose this as read/writable object for all kinds of operations things get really complicated. As the GPX format is not capable holding all data that is available in the internal data model a read/writable exchange is tricky and limited.

But that's the case independent of wether you code a plugin or c++ code, isn't it?

> And according to the Web QMapShack is already way to complex for most of them.

Wouldn't plugins (be it that they are seperate as python scripts or already compiled and included as c++ code) be a good possibility to reduce the complexity of QMS? There are some features that are perfectly integrated (and by default visible) that I have never (or very rarely) used. Those features (and others that I may use by chance) could be invisible by default, so you have to install/activate them via some plugin menu and newbie users aren't confused by the mass of options.

A benefit of plugins that wasn't mentioned before, is that the feature is available immediatly, opposed to having to wait for the next release. So plugins would make features available earlier without putting extra work to the maintainers besides the initial (?) effort to bundle Python,Qt and QMS. But I can't really estimate the effort, as I failed at my try compiling QMS for Windows 😆. (At least when I still used QMS on Windows but coded it in an Ubuntu VM it was always a bit frustrating having to wait for the next release to use 'my' new feature. I guess it is similar if someone else codes something for you when you are not able to code and compile yourself.)

Just as @gerion0 said, you are the one that does most of the maintenance work here, so it is your decision wether there will be plugins or not. In any case I think I will do some work on my branch and keep it mergable, so I can use at least the CLI to be able to hack around before having to cast stuff into C++ code.

kiozen:
I know the benefits of plugins you list, however in my real life experience it never paid out well. Or let's say it rarely paid out. For instance  I hardly use any add-on or plugin in my browsers because I am fed up with the troubles on every update. I am fed up with trying out stuff of questionably quality. The idea is appealing - the execution seldom of a benefit.

I also don't think that just dumping the API as a Python wrapper is a good idea. You have done enough work on the code to know that it's quite tricky to change stuff, to update the complete GUI with it's multiple views and dialogs and to keep the history of changes.  You do not want to place that burden into a plugin. 

Imho the only reliably way is to pass something like a GPX to the plugin. The plugin can use/alter the data and pass back the changed GPX if necessary. Now the code can copy over the altered data into the original data, update the GUI and add an entry into the history. Sounds simple, but might be tricky if the plugin altered/extended the extensions. 

Another topic to solve is to define a mechanism to exchange capabilities with the plugin. On what items does the plugin apply? Only tracks, only waypoints etc... Or does it need a project? Is it fit for bulk selections? Does it work on mixed selections? To which project new content belongs? Especially if there are multiple items from multiple projects. Just to name a few....

How do we make sure a plugin provides all information? Strings for the menus, for the change history? Should there be a way to provide icons? etc..

The challenge of a  good plugin API is to define an interface that allows creative use without opening up vulnerabilities or making stuff too complex. You do not want to pull the complexity of the app  into the plugin. And you do not want the plugin to add complexity to the app. I lack a good solution for that problem. So far it's just a request that's not well defined at all. 

If we do not find a really clever solution for that we just end up in implementing POIs along track the worst and most complicated way ever as it won't be used by anyone else. And probably we already spent more time on this than ever will be spend on writing plugins ;)

kiozen:
Btw another of these user extendable features is writing templates to document stuff. It's well documented but no one is using it.

https://github.com/Maproom/qmapshack/wiki/DocGisTemplates

wthaem:
Sorry, I do! And it helps me a lot to collect quickly additional information in several use cases. Nothing for the public - nevertheless: I enjoy it.

kiozen:
How to proceed? Close it? Keep it?

pingurus:
As I'll be occupied with the other feature request, I'll close it.

Issue #289, opened 2021-01-03
Add support for HiDPI screens


cryptomilk:
**Important Notice:**

**Do not delete any of the sections. Read what is requested.**
**_Replace(!)_ the instructions by your _answer(!)_. Bug**
**reports that are not properly filed might be closed without**
**further notice.**

**Choose a good title**

What is a good title?
The title will be listed in the changelog when the issue is fixed. Do
you think someone else would understand what's fixed or done from your
title? Does it really name the real topic of the issue?

[x] Did you choose a good title?

**Describe the bug**

The icons are not rendered correctly on HiDPI screen and look pixelated.

**What have you done to circle down the problem?**

Did you compare to previous versions? What versions? Did you check
if others have the same problem?

**To Reproduce**

1. Open QMapShack on a HiDPI screen

**Expected behavior**

Icons shouldn't look horrible on HiPDI screens.

**Screenshots**

**Attachments**

Can you provide an example file to trigger the problem?

**Tracebacks**

If QMapshack crashes the fastest way to get help is a backtrace. We are
aware that it's not that easy to record one. However if you provide one
chances are good to have it solved very fast. If the problem can't be
reproduced on the developer's environment we will sooner or later request
a traceback.

For Linux see:
https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux

For Windows it would need a debug build and running QMapshack in Visual Studio.

**Desktop (please complete the following information):**

- Latest version

**Additional context**

PR #290, opened 2021-01-03
Add support for HiDPI screens

cryptomilk:
Signed-off-by: Andreas Schneider <asn@cryptomilk.org>

_**by your answer**_

**What is the linked issue for this pull request (start with a `#`):** QMS-#289

I enabled support for HiDPI screens so that icons are rendered in the correct resolution.

1. Open qmapshack
2. Check that icons are rendered nicely and are not looking pixelated.

- [x] yes

- [ ] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@ V1.XX.X
 [QMS-275] Routino: Add Spanish and Czech as selectable languages for turn instructions
 [QMS-279] Track metrics not updated when using UNDO / REDO in Edit mode
 [QMS-282] Tags icons/rating disappear from workspace after saving and closing a project
+[QMS-289] Added suport for HiDPI screens
 
 V1.15.2
 [QMS-264] Windows: adapt build scripts for 1.15.1 release
--- a/src/qmapshack/main.cpp
+++ b/src/qmapshack/main.cpp
@@ -30,6 +30,9 @@ int main(int argc, char ** argv)
 {
     QApplication app(argc, argv);
 
+    QApplication::setAttribute(Qt::AA_UseHighDpiPixmaps);
+    QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+
     QCoreApplication::setApplicationName("QMapShack");
     QCoreApplication::setOrganizationName("QLandkarte");
     QCoreApplication::setOrganizationDomain("qlandkarte.org");
--- a/src/qmaptool/main.cpp
+++ b/src/qmaptool/main.cpp
@@ -28,6 +28,9 @@ int main(int argc, char ** argv)
 {
     QApplication app(argc, argv);
 
+    QApplication::setAttribute(Qt::AA_UseHighDpiPixmaps);
+    QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+
     QCoreApplication::setApplicationName("QMapTool");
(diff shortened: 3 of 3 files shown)


changelog.txt
```diff
@@ -1,4 +1,5 @@
 V1.XX.X
+[QMS-289] Added suport for HiDPI screens
```
kiozen, line 2: please keep them in order.

src/qmapshack/main.cpp
```diff
@@ -30,6 +30,9 @@ int main(int argc, char ** argv)
 {
     QApplication app(argc, argv);
 
+    QApplication::setAttribute(Qt::AA_UseHighDpiPixmaps);
+    QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+
```
kiozen, line 35: Those two lines have to go to the qmaptools binary, too.

kiozen (changes requested):
I can  not test the PR therefore I just have to ask:

* How does that patch affect icons on the map? If they are scaled, too, are they placed correctly? 
* How does the patch affect icons on a track profile graph (place waypoints close to the track)?
* Did you check the dialogs for correct layout?

Maybe someone else with a HiDPI screen can test, too.

cryptomilk:
Current qmapshack
qmapshack_dev

With HiDPI support enabled
qmapshack_hidpi_enabled

changelog.txt
cryptomilk: I've moved it to the end.

src/qmapshack/main.cpp
cryptomilk: Done.

cryptomilk:
This is the first step to show icons correctly. For all other stuff you need to use pt or em values instead of px, in case you use qml. For CXX code you need to calculate it correctly.

The splash screen is for example still rendered incorrectly.

Example: https://github.com/KDE/gwenview/commit/97e7de52276b46bc99b64117abfee0965e6f58e6

cryptomilk:
What the patch does is e.g. if you have defined to use 32x32 px icons for the toolbar, it will load 64x64 px ones on hidpi screens. For everything else you which has been done in CXX code you have to keep the display ratio in mind and calculate the stuff correctly.

For qml use pointsize instead of pixel: https://github.com/KDE/elisa/commit/6067340472468d5ce37c7674ddbc44d59846c7fd

kiozen:
Ok, I just wanted to make sure it doesn't make the situation worse. Yes I am aware that the screen resolution should have been taken into account.  Back in these day HiDPI screens where no real thing :smiley: 

And without one I will not be able to fix and test all occurrences. And there are plenty. So far HiDPI users help themselves by setting the environment variable:

```
export QT_SCALE_FACTOR=2
```

I guess the patch is breaking that a bit as the icons are rendered to big in that case.

kiozen:
How to proceed here? My current status is:

* The patch increases nothing else as the icon sizes of buttons and menus
* If used in combination with `QT_SCALE_FACTOR` this might lead to unintended large icons.

Therefore I am a bit reluctant to merge it.

kiozen:
I close this one as it seems to be stuck.

cryptomilk:
I didn't have time to look into. But feel free to correctly support HiDPI screens. They will be more and more in future.

Issue #294, opened 2021-01-05
Read Mapsforge POI files


pingurus:
**Important Notice:**

**Do not delete any of the sections. Read what is requested.**
**_Replace(!)_ the instructions by your _answer(!)_. Feature requests**
**reports that are not properly filed might be closed without**
**further notice.**

**Choose a good title**

What is a good title?
The title will be listed in the changelog when the issue is fixed. Do
you think someone else would understand what's fixed or done from your
title? Does it really name the real topic of the issue?

- [x] Did you choose a good title?

**Is your feature request related to a problem? Please describe.**

When planning a trip it is convenient to be able to see POIs that are relevant to you on the map. Depending on the zoom level and style of the background map this is not always possible.

**Describe the solution you'd like**

Loading the poi files provided with the mapsforge maps as a project similar to the geosearch would enable users to display the pois that are relevant to them.

**Describe alternatives you've considered**

As an alternative to the .poi file one could load the .db file generated for locus maps. The .poi file seems easier to handle to me (both are SQLite files).

**Give a reason why someone else should spend time on that request**

I think POIs are relevant to most of the users and currently there is no fully integrated way to view POIs directly on the map.
Having POIs loaded as a project would enable a user to directly copy those that are relevant to them into their respective project without having to manually fill in everything.

**Additional context**

kiozen:
"Loading the poi files provided with the mapsforge maps as a project similar to the geosearch would enable users to display the pois that are relevant to them."

Can you provide more details on that? Maybe sketch some example dialogs? I would expect that the files could be loaded just like any other project. Therefore it would be an extension of the supported GIS data formats. But this sentence reads like you want to do something else.

pingurus:
Yes, I want that the files are loaded just like any other project. However, the poi file contains nearly 1 million pois only for Baden-Württemberg. Loading all of those would a) clutter the map b) hang QMS.

My solution idea was that there is a settings dialog for the project (just like there is one for the geosearch), where you can select the poi types that you want to view with tickmarks in a treeview. QMS then only reads pois of the selected types from the database and only creates gisItems for those. 

As that still may be too much to handle, I am thinking whether it is possible to only load data that is spatially relevant to the user. An idea would be that only those are loaded, that are in the current view (which would have to be updated on every map move) and those that are in the bounding box of the other project. I am not sure though if that is feasible and a good idea.

kiozen:
[...]

PR #296, opened 2021-01-09
Read Mapsforge POI files

pingurus:
_**by your answer**_

**What is the linked issue for this pull request (start with a `#`):** QMS-#294

I added a class derived from IGisProject to load Mapsforge *.poi files

0. Download poi file from i.e. openandromaps.org (is only available together with map)
1. Open poi file as you would open any project
2. Click on the settings button on the project to select what you want to see
3. Select a category. Be careful what you wish for, as there are no limits yet on how much is loaded.
4. See the POIs you selected

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/CMakeLists.txt
+++ b/src/qmapshack/CMakeLists.txt
@@ -106,6 +106,8 @@ set( SRCS
     gis/ovl/CDetailsOvlArea.cpp
     gis/ovl/CGisItemOvlArea.cpp
     gis/ovl/CScrOptOvlArea.cpp
+    gis/poi/CPoiProject.cpp
+    gis/poi/CPoiProjectSettingsDialog.cpp
     gis/prj/CDetailsPrj.cpp
     gis/prj/IGisProject.cpp
     gis/qlb/CQlbProject.cpp
@@ -458,6 +460,8 @@ set( HDRS
     gis/ovl/CDetailsOvlArea.h
     gis/ovl/CGisItemOvlArea.h
     gis/ovl/CScrOptOvlArea.h
+    gis/poi/CPoiProject.h
+    gis/poi/CPoiProjectSettingsDialog.h
     gis/prj/CDetailsPrj.h
     gis/prj/IGisProject.h
     gis/qlb/CQlbProject.h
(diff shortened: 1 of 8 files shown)


pingurus:
Any feedback is appreciated.

kiozen:
Ok, here is my first impression:

1. It takes ages to load the POIs. I started to load 4000 POIs and QMS is loading now for several minutes.
2. While loading the cursor blinks between wait and arrow.  It should stay as wait cursor. And a progress dialog must be shown with the chance to cancel
3. The default icon is ugly as hell for a reason: To make everyone to pick an appropriate icon asap. As there is no way to save a change and QMS is quite inefficient changing icons for several thousand points appropriate icons  should be provided.
4. The POI format is missing in the list of supported formats (file dialog when opening a POI file)
5. The project is not restored in the workspace.
6. The number of POIs available only show up when expanding the items (users will complain)

I have a bit of a problem with such kind of "he-who-dies-with-the-most-wins" collections (aka Archies camp site list, aka Tonies Hüttenliste). They do exist, there are users that insist on using them, they are a huge pain in the butt because they stress the system. And it will be those users using these collections complaining that QMS is crappy as it does not perform very well. The experience from above simply confirms that experience from the past another time.

As already mentioned in the ticket it needs a bit more intelligence for such kind of databases. They can't be handled like any other GIS project file. Imho they are more like another feature next to maps, dem, routing etc. There need to be some algorithm to load and show only those POIs that fit into the viewport. And for outer zoom levels none or only selected POIs are shown. Additionally users want to search these databases and they want to spice their tracks with POIs along tracks. These are all requirements that are not really covered by the workspace. And for most of them the workspace as a place for user data to work with is not the right scope for the task. 

I think working with POI collections is an interesting topic and is useful while planning. The more I think about it, it's an entirely new feature and should go into it's own docker widget. Just like maps or DEM data. POI collections can be activated or deactivated per view. Within that docker we are much more flexible to do stuff the right way as we do not have to stick to the rules of the workspace. 

This will be quite a large change and I do not want to review it as one complete blob. This will need a feature branch next to dev. This allows us to create tickets that can be processed as a PR without providing a complete feature ready for the users. I have created such a branch in the main repo. It's name is `feature/pois`. Please use it instead of dev. I will rebase it after each PR on dev. Your PRs will target `feature/pois` in the main repo. You can create the tickets step by step in the normal tracker. Afaik there is no way to define tickets as epics with sub-tasks. Maybe it's good practice to reference the initial ticket with each new ticket.

Anyway the idea is to plan this step by step and to finish one step before you proceed to the next one. By that we can test and verify the implementation in smaller atomic steps. You do not have to plan all steps in advance. Stay agile and define ticket(s) that can be done in a reasonable time frame. They do not have to result into something that is ready for the user. By that it should be quite easy to tackle the beast.

Issue #301, opened 2021-01-10
Load POI file


pingurus:
**Important Notice:**

This is for developers only. As a user use the `bug report` or `Everything that's not a bug` template.

**The parent ticket**

_add ticketnumber behind the #_

QMS-#294

**What do you plan to do**

Load the POI file "blindly", just as it was already the case in #296.

**The result will be**
- [x] QMS is able to load a large number of POIs without too much lag
- [x] It is selectable which types of POIs to load
- [x] The POIs have the icons that correspond to their type

kiozen:
Also consider not to load all pois and to iterate over them but to make an intelligent query that sets limits to the coordinates and other filtering. For the first try limiting the coordinate to the viewport would be fine.

PR #302, opened 2021-01-11
[QMS-301] Load POI file 

pingurus:
_**by your answer**_

**What is the linked issue for this pull request (start with a `#`):** QMS-#301

I added code to display POIs. The logic when to do which database requests might not be perfect yet. (But it can handle all hotels in the baden-wurttemberg openandromaps poi file at once)

Also, only two categories are implemented for now and icons do not work.

1. Add folder with poi files
2. Select a category
3. See Icons appear

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/poi/CPoiPOI.cpp
+++ b/src/qmapshack/poi/CPoiPOI.cpp
@@ -1,5 +1,6 @@
 /**********************************************************************************************
     Copyright (C) 2020 Oliver Eichler <oliver.eichler@gmx.de>
+    Copyright (C) 2021 Henri Hornburg <pingurus@t-online.de>
 
     This program is free software: you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
@@ -17,18 +18,28 @@
 **********************************************************************************************/
 
 #include "helpers/CDraw.h"
+#include "poi/CPoiCategory.h"
 #include "poi/CPoiDraw.h"
 #include "poi/CPoiPOI.h"
 
+#include <QSqlDatabase>
+#include <QSqlError>
+#include <QSqlQuery>
 #include <QtWidgets>
 
 CPoiPOI::CPoiPOI(const QString &filename, CPoiDraw *parent)
     : IPoi(parent)
     , filename(filename)
+    , loadTimer(new QTimer(this))
 {
     // Set true if the file could be open and loaded successfully
     // If not set true the system will take care to destroy this object
     isActivated = true;
+
+    //Set up a timer to only start reloading and redrawing the map when the user has made his selections
+    loadTimer->setSingleShot(true);
+    loadTimer->setInterval(500);
+    connect(loadTimer, &QTimer::timeout, poi, &CPoiDraw::emitSigCanvasUpdate);
 }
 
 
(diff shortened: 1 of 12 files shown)


pingurus:
@kiozen How can I add the icons to the resources.qrc file? Qt won't let me do it:
Screenshot from 2021-01-11 22-45-21

kiozen:
I always edit the resource file with a normal text editor. It's more or less ordered and I would like to keep it like that.

kiozen:
Can we do the icons as separate ticket and PR, please.

kiozen:
much better, thx!

pingurus:
Two things remain to do from my side for this ticket:

- [ ] fill tagMap
- [ ] find cause why checking the parent category doesn't check the children. (This happened automatically in the last PR)

So from my point of view you could start reviewing the logic of loading the pois. I am not sure if the mutex lock couldn't be narrowed down a bit, but I think not.

kiozen:
[ ] smaller icons :wink:

src/qmapshack/poi/CPoiPOI.cpp
```diff
@@ -57,9 +66,128 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
...
+    const QString queryString = "SELECT main.poi_index.maxLat, main.poi_index.maxLon, main.poi_index.minLat, main.poi_index.minLon, main.poi_data.data"
+                                " FROM main.poi_index"
+                                " INNER JOIN main.poi_data ON main.poi_index.id=main.poi_data.id"
+                                " WHERE main.poi_index.maxLat<" + QString::number((minLatM10 + 1) / 10., 'f') +
+                                " AND main.poi_index.minLat>" + QString::number(minLatM10 / 10., 'f') +
+                                " AND main.poi_index.maxLon<" + QString::number((minLonM10 + 1) / 10., 'f') +
+                                " AND main.poi_index.minLon>" + QString::number(minLonM10 / 10., 'f') +
+                                " AND main.poi_data.data LIKE '%" + tagMap[category] + "%'";
+
+    QSqlQuery query(queryString, QSqlDatabase::database(filename));
```
kiozen, line 182: Have a look into `CDBItem.cpp`. Using `bind()` is much safer to get a well formatted SQL query.

```
    QSqlQuery query(db);
    query.prepare("SELECT trash FROM items WHERE id=:id");
    query.bindValue(":id", id);
    QUERY_EXEC(return );
```

src/qmapshack/poi/CPoiPOI_TagMap.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "CPoiPOI.h"
+
+QMap<QString, QString> CPoiPOI::tagMap;
+QMap<QString, QString> CPoiPOI::initTagMap()
+{
+    QMap<QString, QString> map;
+    map["accommodation_hotel"] = "tourism=hotel";
+    map["water_weir"] = "waterway=weir";
+    return map;
+}
```
kiozen, line 28: What is the reason to place this into it's own file?

src/qmapshack/poi/CPoiPOI.cpp
```diff
@@ -57,9 +66,128 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
...
+                if(!loadedPOIs.contains(category) ||
+                   !loadedPOIs[category].contains(minLonM10) ||
+                   !loadedPOIs[category][minLonM10].contains(minLatM10))
+                {
+                    mutex.unlock();
+                    loadPOIsFromFile(category, minLonM10, minLatM10);
+                    mutex.lock();
+                }
+
+                for(poi_t poiToDraw : loadedPOIs[category][minLonM10][minLatM10])
```
kiozen, line 102: `const poi_t& poiToDraw`

src/qmapshack/poi/CPoiPOI.cpp
```diff
@@ -57,9 +66,128 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
...
+        x2 =  180 * DEG_TO_RAD;
+    }
+
     // draw POI
-    QPointF myPoi(12.329699 * DEG_TO_RAD, 45.437951 * DEG_TO_RAD);
-    poi->convertRad2Px(myPoi);
-    QPixmap icon("://icons/waypoints/32x32/FlagBlue.png");
-    p.drawPixmap(myPoi, icon);
+    mutex.lock(); //Does this have to be so broad?
+    for(QString category : categoryActivated.keys(true))
```
kiozen, line 87: `const QString& category`

src/qmapshack/poi/CPoiPropSetup.cpp
```diff
@@ -29,6 +30,19 @@ CPoiPropSetup::CPoiPropSetup(IPoi *poifile, CPoiDraw *poi)
...
+    treeWidgetCategories->setColumnCount(3);
+    treeWidgetCategories->setHeaderLabels(QStringList()
+                                          << tr("Selected")
+                                          << tr("Parent")
+                                          << tr("Name"));
+    QList<QTreeWidgetItem *> topLevelItems;
+    poifile->getCategoryTreeWidgetItems(topLevelItems);
+    treeWidgetCategories->insertTopLevelItems(0, topLevelItems);
+    treeWidgetCategories->sortItems(eTreeColumnCategory, Qt::SortOrder::AscendingOrder);
+    treeWidgetCategories->hideColumn(eTreeColumnParentCategory);
```
kiozen, line 43: Some of these settings can be done by Designer. This is the preferred way.

kiozen (changes requested):

src/qmapshack/poi/CPoiPOI_TagMap.cpp
pingurus: There will be ~250 entries in that map. I put it in a separate file so the other one is not cluttered.

src/qmapshack/poi/CPoiPOI.cpp
pingurus: That's not possible since below there is a call to convertRad2Px which changes the values in-place.

pingurus:
>     * [ ]   smaller icons wink

Should I use the same issue number as for adding them in the first place?

src/qmapshack/poi/CPoiPOI.cpp
pingurus: Do you have an idea on how to reformat that also for the string that comes from tagMap? The problem is, that the query has a different format depending on the category

kiozen:
> > ```
> > * [ ]   smaller icons wink
> > ```
> 
> Should I use the same issue number as for adding them in the first place?

Leave the files as they are. Simply scale the icons while you draw them. Ideally with respect of the screen resolution (use pt instead of px). Btw a pixmap's top left corner is placed at the point given when drawn. Therefore you should offset the point by half the icon size to place the icon's center at the point.

src/qmapshack/poi/CPoiPOI.cpp
kiozen: ok in that case `poi_t& poiToDraw`. The important part is the reference.

src/qmapshack/poi/CPoiPOI.cpp
kiozen: hm, ok I see your point.  Don't have a good solution for that right at hand.

src/qmapshack/poi/CPoiPOI.cpp
pingurus: But if I do the reference it will be changed in the QMap. And if it is drawn the next time, no longer the rad value will be converted to pixel, but the old pixel value will be converted to something else. Or do I get something wrong?

kiozen: Ok, that slipped my attention, loading POIs is accumulative. I wonder if that is a good idea as it will consume more and more memory. On the other hand side it shouldn't be that much.

I would prefer some on-the-fly query however the whole process seems to take quite some time. Haven't placed timers yet to find out what part consumes the most. But we can do the optimization in another ticket.

kiozen:
```
    p.translate(-pp);
    p.setBrush(Qt::white);
    p.setPen(Qt::NoPen);
...
...
This is my slightly enhanced version:
...
                if(poi->needsRedraw())
                {
                    mutex.unlock();
                    return;
                }

                for(const poi_t& poiToDraw : loadedPOIs[category][minLonM10][minLatM10])
                {
                    QPointF pt = poiToDraw.coordinates;
                    poi->convertRad2Px(pt);

                    QRectF r1 = icon.rect();
                    r1.moveCenter(pt);

                    p.drawRect(r1.adjusted(-1, -1, 1, 1));
                    p.drawPixmap(r1.topLeft(), icon);
                }

```

src/qmapshack/poi/CPoiPOI.h
```diff
@@ -22,18 +22,49 @@
...
+        eSqlColumnMaxLat,
+        eSqlColumnMaxLon,
+        eSqlColumnMinLat,
+        eSqlColumnMinLon,
+        eSqlColumnData
+    };
+
     QMutex mutex;
     QString filename;
+    QTimer loadTimer;
```
kiozen, line 61: Qt style is to use a pointer and to construct the instance in the parent's ctor. I would like to keep it uniform.

src/qmapshack/poi/CPoiPOI.h
```diff
@@ -22,18 +22,49 @@
...
+    QMap<QString, bool> categoryActivated;
```
kiozen, line 62: You use this map to store the check state converting the check state into a bool and vice versa. Wouldn't it be much more straight forward to use the check state?

src/qmapshack/poi/CPoiPOI.cpp
```diff
@@ -56,10 +65,139 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
...
+            {
+                if(!loadedPOIs.contains(category) ||
+                   !loadedPOIs[category].contains(minLonM10) ||
+                   !loadedPOIs[category][minLonM10].contains(minLatM10))
+                {
+                    mutex.unlock();
+                    loadPOIsFromFile(category, minLonM10, minLatM10);
+                    mutex.lock();
+                }
+
```
kiozen, line 105:
```
                if(poi->needsRedraw())
                {
                    mutex.unlock();
                    return;
                }
```

src/qmapshack/poi/CPoiPropSetup.cpp
```diff
@@ -29,6 +30,12 @@ CPoiPropSetup::CPoiPropSetup(IPoi *poifile, CPoiDraw *poi)
 
     connect(sliderOpacity,         &QSlider::valueChanged,      poifile, &IPoi::slotSetOpacity);
     connect(sliderOpacity,         &QSlider::valueChanged,      poi,     &CPoiDraw::emitSigCanvasUpdate);
+
+    poifile->addTreeWidgetItems(treeWidgetCategories);
+    treeWidgetCategories->sortItems(eTreeColumnCategory, Qt::SortOrder::AscendingOrder);
+    treeWidgetCategories->hideColumn(eTreeColumnParentCategory);
```
kiozen, line 36: Why do we have this column if we hide it all the time?

src/qmapshack/poi/CPoiPOI.cpp
```diff
@@ -56,10 +65,139 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
...
+        item->setData(CPoiPropSetup::eTreeColumnCategory, Qt::DisplayRole, child);
+        item->setCheckState(CPoiPropSetup::eTreeColumnCheckbox, categoryActivated[category] ? Qt::Checked : Qt::Unchecked);
+    }
+}
+
+void CPoiPOI::checkedStateChanged(QTreeWidgetItem * item, int columnChanged)
+{
+    if(columnChanged == CPoiPropSetup::eTreeColumnCheckbox)  //shouldn't happen, but still
+    {
+        QString parent = item->data(CPoiPropSetup::eTreeColumnParentCategory, Qt::DisplayRole).toString();
```
kiozen, line 151: QTreeWidgetItem::data() / setData() should be avoided as it ends up quickly in cryptic code. Defining your own item class that inherits QTreeWidgetItem is a much cleaner approach

kiozen (changes requested):

src/qmapshack/poi/IPoi.h
```diff
@@ -48,6 +48,11 @@ class IPoi : public IDrawObject
 
     bool activated() const {return isActivated;}
 
+    virtual void addTreeWidgetItems(QTreeWidget* widget) = 0;
+
+public slots:
+    virtual void checkedStateChanged(QTreeWidgetItem*item, int columnChanged) = 0;
```
kiozen, line 54: Stick to the name pattern: slots names start with `slot...` signals with `sig...` It really helps.

pingurus: Yep, forgot about that :D

kiozen (approved):

kiozen:
Ok, well done. I will merge it.

Some points you might consider and create new tickets:

* Instead of having this treewidget in the properties you might add the items directly to the CPoiPOI item. And you can iterate over the child items and add results to the items. In that case the loadedPOIs map would be distributed over these item. 

* Each CPoiCategory item could display it's icon next to the checkbox

* The search has to be optimized. There should be a sweet spot for the area covered by each query. If it is too small you ahve too many queries. If it is too large queries take too long. There might be also some potential in optimizing the SQL query. Or by creating additional temp tables at startup. I am no database specialist but I think there is a lot of potential in how to query a database.

* There has to be some limitation for the outer zoom levels else the POIs spam the map. Maybe we can use the same slider as for the maps to define the zoom range POIs are shown.

* Add some information about the POI as text (e.g. name) or show information when having the mouse close by.

These are just my thoughts while reviewing.

Issue #220, opened 2020-06-26
"Select items from map" misses items


pingurus:
**Important Notice:**

**Do not delete any of the sections. Answer them. If you really**
**think they do not apply to you give a short statement. Bug**
**reports that are not properly filed might be closed without**
**further notice.**

**Choose a good title**

What is a good title?
The title will be listed in the changelog when the issue is fixed. Do
you think someone else would understand what's fixed or done from your
title? Does it really name the real topic of the issue?

- [x] Did you choose a good title?

**Describe the bug**

When I select items from the map to copy them and drag the map in a way, that some items temporarily go off screen, these items are not copied.

**What have you done to circle down the problem?**

The items that are copied are not those, that are currently on screen, but those, that were on screen, when the area was last resized (marked in red). The number indicated in the info box is correct though.

**To Reproduce**

1. Open some items
2. Select "select items from map"
3. Choose an area, drag map so a part of the map moves out of the screen (with items in it) and then resize it
4. See that not all items are marked red anymore. Those that are not marked red, aren't copied.

**Expected behavior**

I expect all items to be copied that are in the selected area, not only those that were on screen during the last resize.

**Screenshots**

![grafik](https://user-images.githubusercontent.com/36167579/85855776-17896580-b7b7-11ea-9540-b0616abb7f21.png)

**Attachments**

Any items will do

**Tracebacks**

If QMapshack crashes the fastest way to get help is a backtrace. We are
aware that it's not that easy to record one. However if you provide one
chances are good to have it solved very fast. If the problem can't be
reproduced on the developer's environment we will sooner or later request
a traceback.

For Linux see:
https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux

For Windows it would need a debug build and running QMapshack in Visual Studio.

**Desktop (please complete the following information):**

 - OS: [e.g. OS X, Debian, Suse, Windows]
 - QMapShack Version [e.g. 1.13.1]
 - Window Manager [e.g. KDE, GNOME]

**Additional context**

PR #305, opened 2021-01-16
[QMS-220]  "Select items from map" misses items

kiozen:
**What is the linked issue for this pull request (start with a `#`):** QMS-#220

Root cause was using the pixel coordinates of the items. This is
quite fast as long as the item is in the visible map. However once
it is outside the visible part these coordinates are invalidated.
Therefore the compare will fail.

Using the real coordinates is a bit more effort especially for
tracks, routes and areas. As the coordinates of these items can be
changed (or points get hidden) at any time the resulting polylline
should be derived on-the-fly. To store a copy would need some
failsafe mechanism to update when needed.

This is a bit slower than before.  Tested with 393 tracks and 2624
waypoints the difference seems to be around factor 1.5. Updating
the GUI takes much more time. Therefore the GUI  lag is quite the
same.

Perform the test as described in the ticket.

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,5 @@
 V1.XX.X
+[QMS-220] "Select items from map" misses items
 [QMS-275] Routino: Add Spanish and Czech as selectable languages for turn instructions
 [QMS-279] Track metrics not updated when using UNDO / REDO in Edit mode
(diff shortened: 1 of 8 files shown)


kiozen:
@pingurus please have a look.

pingurus:
A quick test worked as expected.

Issue #307, opened 2021-01-18
Optimization of POI query


pingurus:
**Important Notice:**

This is for developers only. As a user use the `bug report` or `Everything that's not a bug` template.

**The parent ticket**

_add ticketnumber behind the #_

QMS-#294

**What do you plan to do**

@kiozen wrote:
> The search has to be optimized. There should be a sweet spot for the area covered by each query. If it is too small you ahve too many queries. If it is too large queries take too long. There might be also some potential in optimizing the SQL query. Or by creating additional temp tables at startup. I am no database specialist but I think there is a lot of potential in how to query a database.

**The result will be**

- [ ] The query should be considerably faster.

pingurus:
I am thinking about whether it would be better to not query for the attributes in the SQL query. As far as I've understood the data column is not indexed in the SQLite file, so it is comparably slow to search for items in that column.

My idea would be to load all POIs from a 'tile' (the size may be changed for better speed) and then do a bucket sort into the existing map.
**Advantages:**
* Independent of how many categories are activated, there is only one query per tile executed

**Disadvantages**
* Initial loading may be a bit slower than before
* Memory consumption is drastically higher

I don't think that going half the way (querying every active category at once) would be better, since we would then have to also do a bucket sort, since all POIs are returned as one batch.

Just as @kiozen I am no database expert (This is the first time I really do work with SQL databases) so my reasoning may be flawed.

kiozen:
Always when I am confused I start to ask the Internet. Most times I end up even more confused. But sometimes it helps :smile: 

Maybe it's a good idea to see how Osmand does a query. And the problem is not new therefor it's very likely that solutions have been discussed somewhere.

mitxel-m:
@pingurus  To be honest, I'm not an expert, just a DB user. 
I have taken a look at the structure of a POI file, and there is a 'poi_index" table that contains the lat/lon max and min values.
It contains exactly the same number of rows as the 'poi_data" table and I assume both tables can be realtioned by their 'id' field.

May be you could  search first in the poi_index table   with a WHERE clausule and JOIN with the poi_data
```
select poi_index.*, poi_data.data from poi_index inner join poi_data on poi_index.id=poi_data.id WHERE minLat >= 40.00 AND maxLat <= 40.50 AND minLon >= 4.00 AND maxLon <= 4.5 
```
_That statment worked fine when I tested with the POI file for Spain. ( found 267 matches in 336ms ,from 1267649 total rows )_

[...]

PR #309, opened 2021-01-22
[QMS-308] Enable user to copy POIs into project

pingurus:
_**by your answer**_

**What is the linked issue for this pull request (start with a `#`):** QMS-#

I added the possibility to add POIs via right-click to a project.
Also, you can select multiple Items from the map which will also include POIs.

Single POIs:
1. Open poi file and load pois
2. right-click POIs
3. See it is added with name and description

Multiple POIs:
1. Open poi file and load pois
2. right-click map
3. select "Select Items On Map"
3. Select an area with POIS
4. Copy it over to a project
5. See that you get asked for the icon type if there is no automatic match

- [ ] yes

- [ ] yes


pingurus:
@kiozen Should I add the POI Icons to the WptIconManager or should I create some kind of shortcut, so it doesn't clutter the icon selection?

kiozen:
I think we do not get around coping that stuff as Windows does not cope with links. But I might be wrong. On the other hand side, let's try and maybe @wthaem can report back.

pingurus:
I'm not sure if there wasn't a misunderstanding.. I didn't mean to copy over the files to a different folder. 

What I was talking about is whether I should add the icons for the POIs to `CWptIconManager::wptIcons` or not. Adding them would mean that they pop up in the selection of the possible icons for a waypoint. As that would be ~250 new icons that menu would then fill the screen. I'm not sure if that is really desirable. 

The alternative would be to create code to hand a waypoint an icon that is not in  the QMap and then hand it the icon of the POI directly.

kiozen:
Ok, now I get you concerns.

And there is another one: To add icons into the internal list the name must be a name qualified by Garmin.  I doubt that all categories will have a match. But we have to stick to that rule.

Here is a list of available icons and their name.

https://www.javawa.nl/wpsymbols.html

Or maybe you find some already existing mapping in the OSM world.

pingurus:
I am trying to figure out what garmin categories to use, but have two issues:
1. In the current icons list there are icons which are not supported according to the list you linked (either the key does not show up or it is spelled differently). Is there something more to it that I missed?
2. Should I use the same icons that are used in the display of the POIs or would the user then be confused if then weir and dam have the same symbol as there is only the dam in the list you linked?

kiozen:
1. Yes you have to translate the name/keys. 
2. For simplicity I would use the  POI icons if you have to add something to the waypoint list. I mean you can draw your own icon, too. But I doubt you want to spend time on that.

Anyway you can the following:
* For each category have a category name that can be a filename, too. No funny characters like`:` etc. 
* Check if this category name is part of the waypoint list. If it is use it.
* If it is not, lookup the category name for an alternative name that is part of the built-in waypoint list.

By that users can install the POI icons as custom icons. If they do, they get the same icons as in the POI database. If not they get alternative somehow meaningful icons.

pingurus:
I still don't get how I should handle the categories that are accepted by garmin..
What I've found so far:
[Here](https://github.com/Maproom/qmapshack/blob/03496ad8475624ad04847290396501378a0a59af/src/qmapshack/helpers/CWptIconManager.cpp#L66) the list of icons available to the user starts. It contains icons that seem not to be supported by garmin. However, only for the tcx files I find where these icons are filtered (namely [here](https://github.com/Maproom/qmapshack/blob/03496ad8475624ad04847290396501378a0a59af/src/qmapshack/gis/tcx/serialization.cpp#L60))

Should I fill the useful categories from [the list you linked](https://www.javawa.nl/wpsymbols.html) into `CWptIconManager::wptIcons` and assign them symbols from the POI set? Is there any need to adjust the `knownSymbols` list for the .tcx files? And is there a map that I did not find yet that converts the keys in `CWptIconManager::wptIcons` to something garmin understands?

kiozen:
Garmin understands just the symbol name, e.g. `Parking Area`. The devices will attach an icon to a waypoint with that symbol name. 

Alternatively you can store `*.bmp` files on the device. The device will add the base filenames to the table of known symbol names. QMS does the same. e.g. if I store a file `Parking 2.bmp` I can use `Parking 2` as symbol name.

What does that mean for you:

* You add missing icons to the list of known symbols by adding a line like:
 `wptIcons["Parking Area"]        = icon_t("://icons/cache/32x32/parking.png", 16, 16);`

* If you can not find a matching well known Garmin symbol you have to ask the user to select one from the list of known ones.

However if the user installed some own icons with the matching category names from the POI files it would be possible to use those ones. Of course this only works for the devices if the user also installed these icons on the device.

kiozen:
For TCX no changes are necessary. The TCX format is limited to the few icons in the list of `knownSymbols`.

pingurus:
> Garmin understands just the symbol name, e.g. `Parking Area`. The devices will attach an icon to a waypoint with that symbol name.
> 
> Alternatively you can store `*.bmp` files on the device. The device will add the base filenames to the table of known symbol names. QMS does the same. e.g. if I store a file `Parking 2.bmp` I can use `Parking 2` as symbol name.
> 
> What does that mean for you:
> 
>     * You add missing icons to the list of known symbols by adding a line like:
>       `wptIcons["Parking Area"]        = icon_t("://icons/cache/32x32/parking.png", 16, 16);`
> 
>     * If you can not find a matching well known Garmin symbol you have to ask the user to select one from the list of known ones.
> 
> 
> However if the user installed some own icons with the matching category names from the POI files it would be possible to use those ones. Of course this only works for the devices if the user also installed these icons on the device.

Do I then understand it correctly that I could in theory add every icon we use in the display of the POI file to that list and then it would be added to the device by QMS? That would then leave it up to me to condense the huge amount of categories into a few that are sensible to use? Or should I nevertheless stick to the categories defined by garmin? The user dialog would then be the corner case for categories that I cannot assign to a specific category that I selected sensibly?

pingurus:
Meanwhile I've been trying to implement the code to select multiple POIs at once. I intended to use the same type poi_t as I had already used for selecting a single POI (which is the type also used by the garmin maps). However, the location of the POI is there given in pixels, which become invalid once the user wants to select items that are out of the screen (like #220). Is there a specific reason why this is in pixels or would it be sensible to transform this to degrees (or radians)?

kiozen:
"Do I then understand it correctly that I could in theory add every icon we use in the display of the POI file to that list and then it would be added to the device by QMS?"

No. QMS does not put anything to the device. It simple sends a GPX file with the `<sym>` tag of the waypoint set to a string. If that string is recognized by the device everything is fine. If not the devices does whatever it think is appropriate.

However users might store additional `*.bmp` files to add their special icons. QMS can't assume that. But it provides the same mechanism. By that users can have their custom icons on the device and in QMS.

You have to stick to the well known Gamin icons in the first place. You might test for better custom icons in the second place. But without those you have to use one of the well known Garmin icons. If no matching is available the user has to choose from one of the available ones. 

"Meanwhile I've been trying to implement the code to select multiple POIs at once. I intended to use the same type poi_t as I had already used for selecting a single POI (which is the type also used by the garmin maps). However, the location of the POI is there given in pixels, which become invalid once the user wants to select items that are out of the screen (like #220). Is there a specific reason why this is in pixels or would it be sensible to transform this to degrees (or radians)?"

This was done for performance sake. But if it doesn't fit it has to be changed.

kiozen:
Btw I would enhance the item selection by a 5th group of POIs. Do not add another selection mechanism.

pingurus:
> Btw I would enhance the item selection by a 5th group of POIs. Do not add another selection mechanism.

Yes, that's what I'm doing

wthaem:
I had a first look at the new POI feature (using the commit where a POI can be copied to a waypoint).

Great idea on the way of implementation!

Here is a list of possible enhancements (not knowing, what is still in the pipeline) and questions:

* Show the icon for the POI category in the POI selection window next to the category name. The meaning of an icon in the map window is not always obvious.
* In the POI database names of POI are provided in different languages. Which one is selected for use in QMS? QMS language? System language? 
* Add the usual tooltip info when hovering over a POI on the map.
* What happens, if some POI have the same location? There a nice solution for waypoints in this situation.
* A serious issue seems to be the time necessary for loading POIs (even when outside the visible area of the map). The GUI is blocked. A "busy" cursor appears only after some further clicks. This issue is similar to issue #310 - the user doesn't get sufficient information that he has to wait and how long he has to wait. If there is the default cursor, then there is no information at all that QMS is still busy. Waiting several minutes (and this happened even with a small selection of POI) for the next GUI interaction is hardly accepted by the user.
* Would it be possible to add the number of POIs per category to the POI selection list? Then the user would get an early information about the workload he is going to select.
* The blue circle appearing when moving the mouse on a POI is rarely visible on dark icons. If icons are very close one to the other, then it is difficult to understand which icon is hit. Proposal: make the blue circle bigger than the icon.

pingurus:
@wthaemelt As you are trying to display complete countries, I'd move fixing the loading issues to #307. I still wonder that it takes that long for you, since loading all camp sites in the Baden-Württemberg map (~1000) is noticable, but not really problematic.

pingurus:
@kiozen , @wthaemelt could you please try the copying of icons to projects?

src/qmapshack/canvas/CCanvas.cpp
```diff
@@ -1119,9 +1119,20 @@ void CCanvas::displayInfo(const QPoint& px)
 
 poi_t CCanvas::findPOICloseBy(const QPoint& px) const
 {
+    poi_t poiItem = poi->findPOICloseBy(px);
```
kiozen, line 1122: This is blocking the main thread operating the GUI. Probably it gets stuck in the mutex of the POI thread. This is no good.

kiozen:
I tried. GUI is blocking as already mentioned. When I copy single POIs icon matching seems to work. When I am selecting POIs via mouse selection I have to provide icons for the same category. I also had the case where I selected 3 POIs but the on screen table said it's just 1. 3 where copied.

kiozen:
This PR gets again quite large. Can we convert the low hanging fruits like filling up the waypoint table in it's own PR.
And split single POI copy and copy by selection. 

It's really better to get small topics done, review and tested. Instead of having all in one with multiple issues to tackle.

wthaem:
* "Whole countries": I just selected POI files from the Mapsforge list. Neither for Ethiopia nor for Crimea I can download POI files for smaller regions. And the POI file size is certainly not big: Ethiopia ~3.1MB, Crimea ~8.8 MB. Russia North-Western district (Skt.Petersburg) would be 55.4 MB - much bigger. 

* What does mean "Copying of icons"? I copied a POI to a waypoint in a project. The icon shown for this new waypoint is my default icon. What should i do to get a different one?

pingurus:
> What does mean "Copying of icons"? I copied a POI to a waypoint in a project. The icon shown for this new waypoint is my default icon. What should i do to get a different one?

Yes, I meant copying of POIs, sorry. I also updated the way to test this PR on top.
For single POIs you can select a different one just as you would change the waypoint icon. If you copy multiple icons it should as you for the default name and the default icon for each category.

wthaem:
When selecting GIS objects with the context menu "Select items on map" in the map window, then the pop-up help text for activating the POI selection reads "Add area to selection". Must be changed to "Add POI to selection".

kiozen:
Do we still need that PR? I think it's obsolete, isn't it?

Issue #317, opened 2021-01-29
Enable user to add POI to project via right-click


pingurus:
**Important Notice:**

This is for developers only. As a user use the `bug report` or `Everything that's not a bug` template.

**The parent ticket**

_add ticketnumber behind the #_

QMS-#294

**What do you plan to do**

Enable user to add POI to project via right-click

**The result will be**

User can add POI via right-click to project. No automatic icon matching yet.

Issue #308, opened 2021-01-22
Enable user to copy POIs into project


pingurus:
**Important Notice:**

This is for developers only. As a user use the `bug report` or `Everything that's not a bug` template.

**The parent ticket**

_add ticketnumber behind the #_

QMS-#294

**What do you plan to do**

It is useful for the user to copy  POIs into a project, for example to then export it. This issue is about enabling the user to do so.

**The result will be**

- [x] As is with garmin maps, a right-click and 'add waypoint' on a POI should lead to a waypoint created that contains the data (and preferably the icon) of this poi
- [x] The user should be able to select multiple POIs to copy with the respective tool

pingurus:
The desired results have been achieved.

PR #320, opened 2021-01-29
[QMS-317] Enable user to add POI to project via right-click

pingurus:
_**by your answer**_

**What is the linked issue for this pull request (start with a `#`):** QMS-#317

I added the possibility to add POIs via right-click to a project.

1. Open poi file and load pois
2. right-click POIs
3. See it is added with name and description

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/poi/CPoiDraw.cpp
+++ b/src/qmapshack/poi/CPoiDraw.cpp
@@ -17,6 +17,7 @@
 **********************************************************************************************/
 
 #include "CMainWindow.h"
+#include "gis/Poi.h"
 #include "helpers/CSettings.h"
 #include "poi/CPoiDraw.h"
 #include "poi/CPoiItem.h"
@@ -135,6 +136,34 @@ void CPoiDraw::savePoiPath(QSettings &cfg)
 void CPoiDraw::loadPoiPath(QSettings &cfg)
 {
     poiPaths = cfg.value("poiPaths", poiPaths).toStringList();
+}
+
+poi_t CPoiDraw::findPOICloseBy(const QPoint &px) const
+{
+    poi_t poi;
+    if(CPoiItem::mutexActivePois.tryLock() && poiList)
+    {
+        for(int i = 0; i < poiList->count(); i++)
+        {
+            CPoiItem* item = poiList->item(i);
+
+            if(!item || item->getPoifile().isNull())
+            {
+                // as all active maps have to be at the top of the list
+                // it is ok to break as soon as the first map with no
+                // active files is hit.
+                break;
+            }
+
+            if(item->findPoiCloseBy(px, poi))
+            {
+                // stop at the 1st one found
+                break;
+            }
+        }
+        CPoiItem::mutexActivePois.unlock();
+    }
+    return poi;
 }
 
 void CPoiDraw::buildPoiList()
--- a/src/qmapshack/poi/CPoiPOI.cpp
+++ b/src/qmapshack/poi/CPoiPOI.cpp
@@ -17,6 +17,7 @@
 
 **********************************************************************************************/
 
+#include "gis/Poi.h"
 #include "helpers/CDraw.h"
 #include "poi/CPoiCategory.h"
 #include "poi/CPoiDraw.h"
@@ -109,7 +110,7 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
                     mutex.unlock();
                     return;
                 }
-                for(const poi_t& poiToDraw : loadedPOIs[category][minLonM10][minLatM10])
+                for(const rawPoi_t& poiToDraw : loadedPOIs[category][minLonM10][minLatM10])
                 {
                     QPointF pt = poiToDraw.coordinates;
                     poi->convertRad2Px(pt);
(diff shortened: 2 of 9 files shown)


src/qmapshack/poi/CPoiDraw.cpp
```diff
@@ -137,6 +138,36 @@ void CPoiDraw::loadPoiPath(QSettings &cfg)
     poiPaths = cfg.value("poiPaths", poiPaths).toStringList();
 }
 
+poi_t CPoiDraw::findPOICloseBy(const QPoint &px) const
+{
+    poi_t poi;
+    CPoiItem::mutexActivePois.lock();
```
kiozen, line 144: Use `QMutexLocker`. Much safer.

src/qmapshack/poi/CPoiDraw.cpp
```diff
@@ -137,6 +138,36 @@ void CPoiDraw::loadPoiPath(QSettings &cfg)
...
+    if(poiList)
+    {
```
kiozen, line 146: With `QMutexLocker` you could simplify the code:

```
if(!poiList)
{
    return poi;
}
``

File: src/qmapshack/poi/CPoiDraw.cpp [hunk 3]
```diff
@@ -137,6 +138,36 @@ void CPoiDraw::loadPoiPath(QSettings &cfg)
...
+        for(int i = 0; i < poiList->count(); i++)
+        {
+            auto item = poiList->item(i);
```
kiozen, line 149: I do not really like auto. Sometimes it's really nice however IDEs can't resolve `auto`. This makes the code hard to read. Please stick with explicit types as long as possible.

src/qmapshack/poi/CPoiDraw.cpp
```diff
@@ -137,6 +138,36 @@ void CPoiDraw::loadPoiPath(QSettings &cfg)
...
+
+            if(!item || item->getPoifile().isNull())
+            {
+                // as all active maps have to be at the top of the list
+                // it is ok to break as soon as the first map with no
+                // active files is hit.
+                break;
+            }
+
+            item->getPoifile()->findPOICloseBy(px, poi);
```
kiozen, line 159: The only API exposed by item should be the one of `CPoiItem`. At this point I do not care if there is a poi file or what ever. I call `item->findPoiCloseBy()` and let the item do what ever it has to do.

src/qmapshack/poi/CPoiDraw.cpp
```diff
@@ -137,6 +138,36 @@ void CPoiDraw::loadPoiPath(QSettings &cfg)
...
+            if(poi.pos != NOPOINTF)
```
kiozen, line 160: `findPOICloseBy()` could have a boolean return value to tell if the search has been successful.

src/qmapshack/poi/CPoiPOI.cpp
```diff
@@ -126,6 +127,43 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
...
+        {
+            QPointF pt = poiItemFound.coordinates;
+            poi->convertRad2Px(pt);
+            QPointF x = px - pt;
+            if(x.manhattanLength() < 10)
+            {
+                poiItem.pos = poiItemFound.coordinates;
+                for(auto tag : {"name=", "brand=", "operator="})
+                {
+                    auto matches = poiItemFound.data.filter(tag);
```
kiozen, line 149: again `auto` is a bad choice. Everyone has to lookup the type of data and the return type of filter to understand that auto is in fact a `QStringList`. It should be a `const QStringList&` and I am not sure if auto does that.

src/qmapshack/poi/CPoiPOI.cpp
```diff
@@ -126,6 +127,43 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
...
+                    if (matches.length() > 0)
```
kiozen, line 150: well I would use `isEmpty()` as it is what you want to know. But ok, that's a matter of taste.

kiozen (changes requested):
My main concern with this MR is the fact that it is blocking the GUI if the POI thread is busy querying the database. And as it is busy each time I zoom/move the map the user will run into the problem of a blocking GUI quite frequently. 

Either you 
optimize the database access to happen in fractions of a second 

or 

you find a way to unlock the mutex on a very regular, frequent base

or 

you find another solution (e.g. do not search for POIs if the mutex would block)

pingurus:
> My main concern with this MR is the fact that it is blocking the GUI if the POI thread is busy querying the database. And as it is busy each time I zoom/move the map the user will run into the problem of a blocking GUI quite frequently.
> 
> Either you
> optimize the database access to happen in fractions of a second
> 
> or
> 
> you find a way to unlock the mutex on a very regular, frequent base
> 
> or
> 
> you find another solution (e.g. do not search for POIs if the mutex would block)

I'm not sure if I understand you correctly, but nothing in this PR triggers access to the database. There is only a lookup in the table of loaded POIs.

The blocking of the user interface already occurs during the loading of the POIs to be displayed. This should either have been fixed in #302 (well, to late..), or when tackling #307. If you prefer, I can first work on #307 before proceeding with the tasks that emerged from #308 / #309.

kiozen:
So far the POI loading takes a long time but since it is running in it's own thread and nothing competes it just takes time but does not block the GUI.

Now you introduce the POI lookup locking `mutexActivePois` which is totally ok. However while the POI thread loads POIs in the background this mutex will block until the mutex is unlocked by the POI thread. As the method is called in the main thread - the GUI - thread the complete GUI is blocked. In an event driven GUI application all methods called must not block for a longer timer. 

As this PR introduces the problem it has to be solved by this PR or we prioritize another PR to solve it first.

pingurus:
I've made some benchmarking on the current feature/pois branch for the `CPOIPoi::draw()` method. The results for loading all POIs from the water category in the Baden-Wuerttemberg map (where I can see the effect described when using the branch of this PR) are the following: 
```
total Time: 8965 ms
locked Time: 3 ms
unlocked Time: 5996 ms
missing Time: 2966 ms
mean of unlockedTimes: 0.7305068226120858 ms
standard deviation of unlockedTimes: 1.7457090631014056 ms
min of unlockedTimes: 0 ms
max of unlockedTimes: 32 ms
```
For me this looks like the mutex is regularly unlocked while loading the POIs from the sql file. So I think the best solution to fix the hanging of the GUI is to do a `CPoiItem::mutexActivePois.tryLock();` and to skip checking for POIs if it can't lock the mutex.

src/qmapshack/poi/CPoiDraw.cpp
pingurus: Should I somehow restrict access to `CPOIPoi::findPOICloseBy` or is it enough to just provide a `CPoiItem::findPOIClosBy`?

pingurus:
Loading of the POIs is still slow, but it no longer blocks the GUI.

src/qmapshack/poi/CPoiDraw.cpp
kiozen: Imho it's enough to provide a `CPoiItem::findPOIClosBy()`.

src/qmapshack/poi/CPoiPOI.cpp
```diff
@@ -126,6 +127,45 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
...
+                {
+                    const QStringList& matches = poiItemFound.data.filter(tag);
+                    if (!matches.isEmpty())
+                    {
+                        poiItem.name = matches[0];
+                        poiItem.name.replace(tag, "");
+                        break;
+                    }
+                }
+                if(poiItem.name == "")
```
kiozen, line 157: `isEmpty()` is always preferred over comparison with a null string. I would even expect it to be more performant.

kiozen (changes requested):
GUI interaction is much better now. And I guess it's acceptable for the user not  to select POIs while rendering is still in progress. Hopefully we can optimize that.

Only two minor cosmetic nitpicks left. I would merge when they are done if that's ok for you.

Issue #321, opened 2021-01-29
Improve name selection for POI


wthaem:
**Important Notice:**

**Do not delete any of the sections. Read what is requested.**
**_Replace(!)_ the instructions by your _answer(!)_. Feature requests**
**reports that are not properly filed might be closed without**
**further notice.**

**Choose a good title**

What is a good title?
The title will be listed in the changelog when the issue is fixed. Do
you think someone else would understand what's fixed or done from your
title? Does it really name the real topic of the issue?

[ x] Did you choose a good title?

**Is your feature request related to a problem? Please describe.**

In a POI file a POI may have several name fields, e.g. depending on language choice. One of them is `alt_name` (alternate name?).

Currently, it may happen, that QMS displays the `alt_name` field content, maybe, with an additional `alt_` prefix. It would certainly  be better to use the `name` field content (compare attached snapshot, left side: DB browser content, right side: QMS workspace/POI window).

**Describe the solution you'd like**

Set the value of the name field as name of the QMS waypoint obtained from a POI.

In the case, that the `name` field is missing or doesn't have a value (there are such cases!), use the value of one of the other `name:xx` fields, preferrably, `xx=en` or `xx=QMS language`. A QMS waypoint should have a name! Otherwise, the items list in the workspace looks strange!

For the description text in the waypoint, try to keep/insert linebreaks as seen in the DB Browser, otherwise information is difficult to read.

**Describe alternatives you've considered**

**Give a reason why someone else should spend time on that request**

Keep in mind that you ask others to spend their spare time for your request. Convince them
by giving a reason why solving your request has a benefit for them, too. Good reasons are
if you are able to prove

* The request solves a fundamental problem for everyone
* The request is of general interest for many users

Bad reasons are:
* Because software xyz does it
* Because you just want it to have that way
* Because it will solve a very particular problem.

**Additional context**

![POINames](https://user-images.githubusercontent.com/55309218/106283444-3b77c280-6242-11eb-8bf5-7c2e63b1d47f.png)

pingurus:
Good catch for the `alt_name=` case, I wasn't aware of that possible tag.

> In the case, that the name field is missing or doesn't have a value (there are such cases!), use the value of one of the other name:xx fields, preferrably, xx=en or xx=QMS language. A QMS waypoint should have a name! Otherwise, the items list in the workspace looks strange!

Currently it should default to the name of the category, if there is no name recognized. Are the language abbreviations of the languages the same in QMS and OSM?

[...]

PR #325, opened 2021-02-02
[QMS-322] Make POIs show up in 'Select Items On Map'

pingurus:
* Add possibility for user to select multiple POIs to copy to project

_**Note: Do not delete any of the sections**_
_**Answer them all. Replace the descriptive text**_
_**by your answer**_

**What is the linked issue for this pull request (start with a `#`):** QMS-#322

**Describe roughly what you have done:**

I added the possibility to copy over multiple Items to a project.

**What steps have to be done to perform a simple smoke test:**

1. Open poi file and load POIs
2. right-click map
3. select "Select Items On Map"
4. Select an area with POIs
5. Copy it over to a project

**Does the code comply to the coding rules and naming conventions [Coding Guidelines](https://github.com/Maproom/qmapshack/wiki/DeveloperCodingGuideline):**

- [x] yes

**Is every user facing string in a tr() macro?**

- [x] yes

**Did you add the ticket number and title into the changelog? Keep the numeric order in each release block.**

- [x] yes, I didn't forget to change changelog.txt

PR #327, opened 2021-02-03
[QMS-321] Improve name selection for POI

pingurus:
_**by your answer**_

**What is the linked issue for this pull request (start with a `#`):** QMS-#321

I improved the name selection process. The assumption is made that when there is an `alt_name` that there would be a better 'normal' name, if that is not the case, I can add alt names to the selection process, however there seems to be disagreement on how to add those in OSM (I've seen `alt_name=`, `alt_name1=` and `alt_name_1=`). Also there are `loc_name` `short_name` `official_name` `reg_name`

1. Select POI with different name possibilities to copy to workspace
2. See that your locale is the first choice, then English, then the default name and then any other name. 

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/poi/CPoiDraw.cpp
+++ b/src/qmapshack/poi/CPoiDraw.cpp
@@ -141,7 +141,7 @@ void CPoiDraw::loadPoiPath(QSettings &cfg)
 poi_t CPoiDraw::findPOICloseBy(const QPoint &px) const
 {
     poi_t poi;
-    if(CPoiItem::mutexActivePois.tryLock() && poiList)
+    if(poiList && CPoiItem::mutexActivePois.tryLock())
     {
         for(int i = 0; i < poiList->count(); i++)
         {
@@ -168,7 +168,7 @@ poi_t CPoiDraw::findPOICloseBy(const QPoint &px) const
 
 void CPoiDraw::findPoisIn(const QRectF &degRect, QList<poi_t> & pois) const
 {
-    if(CPoiItem::mutexActivePois.tryLock() && poiList)
+    if(poiList && CPoiItem::mutexActivePois.tryLock())
     {
         for(int i = 0; i < poiList->count(); i++)
         {
@@ -184,8 +184,35 @@ void CPoiDraw::findPoisIn(const QRectF &degRect, QList<poi_t> & pois) const
 
             item->findPoisIn(degRect, pois);
         }
+        CPoiItem::mutexActivePois.unlock();
     }
-    CPoiItem::mutexActivePois.unlock();
+}
+
+bool CPoiDraw::getToolTip(const QPoint &px, QString &str)
+{
+    if(poiList && CPoiItem::mutexActivePois.tryLock())
+    {
+        for(int i = 0; i < poiList->count(); i++)
+        {
+            CPoiItem* item = poiList->item(i);
+
+            if(!item || item->getPoifile().isNull())
+            {
+                // as all active maps have to be at the top of the list
+                // it is ok to break as soon as the first map with no
+                // active files is hit.
+                break;
+            }
+
+            if(item->getToolTip(px, str))
+            {
+                CPoiItem::mutexActivePois.unlock();
+                return true;
+            }
+        }
+        CPoiItem::mutexActivePois.unlock();
+    }
+    return false;
 }
 
 void CPoiDraw::buildPoiList()
--- a/changelog.txt
+++ b/changelog.txt
@@ -12,6 +12,7 @@ V1.XX.X
 [QMS-315] Make POI Icons User-Selectable
 [QMS-317] Enable user to add POI to project via right-click
 [QMS-318] Change poi_t to use radians
+[QMS-321] Improve name selection for POI
 [QMS-322] Make POIs show up in 'Select Items On Map'
 
 V1.15.2
--- a/src/qmapshack/canvas/CCanvas.cpp
+++ b/src/qmapshack/canvas/CCanvas.cpp
@@ -910,10 +910,13 @@ void CCanvas::slotTriggerCompleteUpdate(CCanvas::redraw_e flags)
 void CCanvas::slotToolTip()
 {
     QString str;
-    map->getToolTip(posToolTip, str);
-    if(str.isEmpty())
+    if(!poi->getToolTip(posToolTip, str))
     {
-        return;
+        map->getToolTip(posToolTip, str);
+        if(str.isEmpty())
+        {
+            return;
+        }
     }
     QPoint p = mapToGlobal(posToolTip + QPoint(32, 0));
     QToolTip::showText(p, str);
--- a/src/qmapshack/poi/CPoiDraw.h
+++ b/src/qmapshack/poi/CPoiDraw.h
@@ -54,6 +54,7 @@ public:
 
     poi_t findPOICloseBy(const QPoint& px) const;
     void findPoisIn(const QRectF& degRect, QList<poi_t>& pois) const;
+    bool getToolTip(const QPoint& px, QString& str);
 protected:
     void drawt(buffer_t& currentBuffer) override;
 
--- a/src/qmapshack/poi/CPoiItem.h
+++ b/src/qmapshack/poi/CPoiItem.h
@@ -91,6 +91,10 @@ public:
     {
         getPoifile()->findPoisIn(degRect, pois);
     }
+    bool getToolTip(const QPoint& px, QString& str)
+    {
+        return getPoifile()->getToolTip(px, str);
+    }
 private:
     friend class CPoiDraw;
     CPoiDraw * poi;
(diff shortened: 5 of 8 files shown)


pingurus:
I'll keep this a draft until #325 is merged, but feel free to test @wthaem

wthaem:
Part of the proposal implemented. Thanks. But there are still strange/exotic data situations as shown in the attached snapshots. My proposal covered these situations, too. Is it possible to do still a bit more on improving the waypoint names? If it is hard to implement, leave it. It is a rare and not the typical data structure I'm mentioning here.


names2
names1

pingurus:
I assume those are non ASCII letters that are missing, right?
I thought I enabled Unicode support for the regexes, which also seems to have worked, as there is a lot of whitespace. @kiozen is there something I need to take care of so these strings are also displayed?

wthaem:
Right you are! I'm a bit surprised: yes, it is Unicode! Neither the DB Browser nor QMS display the Amharic characters as such (they are obviously replaced by whitespace when rendering them). But this issue discussion page shows them correctly:

amenity=hospital
name=Amanuel-Hospital
name:am=አማኑኤል ሆስፒታል
name:en=አማኑኤል ሆስፒታል
phone=+251112757715
source=YAHOO! aerial imagery

No idea, whom to blame for this! Maybe, Unicode handling is restricted to some part of the complete Unicode table. I think, I saw somewhere a parameter like "maximum Unicode code used" what could explain the issue.

Knowing this result, it is certainly easy to find other languages creating similar trouble. But I believe, it is not necessary to spent much time to deal with this trouble.

I copied the "አማኑኤል ሆስፒታል" from this page, inserted it into the QMS search field, saw whitespace with the cursor at the end of the whitespace and got the waypoint I was looking for and not viewing the name of the waypoint. Same behavior as in the DB browser.

BTW: I was wondering, why I couldn't find some Amharic names in the POI file. Now, I know the reason.

kiozen:
@pingurus you might want to play with 
`QTextCodec::setCodecForLocale(QTextCodec::codecForName("UTF-8"));`

wthaem:
Attached is a GPX file with some of the discussed "unnamed" waypoints. Looking into this GPX file with a Unicode-capable editor you see the correct names. Loading this file in the workspace results again in "unnamed" waypoints. Thus, not a POI-specific issue
, but a more general one.

[Unicode.zip](https://github.com/Maproom/qmapshack/files/5927086/Unicode.zip)

kiozen:
> Attached is a GPX file with some of the discussed "unnamed" waypoints. Looking into this GPX file with a Unicode-capable editor you see the correct names. Loading this file in the workspace results again in "unnamed" waypoints. Thus, not a POI-specific issue
> , but a more general one.

I tried it on my Linux box. I can see all names. Could it be that Windows still has it's problems with UTF-8 in 2021? Or maybe the font you are using is not full UTF-8.

pingurus:
> > Attached is a GPX file with some of the discussed "unnamed" waypoints. Looking into this GPX file with a Unicode-capable editor you see the correct names. Loading this file in the workspace results again in "unnamed" waypoints. Thus, not a POI-specific issue
> > , but a more general one.
> 
> I tried it on my Linux box. I can see all names. Could it be that Windows still has it's problems with UTF-8 in 2021? Or maybe the font you are using is not full UTF-8.

On my pretty vanilla us-locale Ubuntu 20.04 the names are also displayed correctly.

kiozen:
I would like to see the name attached to the POI icon. If that turns out to be too chaotic we should at least have an information bubble when hoovering over the POI with the mouse displaying the name and content.

pingurus:
> I would like to see the name attached to the POI icon. If that turns out to be too chaotic we should at least have an information bubble when hoovering over the POI with the mouse displaying the name and content.

You mean like it is with the waypoints?

kiozen:
> > I would like to see the name attached to the POI icon. If that turns out to be too chaotic we should at least have an information bubble when hoovering over the POI with the mouse displaying the name and content.
> 
> You mean like it is with the waypoints?

Yes. Probably we have to to the same de-cluttering for POIs in the same spot etc.

Right now I do not get any information until I converted the POI into a waypoint.

wthaem:
If everything works fine on Linux system, then certainly Windows or Qt for Windows(?) might be the cause of the trouble. 

I know 4 Qt applications where rendering of exotic characters fails: QMS, DB Browser,  **Qt assistant**, my own PyQt5 based help browser.

And I know applications that work correctly: Firefox, Chrome, Microsoft Edge, Microsoft  VS code, Notepad++ editor, ...

Looks like Qt - isn't it? Does somebody know how to convince Qt to make better use of Unicode?

pingurus:
> > > I would like to see the name attached to the POI icon. If that turns out to be too chaotic we should at least have an information bubble when hoovering over the POI with the mouse displaying the name and content.
> > 
> > 
> > You mean like it is with the waypoints?
> 
> Yes. Probably we have to do the same de-cluttering for POIs in the same spot etc.
> 
> Right now I do not get any information until I converted the POI into a waypoint.

Shouldn't we put this into a seperate issue and PR?

kiozen:
Placing the name and providing additional information would fit in here. De-cluttering is another topic. I could also handle the de-cluttering as it has to fit into the rest.

pingurus:
I added the name as a tooltip. What additional information did you have in mind? I don't think that it should be much, so I'd add only the category. Of course the best would be some category sensitive information, but I doubt that that's worth the work.

src/qmapshack/poi/CPoiDraw.cpp
```diff
@@ -188,6 +188,32 @@ void CPoiDraw::findPoisIn(const QRectF &degRect, QList<poi_t> & pois) const
     CPoiItem::mutexActivePois.unlock();
 }
 
+bool CPoiDraw::getToolTip(const QPoint &px, QString &str)
+{
+    if(CPoiItem::mutexActivePois.tryLock() && poiList)
```
kiozen, line 193: :rotating_light: Potential dead lock. Same as in `findPOICloseBy`. If the mutex locks but there is no `poiList` we will leave without unlocking the mutex. 

I just ran into the situation that the thread to retrieve and render all POIs is blocking for ever.

pingurus: So there have to be two nested `if` statements, right? Because unlocking when you haven't locked it isn't good either.

kiozen: hm, but even 
```
if(poiList && CPoiItem::mutexActivePois.tryLock())
```
blocks. The must be more deadlocks. Please crosscheck all your mutex locks!

kiozen (changes requested):

src/qmapshack/poi/CPoiDraw.cpp
```diff
@@ -188,6 +188,32 @@ void CPoiDraw::findPoisIn(const QRectF &degRect, QList<poi_t> & pois) const
...
+            {
+                // as all active maps have to be at the top of the list
+                // it is ok to break as soon as the first map with no
+                // active files is hit.
+                break;
+            }
+
+            if(item->getToolTip(px, str))
+            {
+                return true;
```
kiozen, line 209: ouch there is another one.

src/qmapshack/poi/CPoiDraw.cpp
```diff
@@ -188,6 +188,32 @@ void CPoiDraw::findPoisIn(const QRectF &degRect, QList<poi_t> & pois) const
     CPoiItem::mutexActivePois.unlock();
```
kiozen, line 188: This has to go into the if statement.

kiozen (changes requested):

src/qmapshack/poi/CPoiDraw.cpp
pingurus: Are they fixed now?

kiozen: seems to be ok for now.

pingurus:
@kiozen what should I do for this PR? Is placing the name linked to decluttering or should I copy-paste and adjust the code from the waypoints?

kiozen:
The tooltip is ok for now.

 I am a bit undecided on how sophisticated we want to go. Do we want to raise POIs to the same level as waypoints? This would request to implement stuff similar to IGisItem providing an on screen dialog etc. Or is it fine to have tooltips?

For now I would go for the tooltips. Maybe also add a bit more of informational text if available in the database. But names are ok, too.

If you do not want to add more I would start another round of review to get it merged.

pingurus:
I changed the tooltip to display the name in bold and then all the data underneath. Like this you have all information you can have without copying the POI over to a waypoint, so I wouldn't add anything and you can start a review to get it merged (Assuming decluttering and name placement is treated as a different task).

kiozen:
You should address the description string in another PR. Right now it's just dumping all tags. It would be nicer to create a real address from the address tags and when converting into a waipoint adding links as real waypoint links. But for the scope of this ticket it's fine.

Issue #313, opened 2021-01-28
UX improvements for POIs


pingurus:
**Important Notice:**

This is for developers only. As a user use the `bug report` or `Everything that's not a bug` template.

**The parent ticket**

_add ticketnumber behind the #_

QMS-#294

**What do you plan to do**
Improve the user experience by feedback provided.

**The result will be**

(Remarks by @wthaemelt in #309 , comments in italics by @pingurus)
* [ ] Show the icon for the POI category in the POI selection window next to the category name. The meaning of an icon in the map window is not always obvious.
* [x] _(probably wontfix)_ In the POI database names of POI are provided in different languages. Which one is selected for use in QMS? QMS language? System language? _Whatever the language in the 'name=' tag is. Should be english._
* [x] Add the usual tooltip info when hovering over a POI on the map.
* [x] What happens, if some POI have the same location? _Right now nothing. Also note that some POIs fall in multiple categories_ There a nice solution for waypoints in this situation.
* [ ] A serious issue seems to be the time necessary for loading POIs (even when outside the visible area of the map). The GUI is blocked. A "busy" cursor appears only after some further clicks. This issue is similar to issue #310 - the user doesn't get sufficient information that he has to wait and how long he has to wait. If there is the default cursor, then there is no information at all that QMS is still busy. Waiting several minutes (and this happened even with a small selection of POI) for the next GUI interaction is hardly accepted by the user. _Also related to #307_
* [ ] _(probably wontfix)_ Would it be possible to add the number of POIs per category to the POI selection list? Then the user would get an early information about the workload he is going to select. _We would first have to count them, which is exactly the same workload for the SQL Query as showing them_
* [x] The blue circle appearing when moving the mouse on a POI is rarely visible on dark icons. If icons are very close one to the other, then it is difficult to understand which icon is hit. Proposal: make the blue circle bigger than the icon.

kiozen:
my 2 cents:

* the count function of the database should be more efficient than doing it inside our code:

wthaem:
SQL proposal:

```
SELECT COUNT(P.ID) AS CNT, C.CATEGORY
FROM [POI_Data] P JOIN POI_CATEGORY_MAP C
ON P.ID = C.ID
GROUP BY C.CATEGORY
ORDER BY C.CATEGORY
.```

Result with 234598 POI and 331 categories when using my SQLite browser:

    304 Reihen innerhalb von 1205ms zurückgegeben

POI file size 55.4 MB. 

[...]

Issue #337, opened 2021-03-12
Upgrade to Proj 8 API


kiozen:
**Is your feature request related to a problem? Please describe.**

The old Proj API is discontinued with version 8 of the library. We need to upgrade. In the meanwhile porting is possible as all necessary functions are finally in the library. (from 7.1.0 on)

**Describe the solution you'd like**

This is the chance to hide the Proj mayhem in a class simplifying stuff. Features to implement are:

* initialization with different projections. Projections can change.
* testing for lat/long projections
* transform points and polylines (strictly use Qt types QPointF and QPolygonF)
* projection string validity check

**Describe alternatives you've considered**

n/a

**Give a reason why someone else should spend time on that request**

doing it on my own

**Additional context**

n/a

pingurus:
I tried to update proj, but while make check returns no errors, QMS still fails compiling and `pkg-config --modversion proj` still returns `6.3.1`

@kiozen could you please update the wiki in case something changed except the version number?

kiozen:
@pingurus did you make sure that the installation of Proj 8 did update the system wide cmake script to link against the new version.

pingurus:
How can I check that?

kiozen:
check the script in `/usr/lib/cmake...` or where ever your system is storing the stuff

pingurus:
I'm still not able to compile, but through magic `pkg-config --modversion proj` now returns `8.0.0` and qt creator points me to the right file (proj.h where it is stated that this is version 8.0.0). What file exactly should I look for under `/usr/lib/cmake`? There are some folders with more files and even more references to proj4 in there. I'm on Ubuntu 20.04 if that helps.

Shouldn't the following section in `CMakeLists.txt` be updated?
```
if(${PROJ4_VERSION} VERSION_LESS 5.0.0)
    message( SEND_ERROR "You need at least PROJ4 5.0.0 or newer.")
endif()
```

pingurus:
Somehow I managed to build now, I did
```
sudo add-apt-repository ppa:ubuntugis/ubuntugis-unstable
sudo apt-get update
sudo apt upgrade
```
and some reinstallations of `libproj-dev`, `proj-bin` and `proj-data`. I doubt that my packages are in a clean state, but it works for now.

PR #341, opened 2021-03-14
[QMS-313] UX improvements for POIs

pingurus:
_**Note: Do not delete any of the sections Answer them all. Replace the descriptive text by your answer**_

**What is the linked issue for this pull request (start with a `#`):**

QMS-#313

I added links in the waypoint description for some predefined osm tags.

1. Open a poi file
2. Copy a POI over to the workspace
3. See that links, mails and phone numbers are clickable

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/gis/wpt/CGisItemWpt.cpp
+++ b/src/qmapshack/gis/wpt/CGisItemWpt.cpp
@@ -298,7 +298,6 @@ void CGisItemWpt::newWpt(const poi_t &poi, IGisProject *project, bool openEditWI
     const QString& _name = poi.name.isEmpty() ? getLastName("") : poi.name;
 
     CGisItemWpt * wpt = new CGisItemWpt(poi.pos * RAD_TO_DEG, _name, _icon, project);
-    wpt->setReadOnlyMode(true);
     if(!poi.desc.isEmpty())
     {
         wpt->setDescription(poi.desc);
@@ -316,6 +315,7 @@ void CGisItemWpt::newWpt(const poi_t &poi, IGisProject *project, bool openEditWI
         wpt->editInitial();
     }
 
+    wpt->setReadOnlyMode(true);
     cfg.setValue("Waypoint/lastName", wpt->getName());
     cfg.setValue("Waypoint/lastIcon", wpt->getIconName());
 }
--- a/src/qmapshack/gis/wpt/CGisItemWpt.cpp
+++ b/src/qmapshack/gis/wpt/CGisItemWpt.cpp
@@ -24,6 +24,7 @@
 #include "gis/CGisListWks.h"
 #include "gis/GeoMath.h"
 #include "gis/prj/IGisProject.h"
+#include "gis/Poi.h"
 #include "gis/wpt/CDetailsGeoCache.h"
 #include "gis/wpt/CDetailsWpt.h"
 #include "gis/wpt/CGisItemWpt.h"
@@ -273,10 +274,10 @@ bool CGisItemWpt::getIconAndName(QString& icon, QString& name)
     return true;
 }
 
-void CGisItemWpt::newWpt(const QPointF& pt, const QString &name, const QString &desc, IGisProject * project, const QString& icon, bool openEditWIndow)
+void CGisItemWpt::newWpt(const QPointF& pt, const QString &name, const QString &desc, IGisProject * project)
 {
     SETTINGS;
-    const QString& _icon = icon.isEmpty() ? cfg.value("Waypoint/lastIcon", "Waypoint").toString() : icon;
+    const QString& _icon = cfg.value("Waypoint/lastIcon", "Waypoint").toString();
     const QString& _name = name.isEmpty() ? getLastName("") : name;
 
     CGisItemWpt * wpt = new CGisItemWpt(pt, _name, _icon, project);
@@ -284,7 +285,32 @@ void CGisItemWpt::newWpt(const QPointF& pt, const QString &name, const QString &
     {
         wpt->setDescription(desc);
     }
+    wpt->editInitial();
 
+    cfg.setValue("Waypoint/lastName", wpt->getName());
+    cfg.setValue("Waypoint/lastIcon", wpt->getIconName());
+}
+
+void CGisItemWpt::newWpt(const poi_t &poi, IGisProject *project, bool openEditWIndow)
+{
+    SETTINGS;
+    const QString& _icon = poi.icon.isEmpty() ? cfg.value("Waypoint/lastIcon", "Waypoint").toString() : poi.icon;
+    const QString& _name = poi.name.isEmpty() ? getLastName("") : poi.name;
+
+    CGisItemWpt * wpt = new CGisItemWpt(poi.pos * RAD_TO_DEG, _name, _icon, project);
+    wpt->setReadOnlyMode(true);
+    if(!poi.desc.isEmpty())
+    {
+        wpt->setDescription(poi.desc);
+    }
+    if(!poi.links.isEmpty())
+    {
+        wpt->setLinks(poi.links);
+    }
+    if(poi.ele != NOINT)
+    {
+        wpt->setElevation(poi.ele);
+    }
     if(openEditWIndow)
     {
         wpt->editInitial();
@@ -692,7 +718,7 @@ void CGisItemWpt::drawItem(QPainter& p, const QRectF& /*viewport*/, CGisDraw * g
 }
 
 
-void CGisItemWpt::drawLabel(QPainter& p, const QPolygonF &/*viewport*/, QList<QRectF> &blockedAreas, const QFontMetricsF &fm, CGisDraw */*gis*/)
+void CGisItemWpt::drawLabel(QPainter& p, const QPolygonF & /*viewport*/, QList<QRectF> &blockedAreas, const QFontMetricsF &fm, CGisDraw */*gis*/)
 {
     if(flags & eFlagWptBubble)
     {
--- a/src/qmapshack/gis/wpt/CGisItemWpt.h
+++ b/src/qmapshack/gis/wpt/CGisItemWpt.h
@@ -27,6 +27,7 @@
 #include <QCoreApplication>
 #include <QPointer>
 
+struct poi_t;
 class IGisProject;
 class QDomNode;
 class CScrOptWpt;
@@ -348,7 +349,8 @@ public:
     const searchValue_t getValueByKeyword(searchProperty_e keyword) override;
 
     static QString getLastName(const QString &name);
-    static void newWpt(const QPointF &pt, const QString& name, const QString& desc, IGisProject *project, const QString &icon = "", bool openEditWIndow = true);
+    static void newWpt(const QPointF &pt, const QString& name, const QString& desc, IGisProject *project);
+    static void newWpt(const poi_t &poi, IGisProject *project, bool openEditWIndow = true);
     static bool getIconAndName(QString& icon, QString& name);
 
     static void drawCircle(QPainter& p, const QPointF& pos, const qreal& r, const bool &avoid, const bool &selected);
--- a/src/qmapshack/mouse/CMouseSelect.cpp
+++ b/src/qmapshack/mouse/CMouseSelect.cpp
@@ -26,6 +26,8 @@
 #include "mouse/CScrOptSelect.h"
 #include <QtWidgets>
 
+QMutex CMouseSelect::mutexPoisFound(QMutex::NonRecursive);
+
 CMouseSelect::CMouseSelect(CGisDraw *gis, CCanvas *canvas, CMouseAdapter *mouse)
     : IMouseSelect(gis, canvas, mouse)
 {
@@ -167,6 +169,10 @@ void CMouseSelect::draw(QPainter& p, CCanvas::redraw_e needsRedraw, const QRect 
         return;
     }
 
+    if(!mutexPoisFound.tryLock())
+    {
+        return;
+    }
     QList<IGisItem*> items;
     findItems(items);
 
@@ -175,7 +181,7 @@ void CMouseSelect::draw(QPainter& p, CCanvas::redraw_e needsRedraw, const QRect 
         item->drawHighlight(p);
     }
 
-    for(QPointF pos : posPoiHighlight)
+    for(QPointF pos : qAsConst(posPoiHighlight))
     {
         if(pos != NOPOINTF)
         {
@@ -187,6 +193,7 @@ void CMouseSelect::draw(QPainter& p, CCanvas::redraw_e needsRedraw, const QRect 
     }
 
     IMouseSelect::draw(p, needsRedraw, rect);
+    mutexPoisFound.unlock();
 }
 
 void CMouseSelect::slotUpdate()
@@ -201,6 +208,7 @@ void CMouseSelect::slotCopy() const
     IGisProject* project = CGisWorkspace::self().copyItemsByKey(itemKeys);
     if(project != nullptr)
     {
+        QMutexLocker lock(&mutexPoisFound);
         CGisWorkspace::self().addPoisAsWpt(poisFound, project);
     }
     canvas->resetMouse();
--- a/src/qmapshack/poi/CPoiPOI.cpp
+++ b/src/qmapshack/poi/CPoiPOI.cpp
@@ -544,4 +544,6 @@ void CPoiPOI::loadPOIsFromFile(quint64 categoryID, int minLonM10, int minLatM10)
                                   key, categoryNames[categoryID], garminIcon);
         mutex.unlock();
     }
+    // Close database, as this method is called from mutiple threads.
+    QSqlDatabase::removeDatabase(filename);
 }
--- a/src/qmapshack/poi/CPoiPOI.cpp
+++ b/src/qmapshack/poi/CPoiPOI.cpp
@@ -125,8 +125,13 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
     displayedPois.clear();
     QRectF freeSpaceRect (QPointF(), IPoi::iconSize() * 2);
     //Find POIs in view
-    for(quint64 categoryID : categoryActivated.keys(Qt::Checked))
+    const QList<quint64>& keys = categoryActivated.keys();
+    for(quint64 categoryID : keys)
     {
+        if(categoryActivated[categoryID] != Qt::Checked)
+        {
+            continue;
+        }
         for(int minLonM10 = qFloor(xMin * RAD_TO_DEG * 10); minLonM10 < qCeil(xMax * RAD_TO_DEG * 10); minLonM10++)
         {
             for(int minLatM10 = qFloor(yMin * RAD_TO_DEG * 10); minLatM10 < qCeil(yMax * RAD_TO_DEG * 10); minLatM10++)
@@ -144,7 +149,7 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
                     mutex.unlock();
                     return;
                 }
-                for(quint64 poiToDrawID : loadedPoisByArea[categoryID][minLonM10][minLatM10])
+                for(quint64 poiToDrawID : qAsConst(loadedPoisByArea)[categoryID][minLonM10][minLatM10])
                 {
                     const CRawPoi& poiToDraw = loadedPois[poiToDrawID];
                     QPointF pt = poiToDraw.getCoordinates();
@@ -179,7 +184,7 @@ void CPoiPOI::draw(IDrawContext::buffer_t& buf)
     }
 
     //Draw Icons
-    for(const poiGroup_t& poiGroup: displayedPois)
+    for(const poiGroup_t& poiGroup: qAsConst(displayedPois))
     {
         QFontMetricsF fm(CMainWindow::self().getMapFont());
 
@@ -248,7 +253,7 @@ bool CPoiPOI::findPoiCloseBy(const QPoint& px, QSet<poi_t>& poiItems, QList<QPoi
     poiGroup_t poiGroup;
     if(getPoiGroupCloseBy(px, poiGroup))
     {
-        for(quint64 key : poiGroup.pois)
+        for(quint64 key : qAsConst(poiGroup.pois))
         {
             poiItems.insert(loadedPois[key].toPoi());
         }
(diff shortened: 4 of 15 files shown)


pingurus:
@kiozen @wthaem

Any more links you'd wish to add? OSM also supplies dedicated facebook/vk/instagram/whatsapp links, however with the code we have now, we cover the largest seven link sources: [tag history](https://taghistory.raifer.tech/#***/contact:instagram/&***/contact:website/&***/website/&***/url/&***/contact:facebook/&***/contact:vk/&***/contact:whatsapp/&***/contact:email/&***/contact:phone/&***/phone/&***/email/)
For the other ones someone would have to do some research how we can link to them.

wthaem:
My guess is that the latest commit does already use proj8. If so, I can not yet compile this QMS commit/PR because I did not yet understand how to cope with libtiff required for the proj8 compilation.

pingurus:
You're not on linux right? I documented a bit in #337 what I tried in order to build, but I doubt that's useful for you.

wthaem:
@pingurus: Success! With kiozen's help I have now a way to compile and run QMS with proj version 8 under Windows 10 (there was one annoying QMS/proj issue: an unexpected format change in a projection string).

capture_001_16032021_201003

pingurus:
@kiozen: I have another idea for this, so don't merge for now

kiozen:
Imho this can be done better. This length list of OSM tags is just intimidating. Additionally the GPX waypoint definition has fields for some of the tags. 

Let me show. This is how a waypoint is created by the current code:
image

And this is how it might look like:

image

* The name must not be repeated
* All links a re in the link section and therefor available in the on-screen bubble
* The elevation is taken over in the elevation field.

Still the description field is kind of ugly. It would be nice to generate some natural text out of all tags but I guess this is quite complex. Are there other apps doing better?

kiozen:
Btw it would be niche to have all added waypoints locked by default.

src/qmapshack/poi/CRawPoi.cpp
```diff
@@ -74,8 +74,87 @@ poi_t CRawPoi::toPoi() const
     poi_t poi;
     poi.pos = coordinates;
     poi.name = getName();
-    poi.desc = data.join("<br>\n");
     poi.icon = garminIcon;
+
+    const QRegularExpression emailRegex ("(contact:email)=(.+)", QRegularExpression::UseUnicodePropertiesOption);
```
kiozen, line 79: Imho this should be ´static const` as you do not want to initialize the regex multiple times

src/qmapshack/poi/CRawPoi.cpp
```diff
@@ -74,8 +74,87 @@ poi_t CRawPoi::toPoi() const
...
+    const QRegularExpression phoneRegex ("(contact:phone|phone|contact:mobile)=(.+)", QRegularExpression::UseUnicodePropertiesOption);
+    const QRegularExpression websiteRegex ("(contact:website|website|url)=(.+)", QRegularExpression::UseUnicodePropertiesOption);
+    const QRegularExpression osmRegex ("(osm_id)=(P|W|R)/(.+)", QRegularExpression::UseUnicodePropertiesOption);
+    const QRegularExpression wikipediaRegex ("(.*:)?(wikipedia)(:[A-z]{2})?=(.+)", QRegularExpression::UseUnicodePropertiesOption);
+    const QRegularExpression wikidataRegex ("(.*:)?(wikidata)=(.+)", QRegularExpression::UseUnicodePropertiesOption);
+    poi.desc = "";
+    bool firstLine = true;
+    for(QString line : data)
```
kiozen, line 87: `const QString& line` and an € in the constant reference bucket ;)

src/qmapshack/poi/CRawPoi.cpp
```diff
@@ -74,8 +74,87 @@ poi_t CRawPoi::toPoi() const
...
+        }
+        else
+        {
+            poi.desc += "<br>\n";
+        }
+
+        QRegularExpressionMatch emailMatch = emailRegex.match(line);
+        if(emailMatch.hasMatch())
+        {
+            poi.desc += emailMatch.captured(1).trimmed() + "=<a href=\"mailto:" + emailMatch.captured(2).trimmed() + "\">" + emailMatch.captured(2).trimmed() + "</a>";
```
kiozen, line 101: if you take leading and trailing white space out of the regex capture you wouldn't need `trimmed()`.

src/qmapshack/poi/CRawPoi.cpp
```diff
@@ -74,8 +74,87 @@ poi_t CRawPoi::toPoi() const
...
+        if(firstLine)
+        {
+            firstLine = false;
+        }
+        else
+        {
+            poi.desc += "<br>\n";
+        }
+
+        QRegularExpressionMatch emailMatch = emailRegex.match(line);
```
kiozen, line 98: `const QRegularExpressionMatch&` no need to create copies.

pingurus:
What do you think of the UI now? 

Should we translate the links (i.e. website <=> Webseite)? And remove the leading `contact:` if it is there?

src/qmapshack/poi/CRawPoi.cpp
pingurus: :moneybag: :wink:

src/qmapshack/gis/wpt/CGisItemWpt.h
```diff
@@ -348,7 +348,7 @@ class CGisItemWpt : public IGisItem
     const searchValue_t getValueByKeyword(searchProperty_e keyword) override;
 
     static QString getLastName(const QString &name);
-    static void newWpt(const QPointF &pt, const QString& name, const QString& desc, IGisProject *project, const QString &icon = "", bool openEditWIndow = true);
+    static void newWpt(const QPointF &pt, const QString& name, const QString& desc, IGisProject *project, const QString &icon = "", const QList<link_t> links = QList<link_t>(), quint32 ele = NOINT, bool openEditWIndow = true);
```
kiozen, line 351: Instead of passing this endless list of parameters and default arguments wouldn't it be better to override newWpt and pass the POI object?

kiozen (changes requested):
Yes translation if possible is always desired. 

Would it be possible to derive a nice address form all the address tags? Maybe it's better to store the tags not as a string but as a map. By that the regex voodoo isn't needed. And the conversion map to string could be used for the tooltip, too.

pingurus:
> Yes translation if possible is always desired.
> 
> Would it be possible to derive a nice address form all the address tags? Maybe it's better to store the tags not as a string but as a map. By that the regex voodoo isn't needed. And the conversion map to string could be used for the tooltip, too.

Even with a map, the regex vodoo would still be needed (and more complicated), as for example for wikipedia I need the exact tag to form the URL. And to save resources, regexing should only happen upon conversion to a poi_t, so the map would only be created for `poi_t::toPoi`.

pingurus:
In what format do you want the address to be? It might be a rather cumbersome task to generate valid address strings for each format used around the world :smile:

See for example https://wiki.openstreetmap.org/wiki/Proposed_features/AddrN

pingurus:
@kiozen @wthaem Is the tooltip better now? A said above, I'm not sure how to handle addresses, but they should be transformed in some way, so both description and tooltip look better.

Regarding the code to generate the tooltip I think we can do some optimisation once we know how it should look.

wthaem:
3 remarks:

* Replace "From category" by "Category"
* Replace "Has links to" by "Links" and suppress this line if there is no link
* Very nice trap: cope with left/right justified - see attachment. Waypoint and POI info should have the same justification and writing order (look especially at 8.30 - 11.0!) . And: if there is a link in the POI, then it is sometimes shown in the corresponding waypoint as proper link. But not always - see attachment. Why?

left_right

kiozen:
> A said above, I'm not sure how to handle addresses, but they should be transformed in some way, so both description and tooltip look better.

Why not start simple with 
<street><house number>
<postcode><city>

If any of theses is missing, well it's missing. And all the other extensions are subject to further work if a good use-case can be presented.

wthaem:
* address: maybe, add country, too?

* at least in my region phone numbers can not yet used as URL. Hiding them behind a link in the waypoint edit window is therefore not the best idea. Can the phone number be shown without a link? Windows10: if you click the phone link, you are asked which app to use to open the link.

pingurus:
I've pushed an intermediate piece of code that is not really tested and for example breaks the assignment of the poi icons.

It would be nice if you'd still take a look at the tooltips and waypoints regarding the tags, so I can fix more bugs when I find time to code the next time :smile:

I've also switched to maps for the tags (yeah, I thought about that a bit and it makes life much easier for implementing filtering based on tags :wink:). The loading of the POIs may thus be a bit slower as more code runs in the constructor of the `CRawPoi` items, but I think it's not noticable.

pingurus:
Note to self/note so you don't need to mention it: phone numbers are suppressed right now.

src/qmapshack/poi/CRawPoi.cpp
```diff
@@ -19,63 +19,344 @@
...
+
+    return desc;
+}
+
+QList<IGisItem::link_t> CRawPoi::getLinks() const
+{
+    QList<IGisItem::link_t> links;
+    if(data.contains("contact:email"))
+    {
+        links += {"mailto:" + data["contact:email"], QObject::tr("Email"), "text/html"};
```
kiozen, line 235: Please no `QObject::tr()` Use `Q_DECLARE_TR_FUNCTIONS` macro to define a local `tr()`

pingurus: I knew there was some other way to do it :thinking:

kiozen: Yes and by that Linguist is listing the stuff as part of CRawPoi helping the translators to find the correct context.

pingurus:
@wthaem @kiozen 
Any more remarks regarding the tooltip?

Is there anything I can do about the left/right alignment/writing direction of the tooltips or is this a bug somewhere deep in qt/QMS?

kiozen:
Tool tips are looking fine for now. I guess we will tinker on the details in the future. But for now it's ok. However converting the POI to a waypoint the text is lost. Can you please have a look?

And you have to address the clang/clazy warnings. I added the topic to the coding guidelines lately. 

> Use Clang-Tidy and Clazy. All pull requests will be tested with these tools. They won't be merged on warnings unless it's definitely a false positive. QtCreator has support for these tools. Use Analyze->Clang-Tidy and Clazy to check all files you changes. Don't start a full scan as it takes hours.

kiozen:
I found another "issue". If there is no address there are quite some empty lines between the category and the tags. This should be reduced to a single empty line.

pingurus:
> However converting the POI to a waypoint the text is lost. Can you please have a look?

Can you provide an example? It works for me.

src/qmapshack/gis/wpt/CGisItemWpt.cpp
```diff
@@ -285,6 +286,29 @@ void CGisItemWpt::newWpt(const QPointF& pt, const QString &name, const QString &
...
+    cfg.setValue("Waypoint/lastIcon", wpt->getIconName());
+}
+
+void CGisItemWpt::newWpt(const poi_t &poi, IGisProject *project, bool openEditWIndow)
+{
+    SETTINGS;
+    const QString& _icon = poi.icon.isEmpty() ? cfg.value("Waypoint/lastIcon", "Waypoint").toString() : poi.icon;
+    const QString& _name = poi.name.isEmpty() ? getLastName("") : poi.name;
+
+    CGisItemWpt * wpt = new CGisItemWpt(poi.pos * RAD_TO_DEG, _name, _icon, project);
```
kiozen, line 299: I would suggest to set the new waypoint initialy to read-only. By that the information in the waypoint's edit dialog is more compact. And most likely no one will edit the waypoint anyways.

`wpt->setReadOnlyMode(true);`

kiozen (changes requested):

kiozen:
> Can you provide an example? It works for me.

All good. Looks like I messed up.

pingurus:
> Tool tips are looking fine for now. I guess we will tinker on the details in the future. But for now it's ok. ~ However converting the POI to a waypoint the text is lost. Can you please have a look?~ Forget about that. It's working. I probably started the wrong binary.
> 
> And you have to address the clang/clazy warnings. I added the topic to the coding guidelines lately.
> 
> > Use Clang-Tidy and Clazy. All pull requests will be tested with these tools. They won't be merged on warnings unless it's definitely a false positive. QtCreator has support for these tools. Use Analyze->Clang-Tidy and Clazy to check all files you changes. Don't start a full scan as it takes hours.

When I try to start Clang-Tidy and Clazy from qtcreator I get the following error: `10:11:04: Clazy: Invalid executable "". Stopped.`. Is there a quick fix for that that I'm not aware of?

kiozen:
> > Tool tips are looking fine for now. I guess we will tinker on the details in the future. But for now it's ok. ~ However converting the POI to a waypoint the text is lost. Can you please have a look?~ Forget about that. It's working. I probably started the wrong binary.
> > And you have to address the clang/clazy warnings. I added the topic to the coding guidelines lately.
> > > Use Clang-Tidy and Clazy. All pull requests will be tested with these tools. They won't be merged on warnings unless it's definitely a false positive. QtCreator has support for these tools. Use Analyze->Clang-Tidy and Clazy to check all files you changes. Don't start a full scan as it takes hours.
> 
> When I try to start Clang-Tidy and Clazy from qtcreator I get the following error: `10:11:04: Clazy: Invalid executable "". Stopped.`. Is there a quick fix for that that I'm not aware of?

Usually this should work out of the box. Maybe your designer is outdated. I always install mine with the installation tool provided by Qt.

pingurus:
Can I just more or less blindly accept the proposed fixits?

pingurus:
Good Idea to include the tool, so you don't have to remind me of all the `const QString&` :laughing:

kiozen:
> Can I just more or less blindly accept the proposed fixits?

No not really. In most cases they are ok, however there are many cases where the problem would have been solved if the provided container wouldn't be an instance but a const reference. e.g.

```
QStringList keys  = myMap.keys()

...

for(const QString& key : keys)
{
}
```

The automatic fix would simply wrap `keys` by `qAsConst(keys)`. However the proper fix would be to write 
```
const QStringList& keys  = myMap.keys()
```

avoiding the stupid copy of the list.

Therefore check all automatic fixes first and select only those you really want to apply. And fix the rest properly

> Good Idea to include the tool, so you don't have to remind me of all the const QString&

This was exactly the motivation to clean up the code to make the tool work in a sensible way.

pingurus:
@kiozen So you prefer if we don't use the `qAsConst` even if the 'proper fix' would also just be to introduce a new throwaway variable that is `const`? Or is the code in this PR fine with using the `qAsConst`?

kiozen:
As far as I understood the problem you have to use qAsConst for lvalues and a local const ref for rvalues. So using qAsConst is fine.

pingurus:
I think we can merge this and I'll start a new PR for the filter.
Continuing on #352 only really makes sense if we have a stable interface, which is likely to change whilst implementing the filters.

kiozen:
I still miss a push fixing all the clazy issues.

pingurus:
Isn't  a779457 fixing all of them?

kiozen:
@pingurus  Nope:

image

pingurus:
Then I've missed out those files :thinking:

pingurus:
How should I handle the last warning? (The one in CPoiPropSetup.cpp)

pingurus:
I don't get why we get this warning here, as we don't call the virtual function in the base class but in the derived class and according to the [ISO CPP FAQ](https://isocpp.org/wiki/faq/strange-inheritance#calling-virtuals-from-ctors ) the problem arises when you call a virtual function in the ctor of the base class. Or did I get something wrong?

kiozen:
Well, but you could derive from `CPoiPropSetup` and override `slotPropertiesChanged()`. In that case it's not working properly and therefore it's better to call `CPoiPropSetup::slotPropertiesChanged();` explicitly.

kiozen:
I will give it another functional check and code review, later. And if nothing else pops up I would say we merge this into dev. Agreed or do you want to fix something else before hitting dev?

src/qmapshack/gis/wpt/CGisItemWpt.cpp
```diff
@@ -285,6 +286,30 @@ void CGisItemWpt::newWpt(const QPointF& pt, const QString &name, const QString &
         wpt->setDescription(desc);
     }
```
kiozen, line 288: Why did you remove 

```
wpt->editInitial();
```

This is changing the expected behavior for adding waypoints.

src/qmapshack/poi/CPoiPOI.cpp
```diff
@@ -318,16 +326,35 @@ bool CPoiPOI::getToolTip(const QPoint &px, QString &str) const
             {
                 str += "<b>" + name + "</b><br>\n";
             }
-            str += tr("From category ") + "<b>" + poiFound.getCategory() + "</b><br>\n";
-            str += poiFound.getData().join("<br>\n");
+            str += tr("Category: ") + "<b>" + poiFound.getCategory() + "</b><br>\n";
+            str += poiFound.getDesc();
+            str += "<br>\n";
+            const QList<IGisItem::link_t> links = poiFound.getLinks();
```
kiozen, line 332: no reference?

kiozen (changes requested):

src/qmapshack/gis/wpt/CGisItemWpt.cpp
pingurus: I did not remove it, I moved into a `if` which defaults to true

kiozen: It's missing for 
```
void CGisItemWpt::newWpt(const QPointF& pt, const QString &name, const QString &desc, IGisProject * project)
```

Just compare QMS from `dev` and your branch. The one on dev opens the edit dialog on a new waypoint. Your branch simply sets the waypoint.

pingurus: Ok, my mistake. The way the side by side diff pulled it apart I didn't notice that it was only present in one of the methods :laughing:

kiozen: no worries, all good :D

kiozen:
I hate to spoil the party but with latest commit I can not load pois anymore.

* The poi file activates
* If I select the first category in the list it loads and displays.
* If I select other categories they are not loaded and displayed.

pingurus:
The suggestion of clazy broke it. Why? No idea :laughing: 
It seems that iterating over `foo bar: baz`is not necessarily the same as `foo bar: baz.keys()`even though clazy says so and I thought so aswell.

src/qmapshack/poi/CRawPoi.cpp
```diff
@@ -19,63 +19,363 @@
...
+        }
+        desc += "<b>" + tr("Fax: ") + "</b>" + data["fax"];
+    }
+    if(!desc.isEmpty())
+    {
+        desc += "<br>\n" "<br>\n";
+    }
+    // Append all other tags
+    bool firstTag = true;
+    for(const QString& key : data)
```
kiozen, line 216: I think there is another bug. It should read 
```
    const QStringList& keys = data.keys();
    for(const QString& key : keys)
```
 Could you please check all other locations where you used range loops, too. And actually test you code before you commit ;)

kiozen (changes requested):

src/qmapshack/poi/CRawPoi.cpp
pingurus: Sorry. I tested it better now and found a segmentation fault when selecting POIs from the map but can't figure out it's cause.
It happens right after you selected whether you want to choose a symbol manually or not when you prompt it to copy the selected items over to a project, regardless of your response. It however either faults in the `SETTINGS` macro if you select `No` and in `QString::isEmpty()` if you select `yes`. The mutex locker is there, so that's not the cause and the pois that should be copied look intact. Do you have an idea what's the cause?

kiozen: start with gdb and analyze the backtrace?

kiozen: Btw we still have issus #357, too. Seems to be only on the POI feature branch. Probably some mutex problem.

kiozen: Just to keep track: The segfault you mentioned is not fixed, yet. Issue #357 is not reproducible. Correct?

kiozen: Bump! Any news on that?

pingurus: Didn't have the nerves yet to to investigate the reason, as I currently code much for work.
But I think I will spend some time on it this weekend, as it was a short working week :wink:

pingurus: > Just to keep track: The segfault you mentioned is not fixed, yet. Issue #357 is not reproducible. Correct?

Correct. 

Could you please check if you get the segfault when you try to select items on the map and copy them to a project? I think I have it traced down to the `exec()`calls of the dialogs that spin an additional event loop after which it faults when accessing the `poi_t`that it currently copies. Since it is only after the dialogs that it fails, you have to select an item that has the default POI symbol. Since it is something that definitely worked before, I'd like to check if it might have to do with my flawed libgdal setup (although I doubt it).

kiozen: I couldn't reproduce the crash. However I noticed another bug. You should move the line 

```
wpt->setReadOnlyMode(true);
```

behind `wpt->editInitial();`  else the icon is not editable (unless you unlock the item first)

I just wonder if the crash also occurs when you run QMS with gdb. Or does it produce a core dump that can be analyzed. Are there any Qt related messages when running with `-d`?

pingurus: > I couldn't reproduce the crash. However I noticed another bug. You should move the line
> 
> ```
> wpt->setReadOnlyMode(true);
> ```
> 
> behind `wpt->editInitial();` else the icon is not editable (unless you unlock the item first)

I'll do that :+1: 

> Are there any Qt related messages when running with -d?

There were, but that was related to another edge-case bug that occurred by chance which I managed to fix :laughing: 
I used `qDebug()` to trace down the issue to the `exec()` as `qDebug() << "icon: " << poi.icon;` works before doing `exec()` on the dialog window but not after (at least for me).

src/qmapshack/mouse/CMouseSelect.cpp
```diff
@@ -26,6 +26,8 @@
 #include "mouse/CScrOptSelect.h"
 #include <QtWidgets>
 
+QMutex CMouseSelect::mutexPoisFound(QMutex::Recursive);
```
pingurus, line 29: @kiozen could you please have a look why this mutex isn't effective in stopping `draw()` being run parallel to `slotCopy()`? I'm pretty sure my segfaults are due to `poisFound` being updated while the copying is happening and can't figure out why this mutex doesn't lock.

kiozen: Because  `CMouseSelect::draw()` and `CMouseSelect::slotCopy()` are running in the same thread. See `void CCanvas::paintEvent(QPaintEvent*)` line 640. This is the main draw loop that resides in the main thread. The additional thread is triggered by `void IDrawContext::draw(QPainter& p, CCanvas::redraw_e needsRedraw, const QPointF& f)`.

kiozen: Maybe you get away with it by setting the `mutexPoisFound` to non-recursive and quit `findItems` when the mutex is locked. but I am not sure if that works in the same thread. Or you create a temporary real copy of the POI first. By that you can enter the eventloop and `findItems` can recreate the list without destroying your POI to copy.

pingurus:
Your suggestion for the mutex worked, so I think this is done now. I still can't reproduce #357, so I can't fix that.

Issue #342, opened 2021-03-14
Read GPX files as POIs


pingurus:
**Important Notice:**

**Do not delete any of the sections. Read what is requested. _Replace(!)_ the instructions by your _answer(!)_. Feature requests reports that are not properly filed might be closed without further notice.**

**Choose a good title**

What is a good title?
The title will be listed in the changelog when the issue is fixed. Do you think someone else would understand what's fixed or done from your title? Does it really name the real topic of the issue?

**Is your feature request related to a problem? Please describe.**

I would love to load thousands of geocaches without having to wait unnecessarily long. It makes no sense to load them into the workspace, as an enumeration of all the geocaches only will clutter the workspace.

Other use-cases are possible where one has much data in GPX format and only needs it read-only with the possibility to copy it to the workspace.

**Describe the solution you'd like**

Handle them just as the *.poi files.

**Describe alternatives you've considered**

Do nothing :wink:

**Give a reason why someone else should spend time on that request**

1. I'll implement it myself
2. There is a more or less frequent complaint that loading enormous amounts of data takes long. There would be a response to this complaint other than "Stop using QMS in a way that it was not intended for"

**Additional context**

As I think #294 is almost done, I think we can tackle another POI data type without causing too much headache.

kiozen:
Hm, probably that's why the ticket is still open :thinking: :wink:

flock-vi:
Just wanted to mention the difference to the screenshot in WIKI:
![POI_WIKI](https://user-images.githubusercontent.com/84787872/119672849-3535d380-be3b-11eb-9bd0-aba781aaf761.jpg)

kiozen:
@wthaem think you have to update the Wiki. @pingurus do you plan to restart the PR in the near future?

wthaem:
I wasn't aware that this nice feature didn't reach the dev/release version. I'll remove its description as soon as its future is clarified.

pingurus:
I'd love to restart it, but I'm not sure when I'll find the time. I guess that's only a maybe for the near future, but the next winter will  come for sure :wink:

pingurus:
I might be restarting work on this soon. My thoughts so far:

# Approach 1: Convert gpx file to SQLite database:

I would code a page similar to the VRT builder to enable the user to convert gpx files to a SQLite database that then can be used for POIs. I would simply dump the `<wpt>` objects into the sqlite file as string so it can be fully reconstructed when copied to the workspace.

- Upsides:
[...]

PR #352, opened 2021-03-21
[QMS-342] Read GPX files as POIs

pingurus:
_**Note: Do not delete any of the sections Answer them all. Replace the descriptive text by your answer**_

**What is the linked issue for this pull request (start with a `#`):**

QMS-#342

* Added possibility to read GPX files as POIs

1. Copy a gpx file to your POI directory
2. Activate teh GPX file
3. See that waypoints are displayed

- [ ] yes

- [ ] yes


src/qmapshack/poi/CPoiGPX.cpp
```diff
@@ -0,0 +1,379 @@
...
+    if(xmlGpx.tagName() != "gpx")
+    {
+        throw tr("Not a GPX file: %1").arg(filename);
+    }
+
+    const QDomNodeList& xmlWpts = xmlGpx.elementsByTagName("wpt");
+    for(int n = 0; n < xmlWpts.count(); ++n)
+    {
+        const QDomNode& xmlWpt = xmlWpts.item(n);
+        QSharedPointer<CGisItemWpt> wpt = QSharedPointer<CGisItemWpt>(new CGisItemWpt(xmlWpt, nullptr));
```
pingurus, line 67: @kiozen is this too hacky? I also added a shared pointer to poi_t, so if that is not a `nullptr`, we can simply copy the `CGisItemWpt` to any project the user whishes.

kiozen: Not sure why we need a shared pointer here. Doesn't  the object load all POIs at the beginning? And wouldn't referencing the POIs in that list be sufficient? After all this list does not change when loaded, right?

And yes I know, `CGisItemWpt` is violating the model/view  architecture and this backlashing, now. But creating literally thousands of parentless QTreeWidgetItems with the additional overhead of CGisItemWpt doesn't feel right.

On the other hand side there is `IGisItem::wpt_t` and `IGisItem::readWpt()` probably could be a static method (haven't tried it yet). Maybe that is a much better approach.

kiozen:
This are two alternate GPX/POI collections frequently used:

[alternative_pois.tar.gz](https://github.com/Maproom/qmapshack/files/6177198/alternative_pois.tar.gz)

Especially Archie's Camping List does not turn out very well.

kiozen:
I like how this is handling these huge GPX collections. Much better than pumping these collections into the workspace.

src/qmapshack/poi/CPoiGPX.cpp
pingurus: I took a shared pointer, as I was not sure about the possibly dangling pointers when the user deactivates the gpx-poi file whilst having a "Select Items on Map" open. It might not end well when the file is closed and the wpts are destroyed and then the user tries to copy them over to a project.

Regarding `IGisItem::wpt_t`: it does not include the geocache information, so I'd prefer to avoid using that. :smile:

kiozen: Pointer: hm, ok, that might be a problem. However I would assume all objects on the map are destroyed when deactivating the file. And the "Select Items on Map"  should be updated in that case. At least for stuff on the workspace it does (partially, the summary is not updated :( ).

And there is `CGisItemWpt::geocache_t` that can be used outside of CGisItemWpt.

pingurus:
> Especially Archie's Camping List does not turn out very well.

That's the bad format of the gpx file, I don't think it would be reasonable to code something just to make that file more usable

wthaem:
@pingurus: Load a GPX file with waypoints into the workspace and load it as POI file. Look at a waypoint with a blue flag as icon. Then you get the already discussed problem of 2 blue flags as shown in the attachment. It is in principal 1 point with 2 (different?) icons!

blueflags

wthaem:
When loading GPX files as POI files it happens that the category name is empty. This should be avoided. Insert in this case e.g. `Default category` (happens if all POIs have only a `type` tag (could this tag name used as category name?)) .

NoCategory

Issue #378, opened 2021-05-08
Add option to link map views


kiozen:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

Sometimes it's quite convenient if all map views would have a common point of focus and zoom level. 

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

* There should be an option in the view menu to toggle linkage.
* If active all maps follow the same point of focus and zoom.
* If the map in one view is moved it is moved in all other views, too
* If the map is zoomed in one view it is zoomed in all other views, too

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

So far the user can adjust all views by double click on items in the workspace list for each view. This is just laborious.

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

This is probably a mode expected by many users working with multiple views and makes kind of sense.

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)
n/a

PR #379, opened 2021-05-08
[QMS-378] Add option to link map views

kiozen:

QMS-#378

### What you have done:

* Add toggle action to view menu
* Add signals to canvas
* Make CMainWindow handle the signals iterating
over the list of tab widgets and changing all views.

1. Enable option in view menu
2. Open two or more map views
3. Move map -> all maps must show same viewport
4. Zoom map -> all maps must show same viewport
5. Change projection of one view -> Warning dialog and map view link is switched off

- [x] yes

- [x] yes


kiozen:
@wthaem @pingurus @mitxel-m Does anyone want to test, review?

wthaem:
> Steps to perform a simple smoke test

Tested on Windows 10 following the described smoke test steps with 3 views and a Garmin map, an online one and a VRT one: Switching sync on and off worked as described above. Change of CRS worked as described. 

Nice feature. Up-to-now I used an overlay of several maps in 1 view and played with the opacity to achieve a similar effect. The new feature is easier to handle.

mitxel-m:
I'm afraid I won't be able to test it in the next few days, because I'm away. 

This new feature will be very welcome. A while ago it was mentioned in mendiak and I was thinking about the possibilities.

 
 Although I can't test it by myself let me suggest to also test with views having different zoom type (logarithmic / square) . If the value that the views follow were the zoom level, (ex Z=12) the extension is different. it's just a thought, (Possibly, kiozen has already taken care of this). 
   
   
The same happens with views that have different CRS, especially, if one is lat/lon and others UTM, another may even be rotated. After reading Step 5 to test I assume that the goal is to keep an equal area in all views, and for that you need to have exactly the same configuration for all of them, and if it is not the case then the warning appears and the view link is disabled.
  
Well, this is a good approach if you only expect an exact match for all viewports, but we lose a nice feature for other cases.  What is more important to keep exactly the same area in each view or to keep the best resolution in each view?

If you want to get the best resolution/quality for raster maps, it is recommended to set the view to match the original CRS of the raster map.  Then I would not expect the topleft/bottomroght corners of the views to coincide in all cases. IMHO., if  the views  follow the same central point, and a "more or less" similar scale and rotation, it would be great.
 
A proposal in case that there are views with different CRS:
 rather than switch off the view link,  it would be nice to display  the warning, informing that since there are views with different CRS the areas displayed in each view will not match exactly, and then let the user to choose whether to continue or to switch off the viewlink.

Additionally it could display a checkbox with the text "Do not report again in this session".
 
  
 it's just an idea since I can not test on these days ;-)

kiozen:
As you suspected I already tried to have different scale and projection and the outcome was not satisfying. I took the rectangle of the one viewport and tried to zoom to that rectangle in the other views. Especially for the scale this is not working well. Zoom to a track on the one view produced sub-optimal scale in the other ones. The track was either not 100% visible or quite small. Depending on the area and constellation of zoom levels. If you adjust the view the other ones are bad again. Very unsatisfying.

Projection is a pit better. However scaling is different for each projection. So either your track is smaller on the other views or it is too large for the viewport. 

To solve this we would need to introduce dynamic scaling for the other viewports. However this reaches its limits very fast again. You start zooming on one view switch to the other. With dynamic scaling both would be roughly equal. Now you zoom in at the other. For that operation you need to find some matching zoom level from the table. The result might be a quite rough zoom step in or out, or hardly any if you have been close to one. This is full of corner cases that can result into peculiar behavior. Nothing I would like to start coding.

In the end it boiled down to same scale, same CRS.  This results in the most natural and expected behavior. As QMapShack can reproject maps - sure with a bit of decreasing quality - it's not a problem. As user you have two options: Either live with the imperfections or reproject all you maps to the same CRS in high quality before you use them in QMapShack. At a certain point you have to die one or the other death. 

I attached two images to illustrate the difference in CRS. The one is OSM with Mercator and the other Google with UTM33. Double clicking the track results in both views into the same viewport, but scale is significantly different. This calls for a bug report by user not knowing what they are doing

osm_mercator

google_utm33

mitxel-m:
@kiozen Thank you for the detailed reply. I expected that you had already taken all this into account.

I agree that the behavior expected by many users will be just what you have now implemented, and anything else that differs from that could generate reports of "supposed" bugs from users who have no need to know about projections.

Some time ago I was exploring on this topic.  I didn't code anything, I was just looking at the necessary steps, and the problems that arose in order to make a proposal for enhancement.

I figured it was difficult to get consistent behavior linking views automatically, (exactly the same as you have described), and explored another way. One idea was to have a new entry in the context menu: "copy view position", then change view and do "paste view position". (5 clicks)

Another approach was "Follow view position", then a combo to choice the desired view to follow.  

For both cases there would still be the issue of the zoom level. Althought this could be  secondary because it can be done quickly with the mouse wheel, I left it "frozen" at that point, pending a better study of whether some kind of conversion between the different types of zoom would be possible.

The approach of following "the centre of the view" and not the "limits of the view" was precisely intended to avoid complaints when using different projections.

All this said,  I think what kiozen has already implemented by linking the views automatically is a great improvement. ;-)

Issue #396, opened 2021-05-24
Save route sub-points as ordinary route points in case of track routing


pinysuse:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )
Due to the behavior of my navigation devices I use airline-navigation based on _routes_.
This requires significant upfront planning and defining a sufficient number of route-/way-points.
The effective track evolves by following the natural terrain while the navigation device points to the next way-point.

Often, planning is based on pieces or combinations of existing tracks.
While QMapShack provides powerful tools to manipulate tracks, the current features do not allow for the simple conversion of a track into a route.

**_Track routing_** seems rather close to the intended solution! However, it does not allow to save the route including all the sub-point nicely created by track based routing.

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

Let's make use of the existing feature [Description of routing methods in QMapShack - Method 4: Routing with tracks (track 
routing)](https://github.com/Maproom/qmapshack/wiki/DocGisItemsRouting#description-of-routing-methods-in-qmapshack).

#### Request: 

Save route sub-points as ordinary route points in case of track routing

- The below fragment of an addition to the manual describes the possible user workflow to convert tracks to routes with this feature.
- Implementation option: Instead of saving route sub-points in track routing mode an additional button `Save as new (T)` could be added.

#### Note: 

The section [Save data](https://github.com/Maproom/qmapshack/wiki/DocGisItemsEditMultiple#save-data) could be understood in such way that sub-points created by this method would already be saved in the database. This is not the case. 
Nevertheless, they would not be exported. So a possibility to convert sub-point to ordinary route points is requested.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

Current work around:

1. Export the track
2. Convert to a route in an external tool or editor
3. Re-import the route

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

- It appears this issue is addressed sporadically by users.
- Converting routes to tracks is possible in QMS but not vice versa

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

[...]

PR #410, opened 2021-06-04
[QMS 396] Save route sub-points as ordinary route points in case of track routing

UEckhardt:

QMS-#396

### What you have done:

Implement the basic track to route conversion via a new icon in the track context menu and also to the track range menu so that a route can be created from a track range.

Add a new track to route dialog to configure if the track to route conversion will include the sub-points of the track as route points
and allows to rename the route and select/create a project.

1. Create a small track with 5 points and perform an auto-routing for this track,
2. Open the setup workspace dialog and ensure that the new option "Include sub-points.." is unchecked.
3. Select the track and click on the new route icon in the track context menu.
4. A new route without sub-points is now created.
5. Select the track, click on the range of points button and select a subset of the track.
6. Click on the new route icon in the select range menu.
7. A new route without sub-points is created only for the subset of this track .
8. Open the setup workspace dialog and select the new option "Include sub-points..".
9. Select the track and click on the new route icon in the track context menu.
10. A new route including sub-points is now created.
11. Select the track, click on the range of points button and select a subset of the track.
12. Click on the new route icon in the select range menu.
13. A new route including sub-points is created only for the subset of this track .

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/gis/CGisListWks.cpp
+++ b/src/qmapshack/gis/CGisListWks.cpp
@@ -151,6 +151,7 @@ CGisListWks::CGisListWks(QWidget* parent)
     actionCopyTrkWithWpt = addAction(QIcon("://icons/32x32/CopyTrkWithWpt.png"), tr("Copy Track with Waypoints"), this, &CGisListWks::slotCopyTrkWithWpt);
     actionNogoTrk = addAction(QIcon("://icons/32x32/NoGo.png"), tr("Toggle Nogo-Line"), this, &CGisListWks::slotNogoItem);
     actionNogoTrk->setCheckable(true);
+    actionToRoute = addAction(QIcon("://icons/32x32/Route.png"), tr("Convert to Route"), this, &CGisListWks::slotToRoute);
 
     // waypoint related actions
     actionBubbleWpt = addAction(QIcon("://icons/32x32/Bubble.png"), tr("Show Bubble"), this, &CGisListWks::slotBubbleWpt);
@@ -1059,6 +1060,7 @@ void CGisListWks::showMenuItemTrk(const QPoint& p, const IGisItem::key_t& key)
     menu.addAction(actionEleWptTrk);
     menu.addAction(actionCopyTrkWithWpt);
     menu.addAction(actionNogoTrk);
+    menu.addAction(actionToRoute);
     menu.addSeparator();
     menu.addAction(actionDelete);
     menu.exec(p);
@@ -2466,3 +2468,13 @@ void CGisListWks::slotEleWptTrk()
     CGisWorkspace::self().addEleToWptTrkByKey(selectedItems2Keys<IGisItem>());
 }
 
+void CGisListWks::slotToRoute()
+{
+    CGisListWksEditLock lock(false, IGisItem::mutexItems);
+
+    CGisItemTrk *gisItem = dynamic_cast<CGisItemTrk*>(currentItem());
+    if(gisItem != nullptr)
+    {
+        CGisWorkspace::self().convertTrackToRoute(gisItem->getKey());
+    }
+}
--- a/src/qmapshack/gis/CGisListWks.cpp
+++ b/src/qmapshack/gis/CGisListWks.cpp
@@ -149,9 +149,9 @@ CGisListWks::CGisListWks(QWidget* parent)
     actionCombineTrk = addAction(QIcon("://icons/32x32/Combine.png"), tr("Combine Tracks"), this, &CGisListWks::slotCombineTrk);
     actionEleWptTrk = addAction(QIcon("://icons/32x32/SetEle.png"), tr("Replace Elevation by DEM"), this, &CGisListWks::slotEleWptTrk);
     actionCopyTrkWithWpt = addAction(QIcon("://icons/32x32/CopyTrkWithWpt.png"), tr("Copy Track with Waypoints"), this, &CGisListWks::slotCopyTrkWithWpt);
+    actionToRoute = addAction(QIcon("://icons/32x32/Route.png"), tr("Convert to Route"), this, &CGisListWks::slotToRoute);
     actionNogoTrk = addAction(QIcon("://icons/32x32/NoGo.png"), tr("Toggle Nogo-Line"), this, &CGisListWks::slotNogoItem);
     actionNogoTrk->setCheckable(true);
-    actionToRoute = addAction(QIcon("://icons/32x32/Route.png"), tr("Convert to Route"), this, &CGisListWks::slotToRoute);
 
     // waypoint related actions
     actionBubbleWpt = addAction(QIcon("://icons/32x32/Bubble.png"), tr("Show Bubble"), this, &CGisListWks::slotBubbleWpt);
@@ -1059,8 +1059,8 @@ void CGisListWks::showMenuItemTrk(const QPoint& p, const IGisItem::key_t& key)
     menu.addMenu(IGisItem::getColorMenu(tr("Set Track Color"), this, SLOT(slotColorTrk()), &menu));
     menu.addAction(actionEleWptTrk);
     menu.addAction(actionCopyTrkWithWpt);
-    menu.addAction(actionNogoTrk);
     menu.addAction(actionToRoute);
+    menu.addAction(actionNogoTrk);
     menu.addSeparator();
     menu.addAction(actionDelete);
     menu.exec(p);
--- a/src/qmapshack/gis/CGisWorkspace.h
+++ b/src/qmapshack/gis/CGisWorkspace.h
@@ -379,6 +379,7 @@ public:
 
     void convertRouteToTrack(const IGisItem::key_t& key);
 
+    void convertTrackToRoute(const IGisItem::key_t& key);
 
     void cutTrkByKey(const IGisItem::key_t& key);
 
--- a/src/qmapshack/gis/CGisWorkspace.h
+++ b/src/qmapshack/gis/CGisWorkspace.h
@@ -309,6 +309,16 @@ public:
     void getItemsByKeys(const QList<IGisItem::key_t>& keys, QList<IGisItem*>& items);
 
     void getNogoAreas(QList<IGisItem*>& nogos);
+
+    /**
+           @brief Get access to the workspace list
+
+           @return the workspace list
+    */
+    CGisListWks *getWks(void)
+    {
+        return treeWks;
+    }
     /**
        @brief Delete all items with matching key from workspace
 
--- a/src/qmapshack/gis/CGisWorkspace.h
+++ b/src/qmapshack/gis/CGisWorkspace.h
@@ -311,15 +311,6 @@ public:
     void getNogoAreas(QList<IGisItem*>& nogos);
 
     /**
-           @brief Get access to the workspace list
-
-           @return the workspace list
-    */
-    CGisListWks *getWks(void)
-    {
-        return treeWks;
-    }
-    /**
        @brief Delete all items with matching key from workspace
 
        @param key       the item's key as it is returned from IGisItem::getKey()
--- a/src/qmapshack/gis/trk/CGisItemTrk.cpp
+++ b/src/qmapshack/gis/trk/CGisItemTrk.cpp
@@ -3415,7 +3415,14 @@ void CGisItemTrk::toRoute()
 {
     qint32 idx1, idx2;
     QString routeName;
+    bool saveSubPts;
+    IGisProject *project = getParentProject();
 
+    if (nullptr==project)
+    {
+        qCritical("CGisItemTrk::toRoute no project");
+        return;
+    }
     getMouseRange(idx1, idx2, true);
 
     if(NOIDX == idx1)
@@ -3427,15 +3434,14 @@ void CGisItemTrk::toRoute()
         routeName = getName() + QString(" (%1-%2)").arg(idx1).arg(idx2);
     }
 
-    CTrkToRteDialog dlg(getParentProject(), routeName);
+    CTrkToRteDialog dlg(&project, routeName, saveSubPts);
     if(dlg.exec() == QDialog::Rejected)
     {
         return;
     }
-    IGisProject* project = dlg.getProject();
 
     SGisLine points;
-    getPolylineRangeFromData(points, idx1, idx2, dlg.getSaveSubPoints());
+    getPolylineRangeFromData(points, idx1, idx2, saveSubPts);
 
     new CGisItemRte(points, routeName, project, NOIDX);
 }
--- a/src/qmapshack/gis/trk/CGisItemTrk.cpp
+++ b/src/qmapshack/gis/trk/CGisItemTrk.cpp
@@ -3434,7 +3434,7 @@ void CGisItemTrk::toRoute()
         routeName = getName() + QString(" (%1-%2)").arg(idx1).arg(idx2);
     }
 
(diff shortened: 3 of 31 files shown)


src/qmapshack/gis/CGisWorkspace.h
```diff
@@ -309,6 +309,16 @@ class CGisWorkspace : public QWidget, private Ui::IGisWorkspace
...
+
+    /**
+           @brief Get access to the workspace list
+
+           @return the workspace list
+    */
+    CGisListWks *getWks(void)
+    {
+        return treeWks;
+    }
```
kiozen, line 321: There is no reason to make a private class member public for sure.

src/qmapshack/gis/trk/CToRouteNameProjectDialog.cpp
```diff
@@ -0,0 +1,295 @@
...
+    setupUi(this);
+    SETTINGS;
+    cfg.beginGroup("RouteToTrack");
+    saveSubPoints = cfg.value("saveSubPoints", false).toBool();
+    cfg.endGroup();
+
+    checkBoxSubPoints->setChecked(saveSubPoints);
+    key = IGisProject::getUserFocus();
+    QListWidgetItem* lastSelectedItem = nullptr;
+    const QTreeWidget *parent = CGisWorkspace::self().getWks();
```
kiozen, line 49: Ok, I see what's happening here. You are duplicating code from CGisWorkspace::selectProject().  Please use as much of the existing objects and dialogs as possible. Duplicating code drives every project into a mess that is hard to maintain.

src/qmapshack/gis/trk/IToRouteNameProjectDialog.ui
```diff
@@ -0,0 +1,218 @@
...
+         <horstretch>0</horstretch>
+         <verstretch>0</verstretch>
+        </sizepolicy>
+       </property>
+       <property name="lineWidth">
+        <number>2</number>
+       </property>
+       <property name="text">
+        <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Include sub-points of tracks in converted route.&lt;br/&gt;&lt;span style=&quot; font-size:12pt;&quot;&gt;Enabling this feature leads to routes with many points. Not recommendet for devices which have only limited space for saving route points.&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+       </property>
```
kiozen, line 72: No HTML in simple text fields. Translators will go nuts. If you enter text in designer you can switch between the HTML text editor and the plain text editor. Always use the plain text editor. If you really have to use html do it like CMouseRangeTrk. In the constructor you will see how to separate  text from HTML.

src/qmapshack/gis/trk/CToRouteNameProjectDialog.h
```diff
@@ -0,0 +1,86 @@
...
+#include "gis/CGisWorkspace.h"
+#include "ui_IToRouteNameProjectDialog.h"
+#include <QDialog>
+
+class QTreeWidget;
+
+/*
+ * Dialog Class for the track to route conversion.
+ */
+class CToRouteNameProjectDialog : public QDialog, private Ui::IToRouteNameProjectDialog
```
kiozen, line 32: Wouldn't something like `CTrkToRteDialog`  be more expressive on the purpose of that dialog?

kiozen (changes requested):
The dialog has to be changed.  Sorry. Either you stick to the established user guidance to create new elements or - as an alternative - you select the origin project as default project and add a tool button to select/create another project via the existing dialogs. If you choose the alternative we probably have to change the route to track dialog in another PR to have a somewhat consistent experience.

src/qmapshack/gis/CGisListWks.cpp
```diff
@@ -1059,6 +1060,7 @@ void CGisListWks::showMenuItemTrk(const QPoint& p, const IGisItem::key_t& key)
     menu.addAction(actionEleWptTrk);
     menu.addAction(actionCopyTrkWithWpt);
     menu.addAction(actionNogoTrk);
+    menu.addAction(actionToRoute);
```
kiozen, line 1063: Keep the order of the actions in the menu the same as the tool buttons in the on screen boxes. Users will appreciate it :wink:

src/qmapshack/gis/trk/IToRouteNameProjectDialog.ui
UEckhardt: Changed the dialog layout and used an own label with a smaller font.

src/qmapshack/gis/trk/CToRouteNameProjectDialog.cpp
UEckhardt: This would require a more comprehensive redesign of CGisWorkspace::selectProject(). I am not sure how to implement this. But at least the function getNameAndProject would benefit, since the name and project selection could be integrated into one dialog instead of two.

Do you have an Idea or some tips?

kiozen: The idea is to have something like that
image

The project is initially the project of the track. With the tool button on the right you open the usual  CGisWorkspace::selectProject() dialog. The rest is pre-set route name and other options like selecting subpoints or some future number of route points limit.

UEckhardt: Pushed a new commit with the modified dialog.

src/qmapshack/gis/CGisWorkspace.h
UEckhardt: The function is removed with [QMS-396] Modified TrkToRtDialog

src/qmapshack/gis/trk/CTrkToRteDialog.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include "CMainWindow.h"
+#include "gis/CGisDatabase.h"
+#include "gis/CGisListWks.h"
+#include "gis/db/CDBProject.h"
+#include "gis/db/CSelectDBFolder.h"
+#include "gis/db/CSetupFolder.h"
+#include "gis/gpx/CGpxProject.h"
+#include "gis/qms/CQmsProject.h"
+#include "helpers/CSettings.h"
+#include "gis/trk/CTrkToRteDialog.h"
```
kiozen, line 29: Get rid of include hell. As far as I can see you only use:

```
#include "canvas/CCanvas.h"
#include "CMainWindow.h"
#include "gis/CGisWorkspace.h"
#include "helpers/CSettings.h"
#include "gis/trk/CTrkToRteDialog.h"
```

src/qmapshack/gis/trk/CTrkToRteDialog.h
```diff
@@ -0,0 +1,83 @@
...
+
+**********************************************************************************************/
+
+#ifndef CTRKTORTEDIALOG_H
+#define CTRKTORTEDIALOG_H
+
+#include "gis/prj/IGisProject.h"
+#include "gis/CGisWorkspace.h"
+#include "ui_ITrkToRteDialog.h"
+#include <QDialog>
```
kiozen, line 25: Get rid of include hell: As far as I can see you only need:

```
#include "gis/prj/IGisProject.h"
#include "ui_ITrkToRteDialog.h"
```

src/qmapshack/gis/trk/CTrkToRteDialog.h
```diff
@@ -0,0 +1,83 @@
...
+private slots:
+    void slotProject();
+    void slotRouteChanged(const QString& text);
+
+private:
+    void setType(IGisProject::type_e& t);
+    void buttonBoxEnabled();
+
+    QString key;
+    bool saveSubPoints;
```
kiozen, line 77: This violates the "single source of truth" principle. The state of the checkbox should be sufficient. If you double it by a boolean variable you spent additional effort to keep the checkbox state and the variable in sync. with a good chance to fail and produce bugs.

src/qmapshack/gis/trk/CTrkToRteDialog.cpp
```diff
@@ -0,0 +1,97 @@
...
+    , project(project)
+{
+    setupUi(this);
+    SETTINGS;
+    cfg.beginGroup("TrkToRt");
+    saveSubPoints = cfg.value("saveSubPoints", false).toBool();
+    cfg.endGroup();
+
+    checkBoxSubPoints->setChecked(saveSubPoints);
+    key = IGisProject::getUserFocus();
```
kiozen, line 45: what is key used for?

src/qmapshack/gis/trk/CTrkToRteDialog.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+void CTrkToRteDialog::accept()
+{
+    saveSubPoints = checkBoxSubPoints->isChecked();
+    SETTINGS;
+    cfg.beginGroup("TrkToRt");
+    cfg.setValue("saveSubPoints", saveSubPoints);
+    cfg.endGroup();
+    QDialog::accept();
+    CCanvas::restoreOverrideCursor("~CTrkToRteDialog");
```
kiozen, line 73: Two times `CCanvas::restoreOverrideCursor("~CTrkToRteDialog");` ? See dtor for 2nd time.

src/qmapshack/gis/trk/CTrkToRteDialog.h
```diff
@@ -0,0 +1,83 @@
...
+    QString& routeName;
+    IGisProject* project;
+};
+
+#endif //CSELECTPROJECTDIALOG_H
```
kiozen, line 82: wrong comment ;)

src/qmapshack/gis/trk/CGisItemTrk.cpp
```diff
@@ -3402,3 +3410,32 @@ QMap<searchProperty_e, CGisItemTrk::fSearch> CGisItemTrk::initKeywordLambdaMap()
...
+    if(NOIDX == idx1)
+    {
+        routeName = getName();
+    }
+    else
+    {
+        routeName = getName() + QString(" (%1-%2)").arg(idx1).arg(idx2);
+    }
+
+    CTrkToRteDialog dlg(getParentProject(), routeName);
```
kiozen, line 3430: Be careful! getParentProject() does a dynamic cast and might return with a nullptr.

kiozen (changes requested):
Concerning size changes in the dialog. This should be done with care or best: leave the defaults.

Reasoning: Sizes are usually controlled by the window manager and user setup. If you leave the default font size Qt will use the font size defined by the window manager. If you set the size other than default (the property is bold in designer) Qt enforces that size. 

Additionally non of the other dialogs uses different font sizes for text. I am not sure why you want the explanatory text larger than the other text. Imho this is quite inconsistent.

Also the tool button size is given by the window manger. If you enforce a larger icon the tool button it will always stick to that size. And 32x32 isn't used anywhere else. 

Therefor for consistency reasons and scalability stick to the way the other dialogs are done. Reset properties to the default and stick with the defaults as much as possible. Always keep in mind: It might look good on your screen, but on another screen with different setup it might look very funny.

kiozen (changes requested):
For readability sake:

You use different techniques to exchange variables with the dialog. The one is to pass references and let the dialog alter te content of the reference. The other one is passing variables and provide a getter API for the results. 

Both is fine for me. However if you mix both styles the same dialog things are getting harder to read. Maybe you can clean up a bit.

src/qmapshack/gis/trk/CTrkToRteDialog.cpp
UEckhardt: My intention was to set the explanatory text to a smaller size. This looks good at my setup but obviously this had not the intended effect. I have checked the qt documentation, but I haven't found an other portable way, so I will reset the dialog to it's defaults.

src/qmapshack/gis/trk/CTrkToRteDialog.h
UEckhardt: I could remove only #include <QDialog> in the *.h file, but removed several includes in CTrkToRteDialog.cpp.

src/qmapshack/gis/trk/CGisItemTrk.cpp
UEckhardt: Could this really return null? Since I convert from an existing track, there must exist also a parent project where the track is assigned to.

UEckhardt: Is for getParentProject really a dynamic cast required? As far as i can see, it is always a downcast to IGisProject, so should be a static cast enough?

src/qmapshack/gis/trk/CTrkToRteDialog.h
UEckhardt: Due to the API cleanup this is now only a reference for returning this value in the constructor.

src/qmapshack/gis/trk/CTrkToRteDialog.cpp
UEckhardt: Was unused an is now removed.

src/qmapshack/gis/trk/CTrkToRteDialog.cpp
UEckhardt: Is removed.

src/qmapshack/gis/trk/CTrkToRteDialog.h
UEckhardt: Renamed.

src/qmapshack/gis/trk/CGisItemTrk.cpp
kiozen: Better safe than sorry. Never trust a getter method returning a pointer.

parent() returns a QTreeWidgetItem. That can be everything and must not be necessarily a IGisProject item. Also within the code GIS items are created without a parent (but that probably never applies here).

Anyways it's good practice to check and no it's not done everywhere because it's often forgotten (including code written by me).

UEckhardt: I have added a check.

src/qmapshack/gis/trk/CTrkToRteDialog.h
```diff
@@ -0,0 +1,58 @@
...
+ * Dialog Class for the track to route conversion.
+ */
+class CTrkToRteDialog : public QDialog, private Ui::ITrkToRteDialog
+{
+    Q_OBJECT
+public:
+    /**
+       @brief Shows the Dialog and preselects the project and route name
+     */
+    CTrkToRteDialog(IGisProject** project, QString& routeName, bool& saveSubPoints);
```
kiozen, line 38: We are doing c++ :wink: 

```
CTrkToRteDialog(IGisProject*& project, QString& routeName, bool& saveSubPoints);
...
IGisProject*& project;
```

ans so on.

kiozen (changes requested):
Almost there!

I attached a slightly tweaked version of ITrkToRteDialog.ui making a bit more use of the grid layout.

[ITrkToRteDialog.ui.zip](https://github.com/Maproom/qmapshack/files/6611013/ITrkToRteDialog.ui.zip)

src/qmapshack/gis/trk/CTrkToRteDialog.h
UEckhardt: I code too much plain C :smirk: Your review comments are incorporated.

kiozen:
Well done! Thanks for the work.

Issue #476, opened 2022-01-21
Color the map by elevation


slammyV:
### The parent ticket
[comment]: # (Add ticketnumber behind QMS-# )

QMS-476

### What do you plan to do
[comment]: # (Describe roughly what you want to do.)
I like to add the functionality of coloring the map by elevation for getting a better impression of the terrain shape when planning routes.

### The result will be
[comment]: # (A list of goals that have to be achieved to consider the ticket as done.)
- a color overlay over the map that reflects the elevation
- added an option to the DEM for enabling and disabling the elevation based shading
- a color scale can be enabled/disabled to be shown in the map to assign colors to elevation
- min/max elevation spinners in the DEM prop menu to rescale the color bar
- a slider to change the opacity of the color overlay of the map

huirad:
I like this feature. Thank you!

PR #477, opened 2022-01-21
[QM-476] Color the map by elevation

slammyV:
Added functionality in the DEM to color the map by the elevation
values including a switchable coresponding color scale legend and
min/max elevation spinners in the DEM property dialog.
Modified the scroll setting in the DEM list to allow for proper
scrolling of the DEM property tree widget.

QMS-#476

### What you have done:
[comment]: (Describe roughly.)
I've implemented the code that enables to color the map by elevation. Everything was implemented within the DEM folder.
Changed the scroll policy in the DEM list / property setup widget for a proper scrolling to enable that all widgets can be 
accessed properly. 

1. Open DEM Properties
2. Enable Elevation Shading check box and change elevation limits / toggle legend.

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/dem/CDemVRT.cpp
+++ b/src/qmapshack/dem/CDemVRT.cpp
@@ -222,7 +222,7 @@ void CDemVRT::draw(IDrawContext::buffer_t& buf)
     QPointF bufferScale = buf.scale * buf.zoomFactor;
     outOfScale = isOutOfScale(bufferScale);
 
-    if(outOfScale || (!doHillshading() && !doSlopeColor() && !doElevationLimit()))
+    if(outOfScale || (!doHillshading() && !doSlopeColor() && !doElevationLimit() && !doElevationShading()))
     {
         QThread::msleep(100);
         return;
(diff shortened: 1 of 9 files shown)


slammyV:
It is my first contribution. Hopefully I forgot nothing...

src/qmapshack/dem/CDemVRT.cpp
```diff
@@ -412,6 +412,57 @@ void CDemVRT::draw(IDrawContext::buffer_t& buf)
...
+                    elevationshading(data, w_used, h_used, img);
+
+                    drawTile(img, r, p);
+                }
+
+            }
+        }
+
+        // draw the elevation shade scale
+        if(doElevationShading())
```
kiozen, line 432: make it a method please. This is already too much spaghetti code without these additional lines.

kiozen (changes requested):
Well done. Just minor nitpicking.

slammyV (commented):
Hopefully now it is ok. Thankx for the comments.

mitxel-m:
Thanks for this new feature, very useful. 

 Tested and working for me. 

Screenshot:



screenshot with more detail, limits from 1000 to 1400, the colours match the contour lines. (DEM pixel is 5x5m)



The only thing I can suggest is about the legend, which now looks like this:



and I think it would be more intuitive the opposite, I mean the lowest value and the blue colour at the bottom, and the highest value and the red colour at the top . ( ... if possible)   

:-)

kiozen:
@slammyV Do you want to change the legend? Else I would press the button. ()If you change it use more const references in the legend code to avoid copies of QColor on the stack.)

slammyV:
@kiozen I will change this.

slammyV:
Swapped the legend upside down.
Made the color in the legend const.

kiozen:
Thanks!

Issue #483, opened 2022-02-11
Add alpha transparency based hillshading


JFritzle:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )
Currently implemented grayscale hillshading algorithm behaves as follows:  
The more hillshading magnitude is increased the more hillshading overlay gets opaque, showing finally a very impressive elevation model of landscape.

However, if opacity slider is totally at the right side, underlying map becomes invisible. The more slider moves towards the left side, the more map becomes visible, but always behind a gray haze. To reduce this gray haze, slider must move further to the left. Unfortunately, hillshading impression gets then weaker and weaker and finally disappears. Thus making it impossible to have an impressive hillshading effect and at the same time see the underlying map.

Current grayscale based hillshading where underlying map is nearly invisible:
![Hillshading with shades of gray](https://user-images.githubusercontent.com/62614244/153586243-8a252f06-7ed8-4d15-a138-ebf3bdfa04c5.jpg)

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)
I would like to have the choice to be able to switch between currently implemented grayscale based hillshading algorithm and an alpha transparency based algorithm, where hillshading overlay becomes the more opaque the steeper the slopes and becomes the more transparent the flatter the slopes and is totally transparent in the plain. Thus making the underlying map shining through.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)
-

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)
I am convinced that many QMS users would deploy this new feature as soon as it becomes available.

I for myself have already implemented the few necessary stuff into a clone of QMS dev branch. Therefore, I believe, the time to spend for adding this feature will be short and will probably be restricted to reviewing my code, easily merging into existing code and translating few texts to all languages supported by QMS.

Switching between existing grayscale based hillshading and alpha transparency based hillshading is realized by one single additional check button. Switching takes effect immediately. Thus, both hillshading methods can coexist.

If proposal for this feature is accepted, I am willing to contribute my enhancement.

Alpha transparency based hillshading with same opacity slider and magnitude setting:
[...]

PR #486, opened 2022-02-16
[QMS-483] Add alpha transparency based hillshading

JFritzle:

QMS-#483

### What you have done:
Added alpha transparency based slope shading layer where slope angle is mapped to transpareny.
Slope angle is evaluated using already implemented algorithm.
Layer at slope angle 0 degrees is fully transparent, layer at slope angle 90 degrees is fully opaque.
Slider may reduce slope angle down to 30 degrees where layer is fully opaque.

1. Open window Dig. Elev. Model (DEM)
2. Open VRT entry
3. Check _Slope Shading_ checkbutton
4. Position corresponding slider at the left side
Scaling factor is now 0.25 and slope relief 's opacity is at it's minimum.
5. Move corresponding slider to the right side
Slope relief becomes more and more stronger and slopes are becoming more opaque (darker).
6. Position corresponding slider at the right side
Scaling factor is now 3.0 and slope angles equal to or greater than 30 degrees are mapped to fully opaque.
7. Position corresponding slider 3 steps from left side
Scaling factor is now 1.0 and slope angle range from 0 to 90 degrees is mapped exactly from fully transparent to fully opaque.

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 [QMS-470] Windows build scripts: adapt for release v1.16.1
 |QMS-476] Color the map by elevation
+[QMS-483] Add alpha transparency based hillshading
 V1.16.1
 [QMS-396] Save route sub-points as ordinary route points in case of track routing
 [QMS-400] Set focus in text field
--- a/src/qmapshack/dem/IDem.cpp
+++ b/src/qmapshack/dem/IDem.cpp
@@ -115,6 +115,8 @@ void IDem::saveConfig(QSettings& cfg)
 
     cfg.setValue("doHillshading", bHillshading);
     cfg.setValue("factorHillshading", factorHillshading);
+    cfg.setValue("doSlopeShading", bSlopeShading);
+    cfg.setValue("factorSlopeShading", factorSlopeShading);
     cfg.setValue("doSlopeColor", bSlopeColor);
 
     cfg.setValue("gradeSlopeColor", gradeSlopeColor);
@@ -140,6 +142,8 @@ void IDem::loadConfig(QSettings& cfg)
 
     bHillshading = cfg.value("doHillshading", bHillshading     ).toBool();
     factorHillshading = cfg.value("factorHillshading", factorHillshading).toFloat();
+    bSlopeShading = cfg.value("doSlopeShading", bSlopeShading     ).toBool();
+    factorSlopeShading = cfg.value("factorSlopeShading", factorSlopeShading).toFloat();
     bSlopeColor = cfg.value("doSlopeColor", bSlopeColor      ).toBool();
     gradeSlopeColor = cfg.value("gradeSlopeColor", gradeSlopeColor  ).toInt();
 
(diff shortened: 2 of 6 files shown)


mitxel-m:
I have tested the PR and it looks good.
In general I prefer the classic shading that takes into account aspect and slope to get a more accurate idea of the relief. May be I have my eye  trained to that kind of shading.
However this shading based only on the slope is also good to see other kind of details. For example, the step-like shapes that are usually visible in high resolution DEMs look like contour lines with the slope shading.

Some screenshots to compare : 

First hillshading, second slope Shading. 
all of them with a 5x5m DEM:

changelog.txt
```diff
@@ -1,3 +1,4 @@
+[QMS-483] Add alpha transparency based hillshading
```
kiozen, line 1: please keep order of tickets

src/qmapshack/dem/IDem.cpp
```diff
@@ -184,6 +188,11 @@ void IDem::slotSetFactorHillshade(int f)
     }
 }
 
+void IDem::slotSetFactorSlopeShade(int f)
+{
+    factorSlopeShading = f / 100.;
```
kiozen, line 193: what is this factor shift good for?

kiozen (changes requested):
Please squash the commit history:

`git rebase -i HEAD~11`

follow the instructions displayed. When done:

`git push --force`

changelog.txt
JFritzle: Now kept order of tickets

src/qmapshack/dem/IDem.cpp
JFritzle: Qt sliders work with integer values. Therefore mapping slope shading slider's integer value range [25 .. 300] to slope shading's scaling range [0.25 .. 3.0] is required.
Analogue to hill shading slider's value mapping factor _factorHillshading_.

kiozen:
Thanks for the work!

mitxel-m:
> @mitxel-m, @JFritzle: Could someone please describe the required quality of the used DEM data and the terrain used for getting the nice snapshots shown in this thread?
> 
> Does the used OS have an influence on the obtained result?

@wthaem  I think the OS has no influence. In my opinion the most important is the resolution of the DEM,

The DEM used for the snapshot is a 5x5m DTM. The original is opendata provided by the Spanish IGN with a vertical resolution about 2cm, and comes in .ASC  format . As this means very large files, I convert it to compressed TIF format and also save the elevation value as an integer, . So the vertical resolution of the example is 1m  (which is enough for me ).

Here is another snapshot to compare the 5m DEM and the 25m DEM. Both are from the same source, and both with a vertical resolution of 1m. 





And the same with ViefinderPanoramas 1 arcsecond

wthaem:
Thanks! This is what I expected! The last snapshot is similar to the ones I've seen. Will try to find better DEM data.

JFritzle:
Unfortunately result `NOFLOAT `of `IDem::slopeOfWindowInterp` called by `IDem::slopeShading` is currently not taken into account. This could potentially lead to incorrect slope shading opactity. Sorry!

A further pull request concerning 1 file `IDem.cpp` would resolve issue. Same issue number or new one?

kiozen:
New one

⁣BlueMail for Android herunterladen ​

Issue #508, opened 2022-04-11
[QMS-342] Create common interface for poi items


hhornburg:
### The parent ticket
[comment]: # (Add ticketnumber behind QMS-# )

QMS-#342

### What do you plan to do
[comment]: # (Describe roughly what you want to do.)

Create a common interface for all poi items to avoid code duplication later on when adding support for new poi files. A poi item is a single poi that is displayed on the map and can be converted to a waypoint. 

### The result will be
[comment]: # (A list of goals that have to be achieved to consider the ticket as done.)

- [x] define interface of `IPoiItem`
- [x] adjust `CPoiItemPOI` to inherit from `IPoiItem` and follow interface
- [x] adjust `CgarminPoint` to inherit from `IPoiItem` and follow interface
- [x] change `IPoiItem` to `IPoiItem*` where necessary, as pre 5.13 QMS needs a default constructor for `QVector`, `QMap` etc and an abstract class is not default constructible. Also, an abstract type is an invalid return type.

PR #506, opened 2022-04-10
[QMS-505] [QMS-342] Rename Classes

pingurus:
Renaming for clarity and as preparation for new features

### What is the linked issue for this pull request:
[comment]: # (Add the ticket number behind QMS-# )

QMS-#505

### What you have done:
[comment]: # (Describe roughly.)

Renamed the classes as described in the issue.

### Steps to perform a simple smoke test:
[comment]: # (List the steps.)

1. Try to use both `.img` maps and `.poi` files and see that they still work

### Does the code comply to the coding rules and naming conventions [Coding Guidelines](https://github.com/Maproom/qmapshack/wiki/DeveloperCodingGuideline):

- [ ] no - For now CPoiItem is named as a class but actually still is a struct. This way no actual code has to be changed to demonstrate that the renaming was successful. The naming will be correct again after implementing the next dev step. (However @kiozen, if you prefer, I can make it an all-public class for the sake of protocol).

### Is every user facing string in a tr() macro?

- [x] yes

### Did you add the ticket number and title into the changelog? Keep the numeric order in each release block.

- [x] yes, I didn't forget to change changelog.txt

PR #509, opened 2022-04-11
[QMS-508] [QMS-342] Create common interface for poi items

pingurus:
**Requires #506 to be merged**

QMS-#508

### What you have done:

Implemented the points described in the issue

1. Check that everything works as before (Img and Poi files)

- [x] yes

- [x] yes


kiozen:
I guess you have to rebase this one now.

src/qmapshack/canvas/CCanvas.cpp
```diff
@@ -1133,19 +1133,19 @@ void CCanvas::displayInfo(const QPoint& px)
     QToolTip::hideText();
 }
 
-void CCanvas::findPoiCloseBy(const QPoint& px, QSet<IPoiItem>& poiItems, QList<QPointF>& posPoiHighlight)  const
+void CCanvas::findPoiCloseBy(const QPoint& px, QSet<const IPoiItem*>& poiItems, QList<QPointF>& posPoiHighlight)  const
```
kiozen, line 1136: passing pointers a long methods as return values is super risky.  What happens to `poiItems`? Will it be stored? You can't tell. What happens to the objects pointed to in `poiItems`? You can't tell. They can be destroyed at any time and leaving a set of dangling pointers.

kiozen (changes requested):

src/qmapshack/canvas/CCanvas.cpp
pingurus: My idea was that the `IPoiFile` object "owns" the `IPoiItem` it returns and will take care of its deletion when the file is closed and keeps it stored until then. The pointers should not be stored anywhere, although that is not bullet-proof and I have to check whether I've been true to that.

What are the other options though? const references can't be put into QSets afaik and handing down the items by value creates unnecessary copies all along the way. Shared Pointers would be an option, but as far as I remember you weren't too fond of these :smiley:

kiozen: The policy so far is (If I recall correctly, feel free to find examples contradicting):

* Pass references if possible.
* If references are not possible, pass copies.

We do have a very complex GUI in QMapShack, allowing the user to do a lot of stuff out of the order we developers expect. And there are a bunch of developers hacking the code. So we really should take care to make the code bulletproof. 

For example all GIS items are passed along by their keys, never by a pointer. If I want a pointer I have to use the key to get access to the object and I have to check if I really get an object. By that an item can be deleted while I still keep that list of keys.

If the additional copies become a problem we have to think about a better approach/concept. Passing pointers within a container is always like being blind folded and making a step towards the abyss. The first few steps might be ok but there will be the last fatal step sooner or later.

As it previously has been copies - and I think this is because of the old Garmin IMG based POI code uses copies - I would say that past-me has made the decision that the number of copies is countable and it's ok.

pingurus: > As it previously has been copies - and I think this is because of the old Garmin IMG based POI code uses copies - I would say that past-me has made the decision that the number of copies is countable and it's ok

I think there number of POIs possibly being on screen is much higher on outer zoom levels than those that come from a garmin map. Also, the items changed from being a relatively small struct to a full polymorphic class with a probably larger footprint, So I think we should go for the key based approach then, as the items have a key anyway.

kiozen:
I close it for now as there is no progress.

PR #552, opened 2022-08-23
[QMS-551] Set QDateTime short format back to Qt::ISODate

kkarsten62:

QMS-551

### What you have done:
Set QDateTime short format back to Qt::ISODate in IUnit::datetime2string for qmapshack and qmaptool.
Note:
QDateTime is overloaded and can take two arguments:
enum Qt::DateFormat for Qt::ISODate
and
const QString for QLocale().dateTimeFormat(QLocale::LongFormat)

1. Connect Garmin device with fit format ==> ISODate will be used as track (file) name.
2. Set in View / Setup Time zone long format, Exit QMS and restart
    Edit a track ==> in track info long format will be shown

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -17,6 +17,7 @@ V1.XX.X
 [QMS-537] Moving items to drop zone removes them from workspace
 [QMS-540] MacOS: Updating bundle with GEOS and proper GDAS binaries
 [QMS-542] Changed drag`n drop in workspace
+[QMS-551] Set QDateTime short format back to Qt::ISODate
 
 V1.16.1
 [QMS-396] Save route sub-points as ordinary route points in case of track routing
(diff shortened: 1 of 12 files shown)


kiozen:
After having some time to play with it I think it would be better to keep the short format as it  really is short and human readable. 

To fix the FIT track names it would need to change ` static QString dateTimeAsString(quint32 timestamp)`  in `src/qmapshack/gis/fit/serialization.cpp:51`.  In that case you could select any format you consider as best practice. 

I know you have suggested something like that already.  Sorry. I feel guilty.

kkarsten62:
IsoFormat is quite ok for me for the file/project name (as is was previously).
But I would like to have timezone information, so a call to IUnit::datetime2string is needed.

So now we have 3 formats, "short" and "long" for the user (change by View / timezone) and "Iso" for the FIT file name.

kiozen:
Maybe passing an enum instead of two flags is the better approach. See attached patch:
[0001-QMS-551-Make-time-format-an-enum.patch.zip](https://github.com/Maproom/qmapshack/files/9482628/0001-QMS-551-Make-time-format-an-enum.patch.zip)

Issue #557, opened 2022-09-10
[QMS-557] Splitting a "long" track in multiple tracks


kaiserls:
### The parent ticket
[comment]: # (Add ticketnumber behind QMS-# )

QMS-#557

### What do you plan to do
[comment]: # (Describe roughly what you want to do.)
I would like to implement a feature to split a track into multiple tracks.
This allows to load long tracks on old gps devices, which only allow a small number of track points per track, e. g. 500.
This feature is somewhat similar to the "cut" feature but removes the necessity to choose the cut point by hand, which may be hard to get right if the density of track points is varying, and makes it easier to process very long tracks, which would need multiple cuts.

### The result will be
[comment]: # (A list of goals that have to be achieved to consider the ticket as done.)
- [ ] Implement dialog to ask for number of cuts
- [ ] Implement cut algorithm
- [ ] Write "glue" code
- [ ] Design new icon

### Additional comments
- I will provide a pull request
- I am happy about suggestions for improvements

Could one of the developers tell me, if this pull request has a chance to be approved?
Have a nice weekend!
Greetings Lars

![image](https://user-images.githubusercontent.com/71793357/189505437-90d96234-1b15-4e13-b2b3-8381dd801e12.png)

![Screenshot from 2022-09-11 01-43-15](https://user-images.githubusercontent.com/71793357/189505412-acefcae5-99d1-4838-815e-f19546ccf53f.png)

kiozen:
Yes it will be approved. But make it a track filter instead of another stand alone dialog. Add it to the miscellaneous section. There is already a cutting tool for laps.

kiozen:
Closing this one as it is done. But there is a regression. https://github.com/Maproom/qmapshack/issues/560

PR #559, opened 2022-09-16
[QMS-557] Add FilterSplitTrack

kaiserls:
Sometimes tracks have too many points to be transfered to a gps device. This filter splits the track into multiple smaller tracks. The user can select the number of new tracks.

QMS-#557

### What you have done:
I implemented a filter to split a track into multiple tracks. It can be found in the "miscellaneous" filter group.
This allows to load long tracks on old gps devices, which only allow a small number of track points per track, e. g. 500.
This feature is somewhat similar to the "cut" feature but removes the necessity to choose the cut point by hand, which may be hard to get right if the density of track points is varying, and makes it easier to process very long tracks, which would need multiple cuts.

1. Open a track
2. Right-click on the track, Edit, Filter, Miscellaneous
3. Split track into multiple shorter tracks: Select number of tracks to split into and press apply button.
4. Confirm dialog and check if new tracks are generated

- [ ] yes, but I clazy returned with compilation errors because it doesn't understand the includes.

- [ ] yes, but I didnt understand how to generate the ts files to add translations.


kiozen:
What does the `m` in the spin box to select the number of tracks stand for?

src/qmapshack/gis/trk/filter/filter.cpp
```diff
@@ -655,6 +655,39 @@ void CGisItemTrk::filterLoopsCut(qreal minLoopLength)
...
 }
 
+void CGisItemTrk::filterSplitTrack(qint8 nTracks)
+{
+    IGisProject* project = CGisWorkspace::self().selectProject(false);
+    if(nullptr == project)
+    {
+        return;
+    }
+    qint32 overlapNodes = 1;
```
kiozen, line 665: as this never changes it can be a constexpr

src/qmapshack/gis/trk/filter/filter.cpp
```diff
@@ -655,6 +655,39 @@ void CGisItemTrk::filterLoopsCut(qreal minLoopLength)
...
+        new CGisItemTrk(tr("%1 (Part %2)").arg(trk.name).arg(part), segStartIdx, segEndIdx, trk, project);
+        // update remaining
+        segNodes = segEndIdx-segStartIdx+1;
+        remaining = remaining - segNodes;
+        if(remaining>0)
+        {
+            remaining+=overlapNodes;
+        }
+        // for next segment
+        segStartIdx = segEndIdx + (1-overlapNodes); // if this was last, never used again
```
kiozen, line 684: As overlapNodes is always 1 the (1-overlapNodes) does not make much sense

src/qmapshack/gis/trk/filter/filter.cpp
```diff
@@ -655,6 +655,39 @@ void CGisItemTrk::filterLoopsCut(qreal minLoopLength)
...
+    qint32 segStartIdx = 0;
+    qint32 segEndIdx = segNodes-1;
+    qint32 remaining = cntTotalPoints;
+
+    qint8 part=0;
+    while(remaining>0)
+    {
+        new CGisItemTrk(tr("%1 (Part %2)").arg(trk.name).arg(part), segStartIdx, segEndIdx, trk, project);
+        // update remaining
+        segNodes = segEndIdx-segStartIdx+1;
```
kiozen, line 677: What is this good for? segNodes shouldn't change.  It's just the last segment that might differ in size depending on the remaining points.

src/qmapshack/gis/trk/filter/filter.cpp
```diff
@@ -655,6 +655,39 @@ void CGisItemTrk::filterLoopsCut(qreal minLoopLength)
...
+        remaining = remaining - segNodes;
+        if(remaining>0)
+        {
+            remaining+=overlapNodes;
+        }
+        // for next segment
+        segStartIdx = segEndIdx + (1-overlapNodes); // if this was last, never used again
+        segEndIdx = segStartIdx + qMin(remaining, segNodes)-1; // if this was last, never used again
```
kiozen, line 685: I haven't played all edge cases in my mind but wouldn't it be possible that by using qMin we might get an additional track with just one or two points?

In other words: You expect a number of track but you create tracks by consuming points. This might end up in single track too much.

kiozen (changes requested):
I think the algorithm in filterSplitTrack has to be simplified.  Rather than looping on the remaining points it should loop over the number of expected tracks.

kiozen:
The PR is ready to merge from my point of view. If you do not want to add anything I will merge it.

kaiserls:
Besides updating the language files, which I don't know how to do correctly, I have no further changes to implement.
Thank you for your review!

kiozen:
No worries about translation. This will be done by someone else.

Issue #560, opened 2022-09-18
Missing trackpoints after splitting a track


wthaem:
### Describe the bug
[comment]: # (A clear and concise description of what the bug is)

After applying the "Split track into multiple shorter tracks" filter to a track the union of the shorter tracks can be smaller than the original track.

### What have you done to circle down the problem?
[comment]: # (Did you compare to previous versions? What versions? Did you check if others have the same problem?)

1. Create a track with 5 trackpoints.
1. Select "3" as number of tracks in the mentioned track filter.
1. Run the filter and save the new tracks.
1. The last trackpoint of the original track can't be found in the union of the new tracks.

### To Reproduce
[comment]: # (List the steps to reproduce the behavior)

See previous point

### Expected behavior
[comment]: # (A clear and concise description of what you expected to happen)

The union of the new tracks must be the original track, i.e. the union of all trackpoints of the new tracks must be the set of all trackpoints of the original track.

### Screenshots
[comment]: # (If applicable, add screenshots to help explain your problem)

None

### Attachments
[comment]: # (Can you provide an example file to trigger the problem?)

None

### Tracebacks
[comment]: # (Add your backtrace below if you have one. If QMapshack crashes the fastest way to get help is a backtrace.
For Linux see: https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux.
For Windows it would need a debug build and running QMapshack in Visual Studio.)

None

### Desktop
[comment]: # (please complete the following information)

 - OS: Windows 10
 - QMapShack Version: 
 
     ~~~
     commit ff9ac6dde35e1c1836f2229e76bf47653cd36e3c (origin/dev, origin/HEAD, dev)
     Author: Lars Kaiser <lars.g.kaiser@gmx.de>
     Date:   Sat Sep 17 13:07:13 2022 +0200

     [QMS-557] Simplify filterSplitTrack
     ~~~
     
 - Window Manager: Windows

### Additional context
[comment]: # (Add any other context about the problem here)

When setting the number of new tracks to 1,2, or 4 the bug doesn't happen.

kiozen:
@kaiserls can you please have a look

wthaem:
@kaiserls: Be careful when changing your algorithm: tracks should have 2 points or more!

kaiserls:
You are right with your reports. I shouldn't have hurried with rewriting the splitting filter.

[...]

PR #562, opened 2022-09-29
[QMS-560] Fix FilterSplitTrack

kaiserls:

QMS-#560

### What you have done:
I fixed the filter function filterSplitTrack to include all points of the original track and not generate empty tracks or tracks with just one point. In addition tracks points are distributed more equally over the track "segments".

1. Open a track
2. Right-click on the track, Edit, Filter, Miscellaneous
3. Split track into multiple shorter tracks: Select number of tracks to split into and press apply button.
4. Confirm dialog and check if new tracks are generated and contain the correct track points

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,6 +19,7 @@ V1.XX.X
 [QMS-542] Changed drag`n drop in workspace
 [QMS-551] Set QDateTime short format back to Qt::ISODate
 [QMS-557] Add FilterSplitTrack
+[QMS-560] Fix FilterSplitTrack to include all track points
 
 V1.16.1
 [QMS-396] Save route sub-points as ordinary route points in case of track routing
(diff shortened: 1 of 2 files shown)


kiozen:
@wthaemelt can you please test.

wthaem:
The algorithm seems ok now.

*Question:* Is it a good idea to create an empty project in the case that nr. of points in track <= nr. of splits?

kiozen:
@kaiserls any answer to @wthaem question?

kaiserls:
Sry, I didn't understand that the question is targeting me.
I think an empty project should not created and the user should be explicitly notified, why the splitting of the track failed. However I could not find an example of how this is done in QMapShack. Should I create a QMessageBox?
`QMessageBox::warning(CMainWindow::getBestWidgetForParent(), tr("Error"), tr("The numberpoints in the track must be higher than the number of tracks to split into"), QMessageBox::Abort, QMessageBox::Abort);`

kiozen:
@kaiserls yes please.

kiozen:
@wthaem satisfied with the fix?

PR #588, opened 2023-02-16
Added "Ok" button to DetailsWpt dialog

rayslava:
Usually I use a tiling WM without any window decorations, so adding a waypoint requires to press hotkeys to close the dialog with waypoint description.
This commit just adds an "Ok" button. I do understand that location isn't usual bottom-right, but at least it's aligned with existing grid and doesn't affect layout.


kiozen:
I can't process this PR as it simply breaks the required process. Please read the wiki for details:
https://github.com/Maproom/qmapshack/wiki/DeveloperCommitCode
https://github.com/Maproom/qmapshack/wiki/DocFaqProjectSite

On the topic:
Do we really need an "ok" button? This dialog is change-as-you-type. The data is not just taken over on some "ok". A button would suggest that. 

In my workflow I usually close such kind of dialogs with the escape-key. To my knowledge this is quite universal between window managers.

Issue #622, opened 2023-07-29
Update BRouter setup


wthaem:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

The BRouter project is now maintained in [Github](https://github.com/abrensch/brouter). The revision history page [http://brouter.de/brouter/revisions.html](http://brouter.de/brouter/revisions.html) used in the QMS BRouter setup isn't updated anymore. Thus, the BRouter setup in QMS can't be used anymore to update the BRouter offline version.

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

Use the new revision history page [https://github.com/abrensch/brouter/blob/master/docs/revisions.md](https://github.com/abrensch/brouter/blob/master/docs/revisions.md) in the QMS BRouter setup to support BRouter updates.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

None.

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

Currently, BRouter can't be updated using the QMS BRouter setup.

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

kiozen:
@ntruchsess Can you please have a look?

ntruchsess:
> @ntruchsess Can you please have a look?

sure, will do :-)

ntruchsess:
@kiozen @wthaem fix implemented in https://github.com/Maproom/qmapshack/pull/625
(I used the github-releases page instead of the revisions.md as the urls in the revisions.md are maintained manually and do not work, they point to a repository-location that does not exist any more)

PR #625, opened 2023-08-01
[QMS-622] change installation-source of brouter to github

ntruchsess:

QMS-#622

### What you have done:
the url that is used to download brouter binaries from has been changed to brouter releases page on github

1. Open brouter setup-wizzard
2. choose 'local installation'
3. click 'update existing BRouter installation'
4. click 'v1.7.2'
5. click 'Download and Install'
6. the log-output should end with 'installation successful'

- [X] yes

- [X] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,5 @@
+[QMS-622] Update BRouter setup (install from github)
+
 V1.17.0
 [QMS-429] Bad OSM Tag formatting crashes QMS
 [QMS-470] Windows build scripts: adapt for release v1.16.1
--- a/src/qmapshack/gis/rte/router/brouter/CRouterBRouterDownloadPage.cpp
+++ b/src/qmapshack/gis/rte/router/brouter/CRouterBRouterDownloadPage.cpp
@@ -1,0 +1,266 @@
+/**********************************************************************************************
+    Copyright (C) 2023 Norbert Truchsess <norbert.truchsess@t-online.de>
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+**********************************************************************************************/
+
+#include "gis/rte/router/brouter/CRouterBRouterDownloadPage.h"
+
+#include <JlCompress.h>
+
+#include <QCheckBox>
+#include <QDir>
+#include <QJsonArray>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QMessageBox>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
+#include <QPushButton>
+
+#include "gis/rte/router/brouter/CRouterBRouterSetup.h"
+
+CRouterBRouterReleaseItem::CRouterBRouterReleaseItem(const QString& _name, const QString& _description,
+                                                     const QString& _url)
+    : name(_name), description(_description), url(_url) {}
+
+CRouterBRouterDownloadPage::CRouterBRouterDownloadPage() : QWizardPage() {
+  networkAccessManager = new QNetworkAccessManager(this);
+  networkAccessManager->setRedirectPolicy(QNetworkRequest::NoLessSafeRedirectPolicy);
+  connect(networkAccessManager, &QNetworkAccessManager::finished, this,
+          &CRouterBRouterDownloadPage::slotNetworkAccessManagerReplyReceived);
+  model = new QStringListModel(this);
+}
+
+CRouterBRouterDownloadPage::~CRouterBRouterDownloadPage() {}
+
+void CRouterBRouterDownloadPage::initialize(QListView* listVersions, QTextBrowser* textVersionDetails,
+                                            QLabel* labelInstallLink, QPushButton* pushInstall,
+                                            QTextBrowser* textLocalInstall, CRouterBRouterSetup* setup) {
+  this->listVersions = listVersions;
+  this->textVersionDetails = textVersionDetails;
+  this->labelInstallLink = labelInstallLink;
+  this->pushInstall = pushInstall;
+  this->textLocalInstall = textLocalInstall;
+  this->setup = setup;
+  this->listVersions->setModel(model);
+}
+
+void CRouterBRouterDownloadPage::begin() {
+  setComplete(setup->isLocalBRouterValid);
+  textVersionDetails->setVisible(false);
+  textVersionDetails->clear();
+  textLocalInstall->setVisible(false);
+  textLocalInstall->clear();
+  labelInstallLink->setText(tr("no brouter-version to install selected"));
+  pushInstall->setEnabled(false);
+  wizard()->setOption(QWizard::HaveCustomButton1, false);
+  downloadUrl = QUrl();
+
+  QNetworkRequest request(QUrl(setup->expertBinariesUrl));
+  request.setRawHeader(QByteArray("Accept"), QByteArray("application/vnd.github+json"));
+  request.setRawHeader(QByteArray("X-GitHub-Api-Version"), QByteArray("2022-11-28"));
+  QNetworkReply* reply = networkAccessManager->get(request);
+  reply->setProperty("requestType", eRequestVersions);
+}
+
+void CRouterBRouterDownloadPage::slotListVersionsClicked(const QModelIndex& index) {
+  const QString& version = index.data().toString();
+  for (const CRouterBRouterReleaseItem& item : qAsConst(releaseItems)) {
+    if (item.name == version) {
+      downloadUrl = QUrl(item.url);
+      textVersionDetails->setText(item.description);
+      textVersionDetails->setVisible(true);
+      textLocalInstall->setVisible(false);
+      labelInstallLink->setText(QString(tr("selected %1 for download and installation")).arg(downloadUrl.fileName()));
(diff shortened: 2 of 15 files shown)


kiozen:
Is there any chance to get a better guidance on the Gitub page? I think the average user has no clue what to do (scroll down, find the correct file, press the link, etc.)

Is it possible to extract a list of files form the page? And only list those files that are the jar archive?

ntruchsess:
there is an api that can be used to query all release-metadata including the download-urls of the assets. That could be used to increase the usability of the installer, but it requires to implement a whole new dialog. The current implementation just loads the webpage into the internal browser and waits for the user to click a link on that page. To implement an api-based alternative will certainly take some time.

kiozen:
Right now it's very confusing. A normal user does not know what to do. There is this cropped GitHub page. You have to know that you need to scroll down. You have to know what link to press. And there are a lot of options to fail as there are many other links to press as the wrong one.

It would at least need some instructions to help the user. But even with that most will be scared away.

mitxel-m:
I agree that it is not easy for the average user, due to all the stuff on the github page.

Another approach could be to point to this link, which at least shows the revision list a bit cleaner

https://github.com/abrensch/brouter/tags

as seen here:

BRoutersetup

It would be necessary to give the GUI a bit more space and add more explicit instructions.

```
Here is a list with the latest versions of BRouter:
1- Click on the 'Downloads' link of the desired version.
2- In the bottom left corner, click on the 'Download and Install' button.
```
----
 
not a great solution just a bit better

ntruchsess:
@kiozen @mitxel-m : I did a complete rewrite using the github-api to retrieve available releases + metadata. No more direct display of github-page, everything is done using normal qt-widgets.

changelog.txt
mitxel-m, line 1: IMHO it would be better to place the text widget that shows the version changelog right below the version list widget. 
Doing so the 'Download and install' button would remain always at the same place.

Something like this
pantallazo_069

Instead of this
pantallazo_068

ntruchsess: > No doubt. Now is really clear
> 
> Only one suggestion :
> 
> Right now when you cilck on a version, the widget with its changelog info appears at the bottom, and it forces the 'Download' button to move from the bottom to the middle of the dialog, which is unusual.
> 
> IMHO it would be better to place the text widget that shows the version changelog right below the version list widget. Doing so the 'Download and install' button would remain always at the same place.
> 
> Something like this pantallazo_069
> 
> Instead of this pantallazo_068

@mitxel-m I did change the dialog according to your suggestion adding a separate textBrowser for the description in between the list of versions and the download-button

mitxel-m (commented):
Checked and working for me. 

Thanks for taking care of it.

src/qmapshack/gis/rte/router/brouter/CRouterBRouterDownloadPage.cpp
```diff
@@ -0,0 +1,256 @@
...
+#include <QPushButton>
+
+CRouterBRouterReleaseItem::CRouterBRouterReleaseItem(const QString &_name, const QString &_description, const QString &_url) : name(_name), description(_description), url(_url) {}
+
+CRouterBRouterDownloadPage::CRouterBRouterDownloadPage() : QWizardPage() {
+  networkAccessManager = new QNetworkAccessManager(this);
+  networkAccessManager->setRedirectPolicy(QNetworkRequest::NoLessSafeRedirectPolicy);
+  connect(networkAccessManager, &QNetworkAccessManager::finished, this,
+          &CRouterBRouterDownloadPage::slotNetworkAccessManagerReplyReceived);
+  releaseItems = new QList<CRouterBRouterReleaseItem>();
```
kiozen, line 40: memory leak! `releaseItems` is never destroyed. Why do you allocate it on the heap in the first place?

src/qmapshack/gis/rte/router/brouter/CRouterBRouterDownloadPage.cpp
```diff
@@ -0,0 +1,256 @@
...
+  QNetworkRequest request(QUrl(setup->expertBinariesUrl));
+  request.setRawHeader(QByteArray("Accept"), QByteArray("application/vnd.github+json"));
+  request.setRawHeader(QByteArray("X-GitHub-Api-Version"), QByteArray("2022-11-28"));
+  QNetworkReply* reply = networkAccessManager->get(request);
+  reply->setProperty("requestType", eRequestVersions);
+}
+
+void CRouterBRouterDownloadPage::slotListVersionsClicked(const QModelIndex& index) {
+  const QString &version = index.data().toString();
+  for(const CRouterBRouterReleaseItem &item : *releaseItems)
```
kiozen, line 75: Use `qAsConst` to avoid detaching the container. This applies to all of you ranged loops in this file.

src/qmapshack/gis/rte/router/brouter/CRouterBRouterSetupWizard.cpp
```diff
@@ -18,12 +18,8 @@
...
 
-#include <JlCompress.h>
-
 #include <QFileDialog>
 #include <QMessageBox>
-#include <QNetworkReply>
-#include <QNetworkRequest>
 
 #include "CMainWindow.h"
 #include "gis/rte/router/brouter/CRouterBRouterSetup.h"
```
kiozen, line 25: #include "helpers/CWebPage.h" is not used directly. Remove it and fix any indirect dependencies.

kiozen (changes requested):
I switched formatting from uncrustify to clang using the `.clang-format` definition in the root. It would be nice to use that because the next time someone is editing that file with automated formatting there will be a lot of white space changes.

I also recommend to use qtcreator with clang-tidy enabled. It gives lot of good hints how to cleanup and improve your code.

kiozen:
That is solved really nice, now!

src/qmapshack/gis/rte/router/brouter/CRouterBRouterDownloadPage.cpp
ntruchsess: fixed

src/qmapshack/gis/rte/router/brouter/CRouterBRouterDownloadPage.cpp
ntruchsess: fixed

src/qmapshack/gis/rte/router/brouter/CRouterBRouterSetupWizard.cpp
ntruchsess: fixed

ntruchsess:
> I switched formatting from uncrustify to clang using the `.clang-format` definition in the root. It would be nice to use that because the next time someone is editing that file with automated formatting there will be a lot of white space changes.
> 
> I also recommend to use qtcreator with clang-tidy enabled. It gives lot of good hints how to cleanup and improve your code.

will use clang-format and tidy from now on :+1:

src/qmapshack/gis/rte/router/brouter/CRouterBRouterDownloadPage.h
```diff
@@ -0,0 +1,84 @@
...
+
+  QList<CRouterBRouterReleaseItem> releaseItems;
+  QStringListModel* model;
+  QNetworkAccessManager* networkAccessManager;
+  QListView* listVersions;
+  QTextBrowser* textVersionDetails;
+  QLabel* labelInstallLink;
+  QPushButton* pushInstall;
+  QTextBrowser* textLocalInstall;
+  CRouterBRouterSetup* setup;
```
kiozen, line 78: as these are initialized late by `initialize()` it might be better to default them to `{nullptr}`. It shouldn't be really an issue as this is all internal stuff, but just for the chance programs crash more beautiful with nullptr rather than some dangling pointer.

kiozen (changes requested):
When I start with a complete new configuration (-c complete_new_config.ini) I get this screen:

image

No way to proceed unless I download and install BRouter. This is a bit confusing.

src/qmapshack/gis/rte/router/brouter/CRouterBRouterDownloadPage.h
kiozen: The wording in the dialog is a bit strange, too;

kiozen: Ah... now I get it. The dialog is ok, just the wording is very confusing :laughing:

ntruchsess:
> When I start with a complete new configuration (-c complete_new_config.ini) I get this screen:
> 
> image
> 
> No way to proceed unless I download and install BRouter. This is a bit confusing.

from what I can see this is working as designed. The brouter.jar it finds in the current directory seems to be an outdated version that cannot be started any longer (due to a change we did implement quite a while ago to support the newer versions). Any suggestions how to design the dialog to deal with that scenario in a less confusing way?

wthaem:
What about `is not a valid or an outdated BRouter jarfile`?

src/qmapshack/gis/rte/router/brouter/CRouterBRouterDownloadPage.h
ntruchsess: > nullptr

nullptr-declaration is fixed

ntruchsess:
> What about `is not a valid or an outdated BRouter jarfile`?

adjusted accordingly

kiozen:
This is how I read it:

image

I would expect something like:

"It seems there is an existing BRouter installation but the *.jar file is not valid or outdated. Choose a *.jar file or create/update the installation."

ntruchsess:
> This is how I read it:
> 
> image
> 
> I would expect something like:
> 
> "It seems there is an existing BRouter installation but the *.jar file is not valid or outdated. Choose a *.jar file or create/update the installation."

I did adjust the dialog accordingly.

ntruchsess:
@kiozen : I had to fix a few glitches that were related to the latest refactoring, but now It's done and seems to work as expected. Please do a re-review

kiozen:
Ready to merge?

ntruchsess:
> Ready to merge?

yes, please go ahead

kiozen:
Done! Thanks for the great work!

Issue #623, opened 2023-07-29
Missing/Failing BRouter log window


wthaem:
### Describe the bug
[comment]: # (A clear and concise description of what the bug is)

There is a start/stop icon for the activation of the BRouter offline routing engine. Clicking this icon should start the engine and display `running` in front of the icon. At the same time an additional icon for opening a log window should appear.

In most cases, clicking the start icon changes the above-mentioned text and pops up the log icon only for a very short moment.

### What have you done to circle down the problem?
[comment]: # (Did you compare to previous versions? What versions? Did you check if others have the same problem?)

### To Reproduce
[comment]: # (List the steps to reproduce the behavior)

1. Go to BRouter offline routing
2. Click on start/stop icon
3. See described behavior

### Expected behavior
[comment]: # (A clear and concise description of what you expected to happen)

The new text and the log window icon should be permanently visible.

### Screenshots
[comment]: # (If applicable, add screenshots to help explain your problem)

None

### Attachments
[comment]: # (Can you provide an example file to trigger the problem?)

None

### Tracebacks
[comment]: # (Add your backtrace below if you have one. If QMapshack crashes the fastest way to get help is a backtrace.
For Linux see: https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux.
For Windows it would need a debug build and running QMapshack in Visual Studio.)

None

### Desktop
[comment]: # (please complete the following information)

 - OS: Windows 10, BRouter 1.7.2
 - QMapShack Version:  1.17.0
 - Window Manager: 

### Additional context
[comment]: # (Add any other context about the problem here)

kiozen:
@ntruchsess can you please have a look?

ntruchsess:
> @ntruchsess can you please have a look?
If it's not some kind of windows-specific behaviour this behavior might be caused by a simple error during brouter-startup in combination with a racecondition in the code. As it is not easy to distinguish in between normal shutdown and java terminating due to an initialization-failure of the java-program running within the jvm I used a QTimer that would expire after 200ms and any transition from process-state 'Running' to 'NotRunning' was taken as a normal shutdown. Whenever an initialization-error resulting in termination of the java-process would occur _after_ expiry of this timer the error would be ignored silently. I suppose this is what is happening here.
in #632 I replaced that timer by a simply tracking the stop-action using a boolean flag. Now every termination of java that is not triggered by clicking the start/stop-button will trigger display of the log-window.
[...]

PR #632, opened 2023-08-11
[QMS-623] remove use of QTimer in brouter startup error detection

ntruchsess:

QMS-#623

### What you have done:

to distinguish normal shutdown from errors during startup a timer was used. Any transition to processState::NotRunning that would happen after expiry of this timer would not result in showing the error but would be ignored silently.
The timer has been replaced by a boolean flag 'isBeingStopped' that is set to false on process startup and to true when klicking the stop-button.

1. make sure BRouter is installed locally.
2. validate local BRouter starts fine by clicking the start/stop-button to the right of 'BRouter: stopped'
3. validate 'BRouter: stopped' has changed to 'BRouter: running' and that a second button did appear to the right of the start-button.
4. stop local BRouter by clicking the start/stop-button. Now 'BRouter: stopped' should be displayed again.
5. while QMapShack is still running locate the brouter-jar-file and rename it or move it to a different location.
6. now click the BRouter start/stop button again.
=> the label should switch to 'BRouter: starting' respectivly 'BRouter running' for a very short time and finally show 'BRouter stopped' again. Below the start/stop-button an error-message and a 'Dismiss'-button should be visible.

- [X] yes

- [X] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,5 @@
 [QMS-622] Update BRouter setup (install from github)
+[QMS-623] remove use of QTimer in brouter startup error detection
 [QMS-630] BRouter on-the-fly routing cannot be canceled
 
 V1.17.0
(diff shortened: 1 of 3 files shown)


wthaem:
I merged `ntruchsess/QMS-630` and `ntruchsess/QMS-623` into `kiozen/QMS-624` (I hope correctly). I could see some change in the behavior of the log window but also a crash with the same call stack as before.

kiozen (approved):

ntruchsess:
@kiozen conflict is resolved, PR is ready to merge

Issue #624, opened 2023-07-29
Suspicion: QMS crashes if BRouter RD5 file is missing


wthaem:
### Describe the bug
[comment]: # (A clear and concise description of what the bug is)

***Suspicion:*** QMS crashes when trying to add a new track with BRouter on-the-fly routing with endpoint not in a routing database file (rd5 file).

### What have you done to circle down the problem?
[comment]: # (Did you compare to previous versions? What versions? Did you check if others have the same problem?)

See next point

### To Reproduce
[comment]: # (List the steps to reproduce the behavior)

1. Remove all BRouter `rd5` files
1. Start QMS
2. Activate BRouter offline routing with on-the-fly routing
3. Right click in map window and select `Add track` from context menu
4. Move mouse to next routing point
5. QMS crashes

### Expected behavior
[comment]: # (A clear and concise description of what you expected to happen)

No crash

### Screenshots
[comment]: # (If applicable, add screenshots to help explain your problem)

None

### Attachments
[comment]: # (Can you provide an example file to trigger the problem?)

N.A.

### Tracebacks
[comment]: # (Add your backtrace below if you have one. If QMapshack crashes the fastest way to get help is a backtrace.
For Linux see: https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux.
For Windows it would need a debug build and running QMapshack in Visual Studio.)

None

### Desktop
[comment]: # (please complete the following information)

 - OS: Windows 10, BRouter 1.7.2
 - QMapShack Version:  1.17.0
 - Window Manager: 

### Additional context
[comment]: # (Add any other context about the problem here)

This crash is different from the one caused by rapid mouse move while selecting a routing point located in an existing `rd5` file. The likelihood of that one can be decreased by reducing the mouse speed in the OS mouse settings.

mitxel-m:
I can not reproduce it. 

Without .rd5 files:  
when I click  ( or move) to add the next routing  I get an error, but QMS does not crash.

With .rd5 files: 
I get the same error when I try to add a routing point outside the area covered by the .rd5 files, but no crash.

error:
`Mala respuesta del servidor: Error transferring http://127.0.0.1:17777/brouter?lonlats=-5.954731,36.291186%7C-5.934819,36.286689&profile=car-eco&alternativeidx=0&format=gpx - server replied: Bad Request  `

I am testing on:
 - QMapShack:  1.17.0
[...]

PR #633, opened 2023-08-14
Qms 624

ntruchsess:

QMS-#624

### What you have done:

fixed potential race on CRouterBRouter field 'synchronous'. (It is not propagated to the request-finished slot as a reply property).
fixed potential threading issues mutex not being unlocked by the same thread that did previously acquire the lock before

[comment]: #

BRouter (offline) on-the-fly routing and regular routing should work without crash or freeze.

- [X] yes

- [X] yes

- [X] yes


ntruchsess:
> I'm a bit confused - there are too many PR for me.
this PR is based on kiozen QMS-624, to test there is no need to check out anything else.

wthaem:
What I did:

~~~
git fetch ntruchsess
git checkout -b ntruchsess1 ntruchsess/QMS-624
Switched to a new branch 'ntruchsess1'
branch 'ntruchsess1' set up to track 'ntruchsess/QMS-624'.

git log -3
commit e0399d31a69433b96a36f9ed0de94802219d8327 (HEAD -> ntruchsess1, ntruchsess/QMS-624)
Author: Norbert Truchsess <norbert.truchsess@t-online.de>
Date:   Fri Aug 11 21:50:48 2023 +0200

    QMS-624 fix BRouter mutex threading issues

...
~~~

After that I built QMS as `RelWithDebInfo` and applied the steps described earlier. ***Result:*** Crash. The crash does not yet happen as long as the mouse is still moving. A little while after the mouse comes to a standstill the mouse cursor changes to a rotating circle and again after a little while QMS crashes.

Then I used MSVC for stepwise debugging. Result: With one of the calls to `update()` in `CCanvas::slotTriggerCompleteUpdate` line 804 some Qt modules are called and the crash happens. Before this call is some message and mouse handling.

ntruchsess:
> After that I built QMS as `RelWithDebInfo` and applied the steps described earlier. _**Result:**_ Crash. The crash does not yet happen as long as the mouse is still moving. A little while after the mouse comes to a standstill the mouse cursor changes to a rotating circle and again after a little while QMS crashes.
> 
> Then I used MSVC for stepwise debugging. Result: With one of the calls to `update()` in `CCanvas::slotTriggerCompleteUpdate` line 804 some Qt modules are called and the crash happens. Before this call is some message and mouse handling.

That's an interesting result. I removed the calls to slotTriggerCompleteUpdate(). The caller would update the canvas anyway when receiving the routing-result. (Those calls were copy&paste from the Routino and Mapquest-routers, I never evaluated whether they are actually required. Turns out they are not).

please update and re-check to see whether it makes a difference.

@kiozen : I can see there are quite a lot of direct calls to slotTriggerCompleteUpdate() in the code-base. From my understanding this is only save when all related code is guaranteed to run on the same thread. I suggest that should be changed to utilize Qt-signalling wherever it's unclear that this constraint is fulfilled.

wthaem:
Sorry, no change. Still crashing.

ntruchsess:
> Sorry, no change. Still crashing.

can you please post the call-stack? As relevant parts in the router have changed the resulting call-stack has certainly changed as well and I cannot proceed without.

kiozen:
And use a debug version not relwithdebuginfo.

wthaem:
The call stack is similar or equal to the last one shown in the discussion of issue #624. With the exception of `qmapshack.exe!main(int argc, char * * argv) Zeile 75` all files mentioned are Qt files.

I'll try a debug version in the next few days.

wthaem:
I wasn't aware anymore that building a debug version implies rebuilding all DLLs used in QMS as debug versions. This will take some time.

Building a release with debug info requires only rebuilding the qmapshack.exe - much faster.

Issue #651, opened 2023-12-06
Precision (decimal positions) in displayed distance


AntonioLagoD:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

Hello. I noticed in the track details screen distance is displayed with one decimal position until 20 km and as integer above 20 km. See the image.
![Qmapshack](https://github.com/Maproom/qmapshack/assets/44477003/422f0df3-0f1a-4e59-bffb-1abbb8b85808)

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

Maybe it is posible to increase the precision above 20 km or include an option in the menu 'configuration -> units'. 

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

Maybe in long routes this issue does not matter, but in short routes (walking or cycling) several hundreds of meters can make a big difference in some terrains.

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

mitxel-m:
I would appreciate it too.
Now, both 21.60 Km and 22.40 Km are shown as 22 Km, but there is 800m difference.

The other part comes in the logs and summaries, which shows a mix of distance rounding, which looks strange to me, and in a few cases can lead to inconsistent results

![Summary_metric](https://github.com/Maproom/qmapshack/assets/39693664/cff64811-d061-4917-a74a-c551e8164422)

For now,  distances are shown like this
|range | shown as | example
|---- |-----| --- |
|0-1 Km |  m | 376 m
|1-10 Km | Km rounded to 2 decimals | 1.45 Km
|10-20 Km | Km rounded to 1 decimal | 15.6 Km
| > 20 Km | Km rounded to integer | 22 Km

\* The same applies for  mi / ft

I know this has been discussed before, and I understand that some people want to see long distances rounded, but since the term _'long distance'_  is a very personal measure depending on your activity, I also think it would be useful to have an option in the units configuration UI to choose how to show them. 

However it's not that easy, because at the same time that I want to see distances in km with two decimals I also like to see distances below 1 km in m without decimals.

[...]

PR #657, opened 2023-12-20
[QMS-651] Add configuration option to handle distance rounding

mitxel-m:

QMS-#651 

### What you have done:

Add a box in the units configuration that allows the user to set from which value the rounding of distances will be done for metric and imperial units.

Setup units _001

- The default value is 20 which corresponds to the current behaviour. 
- No configuration is necessary. If the Units config is not touched the current behaviour remains the same by default.
- The modified value is saved in the configuration file for later use. It is named as roundLimit in the "Units" section.

````
[Units]
coordFormat=1
roundLimit=20
slopeMode=1
time\useShortFormat=true
timezone=@ByteArray(Europe/Madrid)
timezone\mode=3
type=0
````

- It is not necessary that this value exists previously in the configuration file, if it does not exist it is set the first time we use the units dialog. This gives compatibility with existing conf files.

- It respects the current concept (which I like) that the larger the distance the less decimal digits are displayed.

It is very simple to use. By default we have the value 20 which corresponds to the current value and gives us this:

|range | shown as | 
|---- |-----| 
|0-1 Km |  m |
|1-10 Km | Km rounded to 2 decimals | 
|10-20 Km | Km rounded to 1 decimal | 
| > 20 Km | Km rounded to integer |

If you set the value to 80 , it would use these ranges

|range | shown as | 
|---- |-----|
|0-1 Km |  m |
|1-40 Km | Km rounded to 2 decimals | 
|40-80 Km | Km rounded to 1 decimal | 
| > 80 Km | Km rounded to integer |

and so on ...

For convenience you can [download  this gpx file](https://www.dropbox.com/scl/fi/q4bz9nxxb7vmtm8kuxzrr/TEST_round.gpx?rlkey=oklmih2961xoxcevy9sbcay1a&dl=1) containing 5 tracks, named with different distances.  Open it in QMS and go to edit the project, so that you see the summary with the data of all 5 tracks in the same view. 

1. First check that you see the distances with the decimals according to the current behaviour.

2. Try with another value: 'Menu - View - Setup Units'  Modify the value. You can use the mouse wheel. Then click on refresh (top right) on the project summary view and check how you see now the distances according to the value you have set. You can repeat it for different values.

3. Check how the distances are displayed with the ruler.

4. You can also change the unit to miles (Imperial) and see how it works.

- [ x] yes

- [ x] yes

- [ x] yes

_Remark: My C++ skills are very limited and I have been able to write this by looking at other parts of the code. Probably my proposal can be refined._

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -35,6 +35,7 @@ V1.XX.X
 [QMS-784] Solved two issues for GNOME/Wayland, Invalid Points Dialog vs Progress Dialog
 [QMS-788] Fix: BRouter (offline) not shown in combobox
 [QMS-791] Error determining BRouter-Web (online) version
+[QMS-793] Auto-routing with BRouter & minor fixes
 [QMS-794] Fix: Links to local files in Waypoint on-screen dialog not opening
 
 V1.17.1
--- a/src/qmapshack/gis/rte/router/CRouterBRouter.cpp
+++ b/src/qmapshack/gis/rte/router/CRouterBRouter.cpp
@@ -82,8 +82,6 @@ CRouterBRouter::CRouterBRouter(QWidget* parent) : IRouter(false, parent) {
 
   cfg.beginGroup("Route/brouter");
   comboProfile->setCurrentIndex(cfg.value("profile", 0).toInt());
-  checkFastRecalc->setChecked(cfg.value("fastRecalc", false).toBool() &&
-                              (setup->installMode == CRouterBRouterSetup::eModeLocal));
   comboAlternative->setCurrentIndex(cfg.value("alternative", 0).toInt());
   cfg.endGroup();
 }
@@ -97,7 +95,6 @@ CRouterBRouter::~CRouterBRouter() {
   cfg.beginGroup("Route/brouter");
   cfg.setValue("profile", comboProfile->currentIndex());
   cfg.setValue("alternative", comboAlternative->currentIndex());
-  cfg.setValue("fastRecalc", checkFastRecalc->isChecked());
   cfg.endGroup();
 }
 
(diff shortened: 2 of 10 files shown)


kiozen:
The PR does not take the differences between imperial and nautical units into account.  

Internally QMS uses and stores metric values. When these values are passed to the UI or read from the UI you need to use one of the methods defined by the instance you get by calling IUnit::self(). This will also provide you with the correct unit string to be displayed.  

Switching between the metric, imperial and nautical system must be a coherent user experience.

mitxel-m:
Ok, I will modify it.
The first proposal focused on making as few changes as possible in terms of usage.
That's why it only rounded distances for Metric and Imperial units, and kept them unrounded for Nautic and Aviation, exactly as is currently done in version 1.17.1

But I agree that to be coherent once the user gets the option to round distances, it should be applicable whatever units type he/she chooses.

So I'm planning to do :
- make the rounding option applicable to all units types.
- correctly display the unit string appended to the value in the UI box, according to the clicked option.

Setup units _002

would this meet the requirements?

Apart from that, and thanks to @kiozen 's previous comment, I have realised that this same thing is wrong in the UI of the speed filter for the tracks. It doesn't show the correct unit suffix, but always shows km/h even if you use mi/h, which leads to confusion.

I think I will be able to fix it, but I would like to do it in a different issue once we finish this one.

mitxel-m:
I think it is now ready to review with commit [1634499]

kiozen:
I hope I find some time to work on this soon. Storing 20000m as 20 and applying all these hardcoded factors is hell to maintain. Imho the IUnit API should provide enough methods to handle this properly. However I can not sketch a simple solution for that. Therefore I need to find the time to create a patch. 

I still doubt that this does make much sense with given resolution and precision of maps and GPS devices.

Issue #694, opened 2024-11-06
Porting Qt6: Fix decoding of device mount point


kiozen:
### The parent ticket
[comment]: # (Add ticketnumber behind QMS-# )

QMS-#648

### What do you plan to do
[comment]: # (Describe roughly what you want to do.)

Decoding the mount point for GPS devices does not work properly due to a trailing 0.

### The result will be
[comment]: # (A list of goals that have to be achieved to consider the ticket as done.)

Garmin devices should be detected on Linux and Windows again.

PR #696, opened 2024-11-06
[QMS-694] Porting Qt6: Fix decoding of device mount point

kiozen:

QMS-#694

### What you have done:

Refactured code to decode DBus messages properly

Check if a Garmin device can be mounted and read

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/device/CDeviceWatcherLinux.cpp
+++ b/src/qmapshack/device/CDeviceWatcherLinux.cpp
@@ -146,7 +146,6 @@ void CDeviceWatcherLinux::slotUpdate() {
 }
 
 QString CDeviceWatcherLinux::readMountPoint(const QString& path) {
-  QStringList points;
   QDBusMessage message =
       QDBusMessage::createMethodCall("org.freedesktop.UDisks2", path, "org.freedesktop.DBus.Properties", "Get");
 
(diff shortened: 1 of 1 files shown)


src/qmapshack/device/CDeviceWatcherLinux.cpp
```diff
@@ -157,30 +156,54 @@ QString CDeviceWatcherLinux::readMountPoint(const QString& path) {
...
+    dbusArg.beginArray();
+    while (!dbusArg.atEnd()) {
+      QByteArray data;
+      dbusArg >> data;
+      // Not sure about this one. But my guts feeling tells me that
+      // we should decode the stuff according to the system's requirements.
+      auto decoder = QStringDecoder(QStringDecoder::System);
+      // As QByteArray has a trailing 0 however QString does not it needs to be
+      // removed. Else the QString has a 0 as unicode character, confusing the
+      // rest of the system.
```
frankystone, line 198: This sounds like the `0` is added by QByteArray, but it's the last entry in the array coming from the DBus.

frankystone (commented):
Works!

Thanks for spending your time on this.

src/qmapshack/device/CDeviceWatcherLinux.cpp
kiozen: According to the docs QByteArray always enforces a trailing 0 . If there is one in the raw data (DBus message) it's ok. If not it is appended by QByteArray. So it'S always guaranteed to be 0 terminated. 

The part I do not understand is why QString does not recognize it as 0 terminated string and takes over the 0 as part of the string.

frankystone: From my understanding QByteArray represents just a number of pure bytes, whereas QString is always UTF16 encoded. From [Using QByteArray](https://wiki.qt.io/Using_QByteArray):

> To convert a QByteArray to QString, one needs to specify the encoding. It is not possible to convert a QByteArray to QString without knowing the encoding. Once the encoding is known, one can use functions like QString::fromUtf8, QString::fromUcs4 or use QTextCodec::toUnicode().

But i have no experience in QT.

kiozen: Yes decoding is another caveat when converting from QByteArray to QString. That is why I introduced the decoder. Anyways this does not explain why the trailing 0 is converted to an explicit "\0" as part of the string.

frankystone: You should ask this in a QT forum…

In python it's the same, i used the int's from my qdbus output:
~~~~
>>> message = [47, 114, 117, 110, 47, 109, 101, 100, 105, 97, 47, 107, 97, 112, 117, 116, 116, 110, 105, 107, 47, 75, 65, 82, 84, 69, 32, 54, 48, 71, 66, 0]
>>> ar = bytearray(message)
>>> print(ar)
bytearray(b'/run/media/kaputtnik/KARTE 60GB\x00')
>>> s = str(ar)
>>> print(s)
bytearray(b'/run/media/kaputtnik/KARTE 60GB\x00')
>>> print(ar.decode())
/run/media/kaputtnik/KARTE 60GB
>>> s = str(ar, encoding='utf-8')
>>> print(s)
/run/media/kaputtnik/KARTE 60GB
~~~~

kiozen: I think the correct way to access the byte array is to use the `const char*` pointer. By that decoding can detect the null termination properly.

kiozen:
@wthaem does this work for Windows, too?

kkarsten62:
I compiled kiozen qmapshack fork with branch [QMS-694] against `| Fedora 41 | Wayland | Gnome 47 | Qt 6.8.0`.

**Work as expected.**

I can read/write from/to my GARMIN Edge 1000 device.

wthaem:
@kiozen: I'll test it in the next few days.

kiozen:
@wthaem I think it's enough to test the porting branch on Windows in general. Not necessarily this branch. I just want to make sure we do not have similar issues in the Windows code.

wthaem:
@kiozen: I mentioned in #648 a failing string compare issue in `CMapRMAP.cpp`. There, an extra zero byte played a role. I think that similar zero byte, QString, QArray, or encoding issues can be hidden elsewhere in the code, too.

kiozen:
> @kiozen: I mentioned in #648 a failing string compare issue in `CMapRMAP.cpp`. There, an extra zero byte played a role. I think that similar zero byte, QString, QArray, or encoding issues can be hidden elsewhere in the code, too.

Ok, can you check if `"CompeGPSRasterImage" != QString(charbuf.data())` does the trick.

wthaem:
Tested QMS with Qt6 on Windows 10:

@kiozen:

* `CMapRMAP.cpp` - `"CompeGPSRasterImage" != QString(charbuf.data())`:  Could load rmap maps. Seems to work. How to identify similar situations elsewhere in the code?
* I connected my Garmin GPSMap 64s to QMS. Works in read and write modes without any problems.
* Currently, I use the QMS Qt6 version for my normal work with QMS. Thus, more or less frequent testing of the features which are used in my workflow.

kiozen:
I merge this one and open a new one to collect all other QByteArray fixes

kiozen:
That is the new ticket https://github.com/Maproom/qmapshack/issues/697

Issue #697, opened 2024-11-12
Check all usage of QByteArray for bad decoding


kiozen:
### The parent ticket
[comment]: # (Add ticketnumber behind QMS-# )

QMS-#648

### What do you plan to do
[comment]: # (Describe roughly what you want to do.)

Grep for QByteArray and check if the API is used correctly

### The result will be
[comment]: # (A list of goals that have to be achieved to consider the ticket as done.)

Hopefully no strings with `\0` terminator as part of the literal string.

PR #698, opened 2024-11-12
[QMS-697]  Fix QByteArray to QString decoding

kiozen:

QMS-#697

### What you have done:

Search for QByteArray and use QByteArray::data() whenever a string conversion is done.

There is no real bad case except in loading RMap.  But there is a list of functions that should be tested as good as possible:

* all time zone related stuff
* RMap support 
* BRouter setup and support
* Realtime data all modules.

- [x] yes

- [x] yes


frankystone:
Tested on archlinux with Qt 6.8:

TimeZones:
* Changed timezone from Local to UTC and vice versa, each time restarted QMS. After changing TZ a track calculated with the speed filter shows all times correct
* Also the tool tip window when the mouse pointer hovers an activated track shows all data depending on TimeZone
* Applying the filter Speed with UTC and Local Timezone

BRouter
* Installed a new profile
* Updated all data (11 tiles)
* Using BRouter with different profiles for Tracks

No issues found so far.

I can't test RealTime Data because i do not know how it works and because i don't want to enable GPS on my Smartphone.

Anything special to test further?

PR #701, opened 2024-11-30
Workaround for bad app styles

frankystone:

QMS-#https://github.com/Maproom/qmapshack/issues/648#issuecomment-2475741222

Since with qt5 the single column menu can be scrolled, i think this should go into porting_qt6 (if accepted).

### What you have done:
Compare screen height with preferred height of a QMenu. If the preferred height is bigger than the screen height add a qt default style to the menu. This forces qt to create a multi column menu.

This solution may not work with multi monitor setups. Not tested if the orientation of a monitor is set to portrait.

I took "fusion" as the default style because "windows" doesn't look good.

1. On KDE use "breeze" as application design
2. Try to change a waypoint icon

- [x] yes

- [x] yes

I guess the changelog will be changed if the porting_qt6 branch get merged into master.


kiozen:
Imho this will cause problems. As you already mentioned this won't work on a multi-screen setup with different sized screens if QMS is run on any other screen than the primary. 

I still wonder why on some systems this does not work. My KDE systems are default Breeze style. Maybe it's better to check and setup the style on application level. Similar to what you have described in the ticket. But within QMS.

frankystone:
> I still wonder why on some systems this does not work. My KDE systems are default Breeze style.
Can you please provide the used version? Here it is 6.2.4:
~~~~
$:> pacman -Qs breeze
local/breeze 6.2.4-1 (plasma)
    Artwork, styles and assets for the Breeze visual style for the Plasma Desktop
local/breeze-gtk 6.2.4-1 (plasma)
    Breeze widget theme for GTK 2 and 3
local/breeze-icons 6.8.0-1 (kf6)
    Breeze icon theme
local/qqc2-breeze-style 6.2.4-1 (plasma)
    Applications useful for Plasma development
~~~~

> Maybe it's better to check and setup the style on application level. Similar to what you have described in the ticket. But within QMS.

Not sure what you are referring. Can you explain?

kiozen:
```
sudo apt list *breeze* | grep installed

breeze-cursor-theme/noble,noble,noble,now 4:6.2.3-0zneon+24.04+noble+release+build11 all [installed]
breeze-gtk-theme/noble,now 6.2.3-0zneon+24.04+noble+release+build8 amd64 [installed]
breeze-icon-theme/now 4:5.116.0-0xneon+22.04+jammy+release+build31 all [installed,local]
breeze5/noble,now 4:6.2.3-0zneon+24.04+noble+release+build11 amd64 [installed]
breeze/noble,now 4:6.2.3-0zneon+24.04+noble+release+build11 amd64 [installed]
kde-style-breeze/noble,now 4:6.2.3-0zneon+24.04+noble+release+build11 amd64 [installed]
kf6-breeze-icon-theme/noble,now 6.8.0-0zneon+24.04+noble+release+build6 amd64 [installed]
kwin-style-breeze/noble,now 4:6.2.3-0zneon+24.04+noble+release+build11 amd64 [installed]
libreoffice-style-breeze/noble-updates,noble-updates,now 4:24.2.6-0ubuntu0.24.04.1 all [installed]
qml6-module-org-kde-qqc2breezestyle/noble,now 6.2.3-0zneon+24.04+noble+release+build8 amd64 [installed]

```

frankystone:
I've changed the code to get the screen from where the parent widget is located. See [windowHandle()](https://doc.qt.io/qt-6/qwidget.html#windowHandle)

I don't know how to dig deeper in the codebase of breeze and QWidget to get an idea why this might happens. I wrote a bug report at [kde.bugs.org](https://bugs.kde.org/show_bug.cgi?id=496930)
I guess from somewhere  [updateActionRects()](https://github.com/qt/qtbase/blob/f401142d321cfa9de5e1632f7f66a6b5c02525c1/src/widgets/widgets/qmenu.cpp#L338)  will be called, but i don't know from where. At line 437 the function `style->sizeFromContents()` is called, the reason why i came up with the idea to change the style. Breeze has this function but provide no spezial calculation for a QMenu, see [sizeFromContents ff](https://github.com/KDE/breeze/blob/f6f0c1d89a44d3ed210e533bfc56b7b90542fc4a/kstyle/breezestyle.cpp#L1086).

The only thing we may can compare are the style settings. The output of `qdbus6 org.kde.KWin /KWin org.kde.KWin.supportInformation` contains a section called "Decoration" which seems to provide this settings:

~~~~
Decoration
==========
Plugin: org.kde.breeze
Theme: Breeze
Plugin recommends border size: None
onAllDesktopsAvailable: false
alphaChannelSupported: true
closeOnDoubleClickOnMenu: false
decorationButtonsLeft: 0, 2
decorationButtonsRight: 6, 3, 4, 5
borderSize: 3
gridUnit: 10
font: Noto Sans,10,-1,0,400,0,0,0,0,0,0,0,0,0,0,1
smallSpacing: 2
largeSpacing: 10
~~~~

Those should be the default ones on my machine because i reset them all to default.

src/qmapshack/helpers/CWptIconManager.cpp
```diff
@@ -314,5 +314,21 @@ QMenu* CWptIconManager::getWptIconMenu(const QString& title, QObject* obj, const
     }
   }
 
+  // Workaround for https://bugs.kde.org/show_bug.cgi?id=496930.
+  // Apply a Qt default style in this case
+  QRect screenGeometry = parent->windowHandle()->screen()->geometry();
```
kiozen, line 319: const QRect& screenGeomettry ...

To avoid needles copies.

src/qmapshack/helpers/CWptIconManager.cpp
```diff
@@ -314,5 +314,21 @@ QMenu* CWptIconManager::getWptIconMenu(const QString& title, QObject* obj, const
...
+  QSize menuPreferredSize = menu->sizeHint();
```
kiozen, line 320: dito

src/qmapshack/helpers/CWptIconManager.cpp
```diff
@@ -314,5 +314,21 @@ QMenu* CWptIconManager::getWptIconMenu(const QString& title, QObject* obj, const
...
+
+  if (menuPreferredSize.height() > screenGeometry.height()) {
+    qDebug() << "Found a misbehaving style: " << QApplication::style()->objectName();
+
+    QStyle* style = QStyleFactory::create("fusion");
+
+    if (style != nullptr){
+      qDebug() << "Applying a default style for widget QMenu";
+      menu->setStyle(style);
```
kiozen, line 329: setStyle does not transfer ownership to QMenu.  In other words `style` is created, used by the QMenu object but never destroyed. This is a memory leak.

https://doc.qt.io/qt-6/qwidget.html#setStyle

I still have a bad feeling about this hack. The documentation clearly states to avoid it for normal use. And if I have learned something from Qt documentation it's to listen to their warnings.

kiozen (changes requested):
Maybe setting the style fro the application is the better approach:

https://doc.qt.io/qt-6/qapplication.html#setStyle

However in this case QMS will not obey the style of the used window manager. We can count to 3 until someone complains.

frankystone:
> Maybe setting the style fro the application is the better approach:
> 
> https://doc.qt.io/qt-6/qapplication.html#setStyle
> 
> However in this case QMS will not obey the style of the used window manager. We can count to 3 until someone complains.

Hm, personally i don't like this. This issue affects only one little part of QMS (afaik) and only some users. Asking a question at startup for an issue which the user may do not understand and he may wouldn't be affected at all is a bit overkill, imho.

src/qmapshack/helpers/CWptIconManager.cpp
frankystone: Would it be sufficient to add `delete style;` directly after `menu->setStyle(style)` to prevent the memory leak? I tested this and it works. If this is sufficient i will change also the other things you mentioned.

Anyway, if you have bad feelings we (i) should find another solution. The warning is surely there on purpose… but i guess it just states "Use a konsistent style for your application!".

kiozen: The problem with the style instance is that is it used by the menu instance. And this is passed as return value for further use. If you delete the style object, the pointer stored in the menu object points to invalid memory. Linux is pretty sloppy if it comes to that situation. As long as the freed memory is not corrupted by another allocation it works. So this is a ticking time bomb. Windows is more strict on that and will crash. On Linux you need to compile with ASAN to catch such stuff.

"Use a konsistent style for your application!" or "This is just some internal hack and we change that for any reason. Use at your own risk".

I still wonder why this is broken on some systems. Do you use the Qt version shipped with the distribution or the one provided by Qt via their online install tool?

kiozen:
> Hm, personally i don't like this. This issue affects only one little part of QMS (afaik) and only some users. Asking a question at startup for an issue which the user may do not understand and he may wouldn't be affected at all is a bit overkill, imho

Totally agree on that.

src/qmapshack/helpers/CWptIconManager.cpp
frankystone: Thanks for the detailed explanation.
After reading about this here and there i found `setParent()` [see here the last chapter](https://doc.qt.io/qt-6/qproxystyle.html#details) which should solve this, imho:

~~~~
    QStyle* style = QStyleFactory::create("fusion");
    style->setParent(menu);

    if (style != nullptr){
      menu->setStyle(style);
    }
~~~~

Using a proxy style might be another solution, but it is not guaranteed that it works. And digging through the QStyle documentation produces a lot of question marks in my brain. I even do not know how find out if the menu has e.g. a scroll option nor i have an idea how to add a scroll option. But this would be the most sufficient solution imho: Check if the menu have multiple columns (or a scroll option), and adjust it if not.

> I still wonder why this is broken on some systems. Do you use the Qt version shipped with the distribution or the one provided by Qt via their online install tool?

I used the version from archlinux. As far i can see they apply only two compiler related [patches](https://gitlab.archlinux.org/archlinux/packaging/packages/qt6-base).  Maybe i will try to use the online installation tool, but i fear this will mess up my installed packages, or the package manager database.

kiozen:
I am still very reluctant to merge this hack. It works around some issue with some window manager. Just from over 30 years of experience - I always regretted such stuff in the past.

frankystone:
ok, i am fine with that. In the meantime i have created a simple QT program which shows the same behavior for me and added it to the [kde bugreport](https://bugs.kde.org/show_bug.cgi?id=496930) .

Hopefully some one over there will fix this. Until then i have to use the QT `-style` option like shown [here](https://github.com/Maproom/qmapshack/issues/648#issuecomment-2508918737).

PR #720, opened 2025-04-03
deal with translations on qt6 branch

henning-schild:

QMS-#

### What you have done:

1. Open whatever
2. Click here

- [ ] yes

- [ ] yes


kiozen:
If you change a string a new translation entry is done. Keeping the old version of the translated string makes sense as you want to see it while you translate the new version. However over time this adds a lot of garbage to the files. This is the reason why we made it a switch.  From time to time I switched on the removal to clean up the files.

I like the idea of built-in translations. Imho this should be the way to go for Linux, MacOS and Windows.

henning-schild:
I think for embedding the code needs to change as well. Look for a ressource "i18n" and not open a file etc. For cmake you just do not set `QM_FILES_OUTPUT_VARIABLE` in `qt_add_translations` and drop the `install *_QM_FILES`.

Have WIN and APPLE ever been translated? I am asking because maybe that alignment can be delayed and written down in an issue. My guess is that it has nothing to do with the cmake/qt update.

Anyhow, please take it from here. I will not be available for 2 weeks.

wthaem:
If any possible: it would be quite reasonable to have a chance to easily change the GUI language of QMS to a language different to the OS language. With qm-files it is very simple. How would this be done with built-in translations? Example: When preparing the translations there is a need to run the English version, but also a version with the target language not equal to de.de while the OS speaks German.

henning-schild:
Nothing would change. It is really just a matter of filesystem layout. Embedding is default, and external would allow distros/users to strip off files to save space.

And looking at msvc_64/ and MacOSX/bundle* the two have been supported and did use non-embedded. So all there is to do here is to make sure those two still work.

Any other discussion about using embedding in the future it probably off topic. Sorry for the confusion i caused because i did not see anything in the cmake files.

Issue #738, opened 2025-05-03
Use Gamin's FIT SDK to support FIT files properly


kiozen:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

The FIT implementation of QMapShack is quite old and hardly maintained. It causes a lot of warning messages when loading FIT files. 

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

As Garmin seems to be pretty relaxed with their SDK I like to add this to the repository (Garmin does not provide one for c++) and use this for FIT file decoding

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

n/a

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

The hope is to have a more robust solution, that needs less maintenance effort

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

nuqleo:
Does this create any issues with the program’s license complying with the requirements for free/open licenses?

kiozen:
🤷  

Read it and decide on your own

https://developer.garmin.com/fit/download/

PR #742, opened 2025-05-08
QMS-738 Use Gamin's FIT SDK to support FIT files properly

kiozen:

QMS-#738

### What you have done:

* Add Garmin's FIT lib as 3rd party code
* Create a new FIT project class to replace the old one

The new class is used to load FIT projects form disk and from MTP devices. The old one is still used to load FIT files from Garmin devices in mass storage mode.

Additionally I updated translations and fixed the CMakeLists.txt for translations.

Just load your fit file

- [x] yes

- [x] yes


kiozen:
@all if you have FIT files that are more than simple activity recordings I would like to have them for testing. Everything with sensor data and laps, courses, etc is welcome.

rootrootde:
I recently got a Garmin Edge Explore 2 - these devices can save "favorite waypoints", stored in a "Locations" fit file (device://Garmin/Locations/Locations.fit)
It would be nice to read/write those Locations.fit files, I think.
to write my own locations.fit files I recently looked a bit into the FIT specifications and found that "Locations" is undocumented in the SDK - among many others

A Locations.fit file fresh from my device (cant upload unzipped):
[Locations.fit.zip](https://github.com/user-attachments/files/20105289/Locations.fit.zip)

Here is some information on undocumented FIT messages in the SDK that I happened to come across, maybe you find it useful:

[FIT Undocumented.ods](https://github.com/user-attachments/files/20105173/FIT.Undocumented.ods)
[FIT Undocumented.xlsx](https://github.com/user-attachments/files/20105175/FIT.Undocumented.xlsx) (for users of inferior OS...)

kkarsten62:
> @ALL if you have FIT files that are more than simple activity recordings I would like to have them for testing. Everything with sensor data and laps, courses, etc is welcome.

Hi @kiozen ,
many thanks for your great engagement in the last week. I can mount all my sensors on my trekking bike and will do a live trekking tour next week. FIT file will then have datas from sensors: laps, powermeter, cadence, temperate, heart rate, speed.
I hope you can wait so long.

kiozen:
Ah, here we have a real feature creep :grin:  Yes this is very much appreciated.  Best we solve this in a separate ticket.

kiozen:
@rootrootde Just made quick test. The SDK does not even know the messages stored in that file (so undocumented and unknown). This is a bit of a problem as the idea to switch to the SDK is to to avoid hacking the FIT protocol on our own.

kkarsten62:
> I recently got a Garmin Edge Explore 2 - these devices can save "favorite waypoints", stored in a "Locations" fit file (device://Garmin/Locations/Locations.fit) It would be nice to read/write those Locations.fit files, I think. to write my own locations.fit files I recently looked a bit into the FIT specifications and found that "Locations" is undocumented in the SDK - among many others
> ...

I am looking since years to modify and delete waypoints from QMS in my connected GARMIN Edge 1000 (per USB mass stoarage) - but I found no solution so far.

Here are my observations:
- During connection procedure it seems for me that the Edge 1000 will create always a new `Locations.fit` file in the folder `/.../GARMIN/Garmin/Locations`
- Deleting or renaming this file on the connected Edge 1000 via a file browser has a no impact, when reconnecting the Edge 1000 the files will be there again. For me it seems that the waypoints informations are stored somewhere else on the Edget 1000 and the `Locations.fit` is only a copy or a kind of extract
- In QMS I can read the waypoints  and can copy them to another project
grafik

- I can delete waypoints in the connected `Locations` project but `Refreshing the project on the device` will have no impact
grafik

- After reconnecting the Edge 1000 the deleted waypoints are still existing
- So the only solution for me is to delete the waypoints directly in the Edge 1000 - but this very boring, because you need 10-20 fingertips for one waypoints

kiozen:
I don't think we can change that behavior with QMapShack. This is how a Garmin works. It's not comparable to a smartphone or PC. The device has it's own internal memory for all data. While booting some information from the user accessible flash is copied into the internal structures. When switching to USB mode the internal data is dumped to the user accessible flash. 

In such a scenario you need strict rules on how to change/delete data. Doing this via the device is in my opinion a good practice. For recent devices you can use the apps on a smartphone to delete stuff via Bluetooth. But again: Using Bluetooth the app tells the device to forget the internal data. It does not edit or change the data on the accessible flash.

I would be nice to have this Bluetooth access, too. But as far as I can see this is pure proprietary API and a pain to reverse engineer.

rootrootde:
@kiozen I see, too bad, thanks for checking anyways!

> I am looking since years to modify and delete waypoints from QMS in my connected GARMIN Edge 1000 (per USB mass stoarage) - but I found no solution so far.

My device uses MTP only - maybe that's a difference.
I managed to implement the missing messages for Locations.fit in the ([python_fit_tool](https://bitbucket.org/stagescycling/python_fit_tool/src/main/) and am now able to write Locations.fit files which my device processes properly.
There is a dedicated settings message you can include in the locations.fit to a)add waypoints to device, b) replace waypoints on device or c) delete all wpts on device
I simply replace the Locations.fit on the device with my own, seems to work pretty well. Deleting/Replacing all wpts already saved me a lot of time.

you can reach me at "locationsfit at rootroot punkt de" if you want to try it, let's not hijack this thread ;)

PR #768, opened 2025-05-31
Improve windows deployment

OgreTransporter:

Nothing

### What you have done:

I have activated compiling on multiple cores for the MSVC and added the deployment scripts for the installation under Windows from Qt to copy the required DLLs.


OgreTransporter:
I know that there is a script msvc_64/copyfiles.bat. However, this has hard-coded paths and will therefore probably not work on any system without adjustments. In addition, Qt5 is set there. It also says ‘Note: Qt5WebEngine deployment is super crazy’. Copying/deploying the Qt files is very annoying. The easiest way is therefore to use the windeployment provided by Qt or to integrate the CMake script into the install routine. The programme copies all necessary files and their dependencies. Calling it for all programmes one after the other ensures that all files are present in the output. Yes, GDAL, etc. are still missing. But that's not so easy, because the dependencies can have different names depending on the source. For example, `proj_8_?.dll` does not work for me because I use PROJ 9. And I don't have a file `zlibwapi.dll` either, because I have linked to the static version of zlib. And I do not have `libwinpthread-1.dll` either, because I have built everything natively with MSVC and do not use MinGW. A code similar to https://github.com/OgreTransporter/CMake-Helpers/blob/master/ImportedDlls.cmake can be used for the dependency packages. If you have better suggestions, feel free to post them in a new PR.

Issue #812, opened 2025-07-12
Set user defined sizes for waypoint and POI icons


JFritzle:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

Depending on screen resolution, built-in waypoint and POI icon size of 22 pixels may be inappropriate, either too large or too small.

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

It would be desirable if users could set the icon size according to their needs.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

I am the one who invests time in this request, which I believe is beneficial for others as well.

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

**What I have achieved so far:**

1. Added new toolbar element _Setup Map Icon Sizes_ with corresponding symbol:
        
     ![Image](https://github.com/user-attachments/assets/860e312b-5e22-49fa-a705-db9bf0047777)
     
     ![Image](https://github.com/user-attachments/assets/f9c88808-14b8-4733-9042-bf3c193298ac)

2. Designed corresponding dialog widget allowing to set different sizes for _waypoint icons_ and _POI icons_ within range from 16 pixels to 48 pixels:
    
   ![Image](https://github.com/user-attachments/assets/eb50ef70-81b9-4d37-b902-bf1fa6bc5127)

   _Restore Defaults_ button sets sizes do previously built-in size of 22 pixels.
   Size values are stored in and restored from QMS configuration.
   Upper and lower range limits are fix. Should range of allowed size values be designed larger?

3. Changed _POI icon_ size is applied immediately to visible POIs after pressing OK button.
4. Changed _POI icon_ size is also applied to POIs shown after restarting QMS.
5. Size of highlight shown around POI when mouse pointer hovers over POI is recalculated based on changed POI icon size and is applied immediately.
[...]

PR #814, opened 2025-07-14
[QMS-812] User defined waypoint and POI icon sizes

JFritzle:

QMS-#812

### What you have done:

- Enable user defined sizes for waypoint and POI icons
- Apply changed sizes immediately to visible map elements
- Add German translations for icon size dialog
- In addition update a few more German translations
- Apply changed map font immediately to visible map elements

1. Create waypoint and/or POI icons on map
2. Open new _Setup Map Icon Sizes_ dialog, change icon sizes and hit "OK" button
3. Verify that waypoint and/or POI icons sizes are applied immediately to visible map icons
4. Verify that icon highlight shown when hovering over icons uses changed sizes too
5. Verify that moving waypoints uses new waypoint changed too
6. Verify that icon's text label location suits to changed sizes
7. Verify that hit "Restore Defaults" button sets both sizes to default size 22 pixels
8. Verify that changed sizes are restored and re-applied after QMS restart
9. Verify that _Setup Map Icon Sizes_ can be added to toolbar
10. Open _Setup Map Font_ dialog, change font name and/or font size and hit "Select" button
11. Verify that new font attributes are applied immediately to visible map labels

- [x] yes

- [x] yes


src/qmapshack/icon/CMapIconSizesSetup.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <QPushButton>
+#include <QSlider>
+
+CMapIconSizesSetup::CMapIconSizesSetup(QWidget* parent) : QDialog(parent) {
+  setupUi(this);
+
+  SETTINGS;
+  slider_1->valueChanged(cfg.value("Icons/sizeWpt", 22).toInt());
+  slider_2->valueChanged(cfg.value("Icons/sizePoi", 22).toInt());
```
kiozen, line 32: I always try to use readable names like sliderWpt or sliderPoi

And valueChanged is  a signal. Either use emit or this is the wrong API to use.

src/qmapshack/icon/CMapIconSizesSetup.h
```diff
@@ -0,0 +1,43 @@
+/**********************************************************************************************
```
kiozen, line 1: Either everything in "helpers" or all dialogs and stuff  concerning icons in "icons"

src/qmapshack/icon/CMapIconSizesSetup.cpp
```diff
@@ -0,0 +1,64 @@
...
+    emit sigSetIconSizePoi(size);
+    done = QDialog::Accepted;
+  }
+  QDialog::done(done);
+}
+
+void CMapIconSizesSetup::slotResetClicked(bool checked) {
+  slider_1->valueChanged(22);
+  slider_2->valueChanged(22);
+}
```
kiozen, line 64: valueChanged is a signal!

src/qmapshack/poi/IPoiFile.cpp
```diff
@@ -47,3 +44,19 @@ IPoiProp* IPoiFile::getSetup() {
...
+  return _iconHighlight;
+}
+
+void IPoiFile::slotSetIconSize(int size) {
+  _iconSize = size;
+   qreal scale = size / 22.0;
+   QImage icon = QImage("://cursors/poiHighlightRed.png");
+  _iconHighlight = icon.scaled(icon.size() * scale, Qt::KeepAspectRatio, Qt::SmoothTransformation);
+}
+
```
kiozen, line 62: same reasoning as with the Wpts

kiozen (changes requested):
Just an idea: Would be nice to have either example icons in the dialog to change size. Or change the icons on the map on-the-fly.  By that a user has direct feedback.

kiozen:
Small observation from testing: Icon is not fitting the circle when multiple items are available for selection.

This branch:
grafik

dev branch:
grafik

JFritzle:
_Tried_ and hopefully _made_ all requested changes.

About:
_Small observation from testing: Icon is not fitting the circle when multiple items are available for selection._ 
Had to redesign workflow off waypoint resizing totally. At first, _scaled_ waypoint icons were requested from WaypointIconManager and stored in GIS data structure. This could result in the icons being too large when unscaled ones were needed. Now, _scaled_ waypoint icons are only requested on-the-fly just before drawing them and are only used for drawing.
It was not sufficient to rescale icons already stored in GIS data structure. _Unscaled_ waypoint icons are downscaled from original image data to 22 pixels. Upscaling such a downscaled image leads to blurry icons where details are hardly to recognize, because pixel accuracy of original image gets lost.

About:
_Just an idea: Would be nice to have either example icons in the dialog to change size. Or change the icons on the map on-the-fly. By that a user has direct feedback._
Good idea. Suggestion is implemented.

Recent screenshot:
Screenshot 2025-07-17 13-41-47

src/qmapshack/helpers/CMapIconSizesSetup.h
```diff
@@ -0,0 +1,49 @@
...
+  void accept() override;
+
+ private slots:
+  void slotSetIconSizeWpt(int size);
+  void slotSetIconSizePoi(int size);
+  void slotResetClicked(bool checked);
+
+ private:
+  static QImage imageWpt;
+  static QImage imagePoi;
```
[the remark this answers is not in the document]
JFritzle: Please excuse any lack of _C++_ expertise, as I have only programmed in _C_ and _Fortran_ almost all my life and and acquired my  _C++_ skills only by doing.
 
My goal was to initialize `imageWpt` and `imagePoi` only once, as their value always stays the same. That is why I defined them "static". I tried to initialize them like `QImage CMapIconSizesSetup::imageWpt = QImage("://icons/48x48/Waypoint.png");` outside constructor. Without success. According to [this discussion](https://stackoverflow.com/questions/64861625/why-declaring-a-qimage-as-static-const-produces-an-invalid-value), this works for images from file system only but not for images from resources. No compiler error, but variables point to null images. So initialization moved from outside to inside constructor. "static" left over. Any better idea? 

I now removed "static" where possible.

Concerning class `IPoiFile`, I took what I found and only modified it slightly.

kiozen:
I suggest a few code changes. See attached patch

[patch.txt](https://github.com/user-attachments/files/21313283/patch.txt)

src/qmapshack/helpers/CMapIconSizesSetup.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include "helpers/CSettings.h"
+#include "helpers/CWptIconManager.h"
+#include "poi/IPoiFile.h"
+
+CMapIconSizesSetup::CMapIconSizesSetup(QWidget* parent)
+    : QDialog(parent), imageWpt("://icons/48x48/Waypoint.png"), imagePoi("://icons/48x48/Poi.png") {
+  setupUi(this);
+
+  int labelSize = qMax(sliderWpt->maximum(), sliderPoi->maximum());
```
kiozen, line 32: Just a minor detail: The label is a styled panel with a frame of size 1px. Therefore the inner area to display stuff is 46x46 instead of 48x48. This is why the icons seem to get a bit larger than the panel. Therefor labelSize should be +2 whatever qMax passes back.

src/qmapshack/helpers/CMapIconSizesSetup.cpp
```diff
@@ -0,0 +1,83 @@
+/**********************************************************************************************
+    Copyright (C) 2014 Oliver Eichler <oliver.eichler@gmx.de>
```
kiozen, line 2: Just noticed: Not my work & copyright :wink:  If you enter your name you will get part of the credits (which is well deserved by now)

src/qmapshack/helpers/CMapIconSizesSetup.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  int labelSize = qMax(sliderWpt->maximum(), sliderPoi->maximum());
+  labelWpt->setFixedSize(labelSize, labelSize);
+  labelPoi->setFixedSize(labelSize, labelSize);
+
+  connect(sliderWpt, &QSlider::valueChanged, this, &CMapIconSizesSetup::slotSetIconSizeWpt);
+  connect(sliderPoi, &QSlider::valueChanged, this, &CMapIconSizesSetup::slotSetIconSizePoi);
+
+  SETTINGS;
+  sliderWpt->setValue(cfg.value("Icons/sizeWpt", 22).toInt());
```
kiozen, line 40: We have this 22 as magic value everywhere in the code now.  This is screaming for trouble if we want to change for 20 or 24. This should be a constexpr or a define. Best together with any other magic value in this patch.

The icon manager  header should be a good place.

kiozen:
Thanks for the work!

Issue #850, opened 2025-10-22
Add custom icons from external icons folder and subfolders


JFritzle:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )
Currently, waypoint icon selection gets redesigned by enhancement [#846](https://github.com/Maproom/qmapshack/issues/846), allowing comfortably to select between a large amount of built-in as well as custom icons. Therefore, adding more and more custom icons becomes interesting for users.

The limitation of having a large number of custom icon files in one single icon folder is not practical.

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)
Icons should be read not only from the specified external icon folder, but recursively from all subfolders of the external icon folder, allowing users to organize and group their custom icons.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

I will remove this limitation by iterating through external icons folder and it's subfolders.

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

As enhancement [#846](https://github.com/Maproom/qmapshack/issues/846) will introduce categories and/or tags for waypoint icons, subfolder names could help to categorize custom icons.

PR #853, opened 2025-10-22
[QMS-850] Add custom icons from external icons folder and subfolders

JFritzle:

QMS-#850

### What you have done:
Collect custom icons by iterating through external icons folder and it's subfolders.

1. Group custom icons in icons folder **and** it's subfolders
2. Run QMS
3. Set icons folder as external icons folder
4. Verify that icons from icons folder **and** it's subfolders are added.

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,6 +4,7 @@ V1.XX.X
 [QMS-843] Add waypoint icons
 [QMS-846] Better selection widget for waypoint icons
 [QMS-847] Remove duplicate Parking Area waypoint icon
+[QMS-850] Add custom icons from external icons folder and subfolders
 [QMS-851] Fix GIS item bubbles get transparent
 [QMS-855] Code cleanup: Convert waypoint icon map into a list
 [QMS-859] Add waypoint icons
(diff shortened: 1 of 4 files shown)


kiozen:
your chance to get it in :wink:  rebase and I press the button.

kiozen:
qt warning is fixed in my current branch see #856

JFritzle:
>That might make sense if you want to icons to have multiple tags and don't want to symlink them in subdirectories or if you're on Windows which doesn't have symlinks like Linux and macOS.

Icon `X.png` with full pathname `<external waypoints folder>/C/A,B/X.png` will have category C and tags A **and** B.
Using symlinks (available through command _mklink_ on Windows too) is no good idea. Using same file name for symlinked files in different folders would replace icon read first by icon read second and having only tag from icon read second. Using different file names for symlinked files would create 2 identical icons having different names and different tags but would require double process memory.

kiozen:
Imho using subfolders as category is a great idea. Using them for tags is overkill. 

The tags have 2 purposes:

- Give an alias as Sebastian already mentioned.
- Allow to translate tags to local language

Both do not apply to external icons. You can name the icons as you please, in the language you prefer.  The icon file name  is used  as name in the name/tag filter (without extension :wink: )

I am a bit concerned about ',' in the path, too. Hope that does not cause a problem with some OS.  I would expect that for external icons a single category is sufficient. The multiple categories per icon is more of another wacky thing by Garmin.

Pay attention that the concept of linked files is not known to Windows as such.

kiozen:
btw, does it make sense to have a section in the wiki with zip files for external icon collections that are free to share?

The internal ones should cover all Garmin stuff, as this is closest to kind of a standard. But everything else could be a download-as-you-need thing.

Man, this reminds me of the 90s having huge collections of additional font and clip art collections. He who dies with the most wins :rofl:

Issue #871, opened 2025-11-06
Installation folder name with non-ASCII characters prevents QMapShack startup on Windows


wthaem:
### Describe the bug

Installation folder name with non-ASCII characters prevents QMapShack startup on Windows

### What have you done to circle down the problem?

Tried various names of installation folders with QMS Windows 11 version. No startup issue if ASCII-only name, some files (routino xml for QMS, gdal files for QMT) not found if German Umlaut in folder name

### To Reproduce

1. Use Windows 11
1. Install QMapShack in folder with ASCII-only name. 
2. Start QMS and QMT and see that there are no startup issues
3. Copy installation folder to folder with German Umlaut or other non-ASCII character
4.  Start QMS and QMT and see that there are warnings about missing files during startup.

### Expected behavior
No missing files during startup

### Screenshots
Screenshots taken when starting from folder "c:\Temp\a aввв" (cyrillic)

![Image](https://github.com/user-attachments/assets/0b860d4f-cb09-4cb1-9556-1888295af2d4)
![Image](https://github.com/user-attachments/assets/213be6fc-6958-4a49-afdd-21242dd9d332)

### Attachments
[comment]: # (Can you provide an example file to trigger the problem?)

### Tracebacks
[comment]: # (Add your backtrace below if you have one. If QMapshack crashes the fastest way to get help is a backtrace.
For Linux see: https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux.
For Windows it would need a debug build and running QMapshack in Visual Studio.)

### Desktop
[comment]: # (please complete the following information)

 - OS: Windows 11
 - QMapShack Version:   latest dev
 - Window Manager:  [e.g. KDE, GNOME]

### Additional context
[comment]: # (Add any other context about the problem here)

wthaem:
I came across the issue when installing a new notebook with a new OS. MS was so kind to insert my name automatically and without my confirmation into the name of the folder recommended for user installations ("system path": c:\Users\<user_name>\AppData\Local). And in my name there is an "ä". I made a change in the Windows installer proposing as default the folder used for "all users" installations and added the recommendation to avoid non-ASCII characters for folder names.

JFritzle:
Renamed installation folder to `qmapshack-installä` containing German umlaut "ä" and encountered same issue.

I fixed issue and will contribute my solution:

```
...
[...]

PR #872, opened 2025-11-08
[QMS-871] Partial fix: Routino & installation folder name with non-ASCII-characters

JFritzle:

QMS-#871

As issues mentioned there have different causes, this code fix only covers the Routino problem, issue screenshot 1.

Routino software is using POSIX functions for file access. By default, Microsoft's POSIX implementation is not Unicode aware.
According to [here](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170), **valid** Windows filenames passed to functions depend on system's current ANSI codepage.
System's ANSI codepage can be checked by Windows console command 
`REG QUERY "HKLM\SYSTEM\CurrentControlSet\Control\Nls\CodePage" -v ACP|find /I "ACP"` 
Western Europe ANSI codepage is usually 1252, see it's character set [here](https://en.wikipedia.org/wiki/Windows-1252).

At least since some Windows 10 release, there is a switch called _Beta: Use Unicode UTF-8 for worldwide language support_ to be found at _Settings -> Time & language -> Language & region -> Administrative language settings -> Change system locale_:
Clipboard 1
When enabled, above console command returns codepage 65001 = Unicode (UTF-8).

To fix issue screenshot 2 too, enable _Use Unicode UTF-8 for worldwide language support_.

### What you have done:
On Windows OS, all filenames passed from QMS to Routino functions are now using system's local encoding.

1. Rename QMS installation folder on Windows OS to folder name containing non-ASCII character from current ANSI character set,
  e.g. folder name _qmapshack-installation-äéø_ containing characters "ä", "é", "ø", etc. when your system's ANSI codepage is 1252
  or
  e.g. folder name _qmapshack-installation-äéøαв_ containing characters "ä", "é", "ø", "α", "в" etc. when your system's ANSI codepage is 65001
2. Run QMS from that folder
3. Do not see any more error message about Routino's missing XML language file
4. Verify that Routino router can be configured and is working as expected
5. When ANSI codepage is 65001, verify that tools internally called by QMS like "VRT Builder", "QMapTool", etc. are working as expected

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,6 @@
+V1.XX.X
+[QMS-871] Partly fix: Routino & installation folder name with non-ASCII characters
+
 V1.18.2
 [QMS-672] Fix track point timestamps to QDateTime UTC conversion
 [QMS-841] Fix waypoint names to match the Garmin symbols
--- a/src/qmapshack/gis/rte/router/CRouterRoutino.cpp
+++ b/src/qmapshack/gis/rte/router/CRouterRoutino.cpp
@@ -32,6 +32,12 @@
 #include "helpers/CSettings.h"
 #include "setup/IAppSetup.h"
 
+#ifdef Q_OS_WIN
+#define ROUTINO_PATH_ENCODING toLocal8Bit()
+#else
+#define ROUTINO_PATH_ENCODING toUtf8()
+#endif
+
 QPointer<CProgressDialog> CRouterRoutino::progress;
 
 int ProgressFunc(double complete) {
@@ -66,7 +72,7 @@ CRouterRoutino::CRouterRoutino(QWidget* parent) : IRouter(true, parent) {
 
   int res = 0;
   IAppSetup* setup = IAppSetup::getPlatformInstance();
-  res = Routino_ParseXMLTranslations(setup->routinoPath("translations.xml").toUtf8());
+  res = Routino_ParseXMLTranslations(setup->routinoPath("translations.xml").ROUTINO_PATH_ENCODING);
   if (res) {
     QMessageBox::critical(this, "Routino...", xlateRoutinoError(Routino_errno), QMessageBox::Abort);
     return;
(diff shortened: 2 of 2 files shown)


kiozen:
@wthaemelt can you please test?

JFritzle:
Path _Settings -> Time & language -> Language & region -> **Administrative language settings** -> Change system locale_ to checkbutton _Beta: Use Unicode UTF-8 for worldwide language support_ seems not to exist on _Windows Home Edition_. Run command _intl.cpl_ from console instead and choose _Change system locale_.

wthaem:
@JFritzle: My system is Windows 11 Pro German and I can select the option "Beta: Use Unicode UTF-8 for worldwide language support/Beta: Verwenden Sie Unicode UTF-8 für die Unterstützung weltweiter Sprachen" (my default: off/aus). But I don't know the implications of this selection. Does it reset any "Deutsch" settings? Can I switch it off again without damage to my system? What happens if I use this option permanently? Any comment welcome. Thanks!

JFritzle:
Checkbutton changes system's codepage from 1252 _Western Europe_ (which is usually set for `current system locale` when choosing any Western Europe language) to 65001 _Unicode (UTF-8)_ but does not change system's locale, here _German_ or _Deutsch_. This is comparable to changing Linux environment variable LANG from _de_DE_ to _de_DE.UTF-8_.

This does not have any impact on most executables which usually handle UTF-8 encoded text strings correctly.

It has an impact on (usually very old) programs, where cp1252 encoded text strings are burnt into program code or read from cp1252 encoded text files. On my system, I found exactly one program affected. It was reading language specific translations from a cp1252 encoded text file marked as ANSI, which is looking strange after checking checkbutton:
grafik
But after converting text file to UTF-8 by some conversion tool, _iconv_ or similar, text is looking correctly, in notepad editor as well as at program execution:
grafik
We see, plain ANSI text files may be affected too, but no word document, no PDF file, ...

And, what is most important, it has an impact on program code using Microsoft's POSIX shell which is a layer above Windows API. Example: calling function _fopen_ interprets the passed filename text string using the system's codepage. With system's codepage 1252, no cyrillic characters in filename are recognized as such and lead to hieroglyphs in filename of created file. With system's codepage 65001 all valid UTF-8 characters are recognized as such and lead to filename as expected.

In your case, it does not reset any "Deutsch" settings but it decides how to interpret text strings containing non-ASCII characters in non-unicode aware program code.
Yes, you can switch it off again without damaging your system. But after switching, a reboot is required.
If you use this option permanently, non-unicode aware program code may behave differently, especially when containing hard-coded German umlauts.

JFritzle:
Note 1:
As said, this patch only solves "Routino not working". Currently, QMS file to be patched already passes filename using system's local encoding to **1 single** Routino function. Patch now passes filenames to **all** Routino functions using system's local encoding. So merging this patch is a consistent code harmonization.

Note 2:
To make QMS installation **fully** UTF-8 filenames compatible, a Windows OS wide option has to be set.
This option solves problem when calling further executables by QMS executable.
This is due to "QStandardPaths::findExecutable(const QString &executableName)" not finding path executable when QMS installation path does contain non-ASCII characters. In my opinion a Qt bug.

wthaem:
@kiozen: Didn't have enough time to test. But I believe you can merge. JFritzle's arguments and tests seem to be convincing. Certainly, I confess, the issue describes quite an abnormal situation which is now at least partially resolved.

Issue #889, opened 2025-11-27
Indicate which DEM is used to resolve elevation


kiozen:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

One million € question: In a stack of DEM files with different resolution, which one is used?  And if DEM overlap, which one is used?

I know it, because I wrote the stuff. But do you know?

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

As we have this lovely hacky tree widget item widget to display all kind of information about a DEM, why not add an indicator when reading elevation data from a file. 

<img width="384" height="174" alt="Image" src="https://github.com/user-attachments/assets/3c50477d-4b26-4ebb-a872-b089eaddda94" />

Right now the idea is to display the elevation read for a second. This happens each time something calls CDemDraw::getElevationAt().  In most cases it's the mouse cursor. 

By that the user has a chance to comprehend what's going on under the hood.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

I am open for any other idea communicate that information.

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

I will open a PR next to test

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

JFritzle:
Verified -> ok!

JFritzle:
Maybe solvable, maybe not:

When activating a map, especially an online map, an amount of tiles has to be downloaded.
When deactivating this map quickly afterwards and **before** all tiles have been downloaded,
"Waiting for tiles" message remains for ever.

<img width="731" height="179" alt="Image" src="https://github.com/user-attachments/assets/a94487b8-e166-4708-863e-cd74d7bfec48" />

PR #890, opened 2025-11-27
[QMS-889] Indicate which DEM is used to resolve elevation

kiozen:

QMS-#889

### What you have done:

When calling getElevationAt for a DEM results in a valid elevation value, the value is displayed for 1 second in DEM's  list entry.

Activate some DEM files and move the mouse within the area covered by the files. Observe the DEM list.

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,6 +7,7 @@ V1.XX.X
 [QMS-882] Add visibility indicator to maps and DEM
 [QMS-884] Add command line option to query QMS version
 [QMS-887] Fix Map/DEM description and buttons disappear from list
+[QMS-889] Indicate which DEM is used to resolve elevation
 
 V1.18.2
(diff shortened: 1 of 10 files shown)


mitxel-m:
Working for me. 

Good improvement !
With this new feature, it is clear where the elevation data is taken from, and for sure it will help to understand how to make better use of various DEMs according to scale.

Selección_005

_In the image, `ESP_IGN-MDT05`  (5x5m online DEM) is only visible at a very close zoom level. So it is not visible at this zoom scale, is not demanding online data for the whole window area, but it provides data for the moue pointer and to be applied to tracks and WPs when drawing, filtering.... `IGN_MDT25_Peninsula` has lower resolution (25x25m and local file) and is used for hillshading and slope coloring._

Issue #1004, opened 2026-03-10
Progress bar in projects while restoring the workspace


kiozen:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

If there are many projects with a large number of items in the workspace, restoring the workspace takes some time. Currently the progress bar dialog is shown, blocking the UI until everything is load. 

The progress bar dialog itself is a problem as it introduces unexpected re-entrancy in the code. Therefore we want to get rid of it. From the user perspective it would be much nicer to have a responsive UI while the projects are loading. It would be also much better to see the progress per project instead of some abstract progress over all projects.

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

With the introduction of a progress bar for devices while loading the projects stored on the device, we have everything in place to do the same while loading projects when restoring the workspace.  The final result should look like this:

<img width="352" height="484" alt="Image" src="https://github.com/user-attachments/assets/71fac6db-1e2c-4290-a1f0-9dc4004c4f75" />

Multiple progress bars that show the progress of each individual project in the workspace.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

No Alternative if we want to get rid of the progress bar dialog one day.

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

Already implementing it

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

PR #1006, opened 2026-03-11
[QMS-1004] Progress bar in projects while restoring the workspace

kiozen:

QMS-#1004

### What you have done:

I broke up the operator<<() method of IGisProject into a part done when called and a thread loading the GIS items of a project. As everything UI related has to be done in the main UI thread, the thread main duty is to queue chunks of code to be executed in the main thread. This is a quite awkward way to do it but with the current data/view model of QMapShack the only way to do it.

It has some caveats: 
* performance is a bit less than before. This is the trade off to have a responsive UI, switching back and forth between threads.
* It's not 100% crash safe. See testing section for details

1. Load a few chunky projects into the workspace, 
2. Restart QMapShack
3. Projects should be restored displaying a progress bar
4. Moving the map while loading should be ok
5. Removing all projects (F8) while loading should be ok
6. Expanding a project while loading should be ok
7. !!! Closing QMS while loading will result into a segfault. This is known. I see no way to fix that.!!!

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,7 @@ V1.XX.X
 [QMS-994] Fix: QMapTool not callable from QMapShack (macOS)
 [QMS-996] Fix: POIs not displayed (macOS)
 [QMS-998] Add progress bar to GPS devices while loading records
+[QMS-1004] Add progress bar in projects while restoring the workspace
 
 V1.20.1
 [QMS-945] Redesign icons for focus, autosave, sync, and DB buttons
--- a/src/qmapshack/device/CDeviceGarmin.cpp
+++ b/src/qmapshack/device/CDeviceGarmin.cpp
@@ -195,7 +195,7 @@ void CDeviceGarmin::createProjectsFromFiles(QString subdirecoty, QString fileEnd
 
             if (project) {
               if (!project->isValid()) {
-                delete project;
+                project->destroyLater();
               } else {
                 project->setVisibility(isVisible());
               }
--- a/src/qmapshack/device/CDeviceGarminArchive.cpp
+++ b/src/qmapshack/device/CDeviceGarminArchive.cpp
@@ -69,7 +69,7 @@ void CDeviceGarminArchive::slotExpanded(QTreeWidgetItem* item) {
               IGisProject* project = new CGpxProject(filename, this);
               if (project) {
                 if (!project->isValid()) {
-                  delete project;
+                  project->destroyLater();
                 } else {
                   project->setVisibility(isVisible());
                 }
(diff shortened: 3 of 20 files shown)


kiozen:
@JFritzle @eyzdh @Phobooky and all others, please test. This code has very likely race conditions therefore testing on as many different platforms as possible is best practice.

JFritzle:
Question:
Should restoring each project run in it's own thread or should whole task "restoring the workspace" be 1 thread?

Yes, each project being restored displays a progress bar. But restoring one long time consuming project blocks restoring others as well blocks e.g. downloading map tiles. Intension?

kiozen:
Each project is restored in it's own thread. But (there is always a 'but')  each GIS item of a project has to be loaded and processed in the main UI thread. As a consequence a huge track with many waypoints to correlate stalls all the other tasks. 

It's not ideal. And if "today's me" would meet "past me" I would tell me that there is a reason why you separate the data model completely from the UI as Qt and any other adult tells you.  But I did not and now we suffer the pain.  So this is all a compromise. At the end of the day everything still has to be processed by the main UI thread. But now it's without the progress dialog, that introduced erratic  re-entrancy. Now we introduce real thread problems. Sounds like a stupid idea, but concurrent thread problems are easier to pin down that the other stuff.

kiozen:
Btw my private laptop is 8 years old. So it's quite slow (8 cores). My laptop at work is quite fast (20 cores). The impression of parallel loading much better on my work laptop than on my private one.  

I guess it's even worse on a computer that has less than 8 cores.  QMapshack usually uses 4 cores (threads), Main UI, map DEM and GIS. There are also POI and RT, but in most cases these are not used. Therefore on start up 4 cores are occupied and the system has to schedule all these additional threads from workspace loading somehow. The less cores, the more overload, the more stalling and overload.

kiozen:
Loading and removing projects via the database is another beast and does not correlate with the current patch.

Nevertheless it should not happen. I found a small suspicious part in the code when removing a database project. Maybe that was the culprit. I can't tell as I could not reproduce the crash. But let's see.

kiozen:
Nice! My wild guess was right :smile:

JFritzle:
Since latest commit, occasional but not reproducible crashes on macOS during track creation seem to be gone too 😄.

kiozen:
Wow! What a patch! It can't get better! Let's merge it before the bug reports hit :grin:

wthaem:
Windows 11 + CPU with 32 cores + 25.6MB GPX file with many tracks: Loads quickly, no progress bar visible.

Issue #1007, opened 2026-03-12
Endless route calculation not interruptable


JFritzle:
### Describe the bug
[comment]: # (A clear and concise description of what the bug is)

It _may_ happen that route calculation (or probably waiting for result of calculation) does not end and is not interruptable, rendering QMS not usable any more. Pressing _Cancel_ button does not interrupt calculation, only user interaction still possible is to move calculation window _Please wait..._ on screen around. 
~~There is no way out but to kill QMS.~~
Correction: Pressing _QMapShack -> About QMapShack_ displays _About_ window. After closing _About_ window, deadlock is over and we are back in track creation, allowing to set next track point.

Observed with router _BRouter_ on platform _macOS_.

### What have you done to circle down the problem?
[comment]: # (Did you compare to previous versions? What versions? Did you check if others have the same problem?)

Never noticed issue on platforms Linux and Windows. Therefore, it _might_ be a macOS problem.

### To Reproduce
[comment]: # (List the steps to reproduce the behavior)

Not reproducible, but sometimes happens while creating a track.
### Expected behavior
[comment]: # (A clear and concise description of what you expected to happen)

At least, calculation should be interruptable by _Cancel_ button.

### Screenshots
[comment]: # (If applicable, add screenshots to help explain your problem)

<img width="1466" height="711" alt="Image" src="https://github.com/user-attachments/assets/b9a5ddfb-782c-4b24-94f5-5c895ec387f6" />

### Attachments
[comment]: # (Can you provide an example file to trigger the problem?)

Probably not due to BRouter profile OM_Hiking:
[OM_Hiking.zip](https://github.com/user-attachments/files/25934273/OM_Hiking.zip)

### Tracebacks
[comment]: # (Add your backtrace below if you have one. If QMapshack crashes the fastest way to get help is a backtrace.
For Linux see: https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux.
For Windows it would need a debug build and running QMapshack in Visual Studio.)

### Desktop
[comment]: # (please complete the following information)

 - OS: macOS (?)
 - QMapShack Version:   current
 - Window Manager:  -

### Additional context
[comment]: # (Add any other context about the problem here)

JFritzle:
This is how to reproduce (on macOS):

1. Start creating track with local BRouter and automatic routing switched on
2. Set track point A
[...]

Issue #388, opened 2021-05-21
Crash when creating a track or a route with BRouter


wthaem:
### Describe the bug

Crash when creating a track or a route with BRouter while probably POI loading/drawing is not yet finished (but this isn't visible!)

### What have you done to circle down the problem?

Run RelWithDbg version to get call stack

### To Reproduce

1. Activate 2 POI files and some categories in each of them
2. Scroll and move the map window several times
3. Wait a little while so that the big rotating cursor isn't seen anymore - but, most likely, handling of the POIs on the map and/or redrawing the map hasn't been finished (couldn't detect the small rotating cursor on the map!).
4. Start adding a track or a route using BRouter (offline)
5. A crash happens and can reproduced several times

### Expected behavior

 No crash

### Screenshots
[comment]: # Not necessary

### Attachments:

 Not necessary

### Tracebacks

 Call stack:

```
 	KernelBase.dll!00007ffaf6e84b89()	Unknown
 	vcruntime140.dll!00007ffae3206480()	Unknown
>	qmapshack_RelDbg.exe!CRouterBRouter::synchronousRequest(const QVector<QPointF> & points, const QList<IGisItem *> & nogos, QPolygonF & coords, double * costs) Line 387	C++
 	qmapshack_RelDbg.exe!CRouterBRouter::calcRoute(const QPointF & p1, const QPointF & p2, QPolygonF & coords, double * costs) Line 339	C++
 	qmapshack_RelDbg.exe!ILineOp::tryRouting(IGisLine::point_t & pt1, IGisLine::point_t & pt2) Line 389	C++
 	qmapshack_RelDbg.exe!ILineOp::finalizeOperation(int idx) Line 421	C++
 	qmapshack_RelDbg.exe!ILineOp::slotTimeoutRouting() Line 249	C++
 	qmapshack_RelDbg.exe!CLineOpAddPoint::leftClick(const QPoint & pos) Line 85	C++
 	qmapshack_RelDbg.exe!CMouseAdapter::mouseReleaseEvent(QMouseEvent * e) Line 118	C++
 	qmapshack_RelDbg.exe!CCanvas::mouseReleaseEvent(QMouseEvent * e) Line 684	C++
 	[External Code]	
 	qmapshack_RelDbg.exe!main(int argc, char * * argv) Line 83	C++
 	[External Code]	
```

### Desktop

 - OS: Windows10
 - QMapShack Version:   commit 422c4c5

### Additional context

None

[...]

PR #1011, opened 2026-03-16
[QMS-1007] Fix: Endless route calculation not interruptable

JFritzle:

QMS-#1007
QMS-#388

### What you have done:

I have more or less completely adopted the changes/expansions of files 
- qmapshack/gis/rte/router/CRouterBRouter.cpp
- qmapshack/gis/rte/router/CRouterBRouter.h

from @eyzdh.

Code is now rich of comment lines which I left untouched, because mods are difficult to understand without.

During my BRouter track creation tests using new code on platforms Windows, Linux and macOS,
- I did not encounter crashes anymore 
- nor happened endless route calculation anymore.

To make sure, that others have the same experience, I would appreciate feedback for each platform from some developer/user before merging code into dev. Does track creation still work as expected?
 
1. Run QMS
2. Start creating track with local BRouter
3. Fire **rapidly** subsequent track points
4. Do not encounter crash or endless route calculation
5. Do not encounter anymore straight connection lines due to BRouter not responding fast enough
6. Set subsequent track points as usual
7. Verify that track creation behaves as usual

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/gis/rte/router/CRouterBRouter.cpp
+++ b/src/qmapshack/gis/rte/router/CRouterBRouter.cpp
@@ -19,8 +19,10 @@
 
 #include "gis/rte/router/CRouterBRouter.h"
 
-#include <QtNetwork>
-#include <QtWidgets>
+#include <QScopeGuard>
+#include <QNetworkReply>
+#include <QNetworkRequest>
+#include <QWidget>
 
 #include "CMainWindow.h"
 #include "canvas/CCanvas.h"
@@ -36,6 +38,9 @@
 
 CRouterBRouter* CRouterBRouter::pSelf;
 
+// ---------------------------------------------------------------------------
+// Constructor
+// ---------------------------------------------------------------------------
 CRouterBRouter::CRouterBRouter(QWidget* parent) : IRouter(false, parent) {
   pSelf = this;
   setupUi(this);
@@ -76,21 +81,48 @@ CRouterBRouter::CRouterBRouter(QWidget* parent) : IRouter(false, parent) {
 
   localBRouter = new CRouterBRouterLocal(*this);
 
+  // Debounce: coalesces rapid calcRoute(key) calls into a single request.
+  // The mutex is NOT held during the debounce window — see header for the
+  // full flow description.
+  timerDebounce = new QTimer(this);
+  timerDebounce->setSingleShot(true);
+  timerDebounce->setInterval(DEBOUNCE_TIME);
+
+  routeCache.setMaxCost(MAX_CACHED);
+
   updateDialog();
 
   SETTINGS;
-
   cfg.beginGroup("Route/brouter");
   comboProfile->setCurrentIndex(cfg.value("profile", 0).toInt());
   comboAlternative->setCurrentIndex(cfg.value("alternative", 0).toInt());
   cfg.endGroup();
 }
 
+// ---------------------------------------------------------------------------
+// Destructor
+// ---------------------------------------------------------------------------
 CRouterBRouter::~CRouterBRouter() {
   isShutdown = true;
+
+  timerDebounce->stop();
+  hasPendingKey = false;
+
+  if (pendingAsyncReply) {
+    pendingAsyncReply->disconnect();
+    pendingAsyncReply->abort();
+    pendingAsyncReply->deleteLater();
+    pendingAsyncReply = nullptr;
+    if (asyncMutexHeld) {
+      asyncMutexHeld = false;
+      mutex.unlock();
+    }
+  }
+
   if (!localBRouter->isBRouterNotRunning()) {
     localBRouter->stopBRouter();
   }
+
   SETTINGS;
   cfg.beginGroup("Route/brouter");
   cfg.setValue("profile", comboProfile->currentIndex());
@@ -98,6 +130,33 @@ CRouterBRouter::~CRouterBRouter() {
   cfg.endGroup();
 }
 
+// ---------------------------------------------------------------------------
+// scheduleNextIfPending
+//
+// Called at the end of every async completion path (success, error, cancel).
+//
+// Calls startAsyncRequest() directly instead of calcRoute() — this skips the
+// DEBOUNCE_TIME ms entirely for catch-up requests. The debounce only makes
+// sense when the user is actively setting points; here the user already
+// waited for the previous request to finish, so adding another DEBOUNCE_TIME ms
+// would just feel sluggish.
+//
+// Uses QueuedConnection so the UI gets exactly one event-loop tick to repaint
+// the newly finished route before the next calculation begins.
+// ---------------------------------------------------------------------------
+void CRouterBRouter::scheduleNextIfPending() {
+  if (!hasPendingKey) {
+    return;
+  }
+  const IGisItem::key_t nextKey = pendingKey;
+  hasPendingKey = false;
+  // Skip debounce — go directly to startAsyncRequest() for catch-up.
+  QMetaObject::invokeMethod(this, [this, nextKey]() { startAsyncRequest(nextKey); }, Qt::QueuedConnection);
+}
+
+// ---------------------------------------------------------------------------
+// UI slots
+// ---------------------------------------------------------------------------
 void CRouterBRouter::slotToolSetupClicked() {
   localBRouter->stopBRouter();
   CRouterBRouterSetupWizard setupWizard;
@@ -106,6 +165,7 @@ void CRouterBRouter::slotToolSetupClicked() {
   setup->load();
   getBRouterVersion();
   updateDialog();
+  routeCache.clear();
 }
 
 void CRouterBRouter::slotToolProfileInfoClicked() const {
@@ -184,9 +244,9 @@ void CRouterBRouter::slotCloseStatusMsg() const {
 }
 
 QString CRouterBRouter::getOptions() {
-  return QString(tr("profile: %1, alternative: %2")
-                     .arg(comboProfile->currentText())
-                     .arg(comboAlternative->currentData().toInt()));
+  return QString(tr("profile: %1, alternative: %2"))
+      .arg(comboProfile->currentText())
+      .arg(comboAlternative->currentData().toInt());
 }
 
 void CRouterBRouter::routerSelected() { getBRouterVersion(); }
@@ -200,9 +260,9 @@ QNetworkRequest CRouterBRouter::getRequest(const QVector<QPointF>& routePoints, 
 
   for (const QPointF& pt : routePoints) {
     if (!lonLats.isEmpty()) {
-      lonLats.append("|");
+      lonLats += '|';
     }
-    lonLats.append(QString("%1,%2").arg(pt.x(), 0, 'f', 6).arg(pt.y(), 0, 'f', 6));
+    lonLats += QString("%1,%2").arg(pt.x(), 0, 'f', 6).arg(pt.y(), 0, 'f', 6);
   }
 
   QString nogoStr;
@@ -213,17 +273,17 @@ QNetworkRequest CRouterBRouter::getRequest(const QVector<QPointF>& routePoints, 
     switch (item->type()) {
       case IGisItem::eTypeWpt: {
         CGisItemWpt* wpt = static_cast<CGisItemWpt*>(item);
-        const qreal& rad = wpt->getProximity();
+        const qreal rad = wpt->getProximity();
         if (rad != NOFLOAT && rad > 0.) {
           const QPointF& pos = wpt->getPosition();
           if (!nogoStr.isEmpty()) {
-            nogoStr.append("|");
+            nogoStr += '|';
           }
-          nogoStr.append(QString("%1,%2,%3").arg(pos.x(), 0, 'f', 6).arg(pos.y(), 0, 'f', 6).arg(rad, 0, 'f', 0));
+          nogoStr += QString("%1,%2,%3").arg(pos.x(), 0, 'f', 6).arg(pos.y(), 0, 'f', 6).arg(rad, 0, 'f', 0);
         }
         break;
       }
-
+ 
       case IGisItem::eTypeOvl:
       case IGisItem::eTypeRte:
       case IGisItem::eTypeTrk: {
@@ -232,26 +292,25 @@ QNetworkRequest CRouterBRouter::getRequest(const QVector<QPointF>& routePoints, 
         QPolygonF polygon;
         line->getPolylineDegFromData(polygon);
         QString nogoPoints;
-        for (const QPointF point : std::as_const(polygon)) {
+        for (const QPointF& point : std::as_const(polygon)) {
           if (!nogoPoints.isEmpty()) {
-            nogoPoints.append(",");
+            nogoPoints += ',';
           }
-          nogoPoints.append(QString("%1,%2").arg(point.x(), 0, 'f', 6).arg(point.y(), 0, 'f', 6));
+          nogoPoints += QString("%1,%2").arg(point.x(), 0, 'f', 6).arg(point.y(), 0, 'f', 6);
         }
         if (item->type() == IGisItem::eTypeOvl) {
           if (!nogoPolygons.isEmpty()) {
-            nogoPolygons.append("|");
+            nogoPolygons += '|';
           }
-          nogoPolygons.append(QString("%1").arg(nogoPoints));
+          nogoPolygons += nogoPoints;
         } else {
           if (!nogoPolylines.isEmpty()) {
-            nogoPolylines.append("|");
+            nogoPolylines += '|';
           }
-          nogoPolylines.append(QString("%1").arg(nogoPoints));
+          nogoPolylines += nogoPoints;
         }
         break;
       }
-
       default: {
         Q_ASSERT(false);
       }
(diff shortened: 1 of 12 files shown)


wthaem:
No issues noticed when adding track, route, or area and playing with them (adding/moving points). Longest track ~2500km, longest track segment calculation on my relatively fast CPU ~ 90sec. Cancelling calculation possible (but don't move the mouse during cancellation, otherwise the calculation restarts due to the new mouse location). No crash in the form I reported a long time ago (but here my new system can be involved, too!).

src/qmapshack/gis/rte/router/CRouterBRouter.h
```diff
@@ -86,6 +98,59 @@ class CRouterBRouter : public IRouter, private Ui::IRouterBRouter {
...
+  // freely update pendingKey with the newest user position.
+  QTimer* timerDebounce{nullptr};
+
+  // The most-recently requested key that has not yet been sent to BRouter.
+  // Set by calcRoute() whenever a newer call arrives; cleared by
+  // startAsyncRequest() when that key is dispatched.
+  // hasPendingKey is true whenever pendingKey contains valid data.
+  bool hasPendingKey{false};
+  IGisItem::key_t pendingKey;
+
```
kiozen, line 140: This is not a good code pattern as it violates the single-source-of-truth pattern. IGisItem::key_t has a clear() method. You can add a isValid() method to test on all empty string. isValid() replaces the  hasPendingKey flag. If the pending key is processed it will be cleared.

src/qmapshack/gis/rte/router/CRouterBRouter.h
```diff
@@ -86,6 +98,59 @@ class CRouterBRouter : public IRouter, private Ui::IRouterBRouter {
...
+  // True while the async path holds the mutex (between startAsyncRequest()
+  // and the corresponding release in slotRequestFinished / cancel).
+  bool asyncMutexHeld{false};
```
kiozen, line 143: The mutex handling in the original code was bad. With that flag it does not get better.  

As far as I understand that code we do not have a multi-threading problem here. We have another CProgressBar problem. Therefore the problem is reentrancy because of  CProgressBar operating the event loop. And not reentrancy because of two concurrent threads. 

As long as everything runs in the same thread we should never be able  to write a variable at the same time. Therefore a single flag to signal that we have a pending operation should be fine.

src/qmapshack/gis/rte/router/CRouterBRouter.h
```diff
@@ -86,6 +98,59 @@ class CRouterBRouter : public IRouter, private Ui::IRouterBRouter {
...
+
+  // ── Route result cache ────────────────────────────────────────────────────
+  struct CachedRoute {
+    QPolygonF coords;
+    qreal costs{0.0};
+  };
+  // MAX_CACHED entries max; negligible memory, eliminates redundant requests
```
kiozen, line 150: I know there are bad examples in the code but coding guidelines require cached_route_t as structure name

https://github.com/Maproom/qmapshack/wiki/DeveloperCodingGuideline

src/qmapshack/gis/rte/router/CRouterBRouter.h
```diff
@@ -86,6 +98,59 @@ class CRouterBRouter : public IRouter, private Ui::IRouterBRouter {
...
+  //   slotRequestFinished / cancel:
+  //     → mutex released
+  //     → scheduleNextIfPending() re-invokes calcRoute(pendingKey) if set
+  //
+  // Net effect: BRouter receives at most one request per DEBOUNCE_TIME ms idle window,
+  // the last user-set position is always eventually computed, and no stale
+  // intermediate results can overwrite a newer one.
+
+  // The in-flight reply (nullptr when idle).
+  QNetworkReply* pendingAsyncReply{nullptr};
```
kiozen, line 127: Use a QPointer. Much easier and more reliable handling of instances getting destroyed.

src/qmapshack/gis/rte/router/CRouterBRouter.h
```diff
@@ -20,13 +20,17 @@
...
 #include <QNetworkAccessManager>
 #include <QProcess>
 #include <QTimer>
 
 #include "gis/rte/router/IRouter.h"
 #include "ui_IRouterBRouter.h"
 
+#define DEBOUNCE_TIME 200
+#define MAX_CACHED 50
+
```
kiozen, line 33: If defines are not used by the header they do not belong in the header. 

In modern c++ this:
constexpr int kDebounceTime = 200;

is better as the define has a type and the compiler can check on that type.

src/qmapshack/gis/rte/router/CRouterBRouter.h
```diff
@@ -69,10 +73,18 @@ class CRouterBRouter : public IRouter, private Ui::IRouterBRouter {
...
   QUrl getServiceUrl() const;
+  QString buildCacheKey(const QVector<QPointF>& points, const QList<IGisItem*>& nogos) const;
+
+  // Fires the actual HTTP request for the given key.
+  // Called by the debounce timer — at this point no request is in-flight.
+  void startAsyncRequest(const IGisItem::key_t& key);
+
+  // Schedules a catch-up call for pendingKey (if set) after the mutex has
+  // been released. Used on success, error, and cancel paths alike.
+  void scheduleNextIfPending();
```
kiozen, line 87: Imho scheduleNext() would be better. Took me a couple of minutes to understand what the "If" is standing for 😆

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -76,28 +81,82 @@ CRouterBRouter::CRouterBRouter(QWidget* parent) : IRouter(false, parent) {
...
   localBRouter = new CRouterBRouterLocal(*this);
 
+  // Debounce: coalesces rapid calcRoute(key) calls into a single request.
+  // The mutex is NOT held during the debounce window — see header for the
+  // full flow description.
+  timerDebounce = new QTimer(this);
+  timerDebounce->setSingleShot(true);
+  timerDebounce->setInterval(DEBOUNCE_TIME);
+
+  routeCache.setMaxCost(MAX_CACHED);
```
kiozen, line 91: What is happening if the max cost is exceeded? Items are deleted immediately. Does require special handling?

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -76,28 +81,82 @@ CRouterBRouter::CRouterBRouter(QWidget* parent) : IRouter(false, parent) {
...
+// ---------------------------------------------------------------------------
+// Destructor
+// ---------------------------------------------------------------------------
 CRouterBRouter::~CRouterBRouter() {
   isShutdown = true;
+
+  timerDebounce->stop();
+  hasPendingKey = false;
+
+  if (pendingAsyncReply) {
```
kiozen, line 111: ok, but it's nicer to compare with nullptr.

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -106,6 +165,7 @@ void CRouterBRouter::slotToolSetupClicked() {
   setup->load();
   getBRouterVersion();
   updateDialog();
+  routeCache.clear();
```
kiozen, line 168: Why at this point. And what about all the other async stuff?

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -232,26 +292,25 @@ QNetworkRequest CRouterBRouter::getRequest(const QVector<QPointF>& routePoints,
         QPolygonF polygon;
         line->getPolylineDegFromData(polygon);
         QString nogoPoints;
-        for (const QPointF point : std::as_const(polygon)) {
+        for (const QPointF& point : std::as_const(polygon)) {
```
kiozen, line 295: using a const reference is perfect 👍 And in this case we can get rid of std::as_const.

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -290,15 +365,27 @@ int CRouterBRouter::calcRoute(const QPointF& p1, const QPointF& p2, QPolygonF& c
   QList<IGisItem*> nogos;
   CGisWorkspace::self().getNogoAreas(nogos);
 
+  const QString cacheKey = buildCacheKey(points, nogos);
```
kiozen, line 368: const ref

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -290,15 +365,27 @@ int CRouterBRouter::calcRoute(const QPointF& p1, const QPointF& p2, QPolygonF& c
...
+  if (const CachedRoute* cached = routeCache.object(cacheKey)) {
```
kiozen, line 369: imho assigning cached first and then checking for nullptr in the if is a bit nicer.

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -307,145 +394,279 @@ int CRouterBRouter::synchronousRequest(const QVector<QPointF>& points, const QLi
   synchronous = true;
 
   QNetworkReply* reply = networkAccessManager->get(getRequest(points, nogos));
+  auto replyGuard = qScopeGuard([reply] { reply->deleteLater(); });
+  Q_UNUSED(replyGuard);
```
kiozen, line 398: What for? And if, why not for mutexGuard?

kiozen (commented):
I am not done yet with the PR. But I have to switch task. Will continue later. Don't bother about the review yet.

JFritzle:
I am sorry, program logic is beyond my understanding. I have to delegate code back to @eyzdh and wait for an update.

wthaem:
**Task: Given the waypoints of the attached GPX files, try to build a route from P1 to P7 using BRouter. Choose trekking option.**

[broutertest1.zip](https://github.com/user-attachments/files/26059981/broutertest1.zip)


The calculation of each leg/route segment takes in my case 10 sec. or more due to the large distance between the points.

* Go to P1 and select "add route" from the context menu at this point
* Try to click rapidly one after the other the given waypoints with the left mouse button and look what happens
* Notice the erratic/irritating behaviour of the last shown green square depending on the mouse position: either the last click gets lost, or a new calculation starts at some intermediate and unwanted new mouse position, or clicking the mouse at the next point is not possible as long as the calculation runs, ...

This behaviour can't be seen with short distances between the waypoints.

More observations:

*Info window isn't removed:*

* Routing engine switched from BRouter to Routino
* Define new route from waypoints - triggers automatic route recalculation
* GUI shows still BRouter info window - no chance to remove it, the window reappears even after loading WMTS map tiles!

BRouterRequest


*Jumping roads:*

* Used map: velo DE (gmapsupp type)
* A waypoint (P1) seems to be exactly positioned on a road
* Depending on the zoom level, it is either on the road or only near the road as shown in the attached snapshots
* Doesn't happen with the OSM Topo map (WMTS type). P1 is exactly on the road for each zoom level.
* My conclusion: not the waypoint but the underlying road location depends on the zoom level (can be an understandable coordinate rounding problem)
* This is important for the construction of routes or tracks because routing depends on the (nearly) exact position of a point on a road. If the road is at a wrong position, the selected point will not be on the road (the one used by the routing engine) and the routing engine may fail.

capture_20260317_160933
capture_20260317_160849
capture_20260317_160822

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -307,145 +394,279 @@ int CRouterBRouter::synchronousRequest(const QVector<QPointF>& points, const QLi
...
-    const QByteArray& res = reply->readAll();
+  if (res.isEmpty()) {
+    throw tr("response is empty");
+  }
 
-    if (res.isEmpty()) {
-      throw tr("response is empty");
+  QDomDocument xml;
+  {
+    const QDomDocument::ParseResult parseResult = xml.setContent(res);
```
kiozen, line 442: for complex objects always const &

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -307,145 +394,279 @@ int CRouterBRouter::synchronousRequest(const QVector<QPointF>& points, const QLi
...
+  // Do NOT call rte->reset() here — the old route stays visible while the
+  // new one is being calculated, which avoids the visual flicker of having
+  // nothing displayed during the request. reset() is called in
+  // slotRequestFinished() just before applying the new result.
+  slotCloseStatusMsg();
+
   synchronous = false;
 
   QNetworkReply* reply = networkAccessManager->get(getRequest(points, nogos));
+  pendingAsyncReply = reply;
```
kiozen, line 585: This code is really hard to read and check. You do stuff in one method and assume that stuff is done in another. Are you sure pendingAsyncReply is always nullptr? If not we have a memory leak.  And it's impossible for someone else to see all these cross method dependencies. The recipe for introducing regressions with future changes. 

Use at least a smart pointer (std::unique_ptr) to prevent memory leaks. In C++ you have to take care of pointers. You have to define who owns the pointer. Who observes the pointer. The owner instantiates the pointer and deletes it. The observer has to notice when a pointer is deleted.  All this should be at very defined single locations. Else disaster is pre-ordered.

Anyway, this is really just bitching around another problem caused by CProgressBar. Imho a threaded approach would be much cleaner. By that everything is calculated in a single thread and when done the UI is updated.

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -307,145 +394,279 @@ int CRouterBRouter::synchronousRequest(const QVector<QPointF>& points, const QLi
...
+  // Report "calculating" status — do NOT trigger a full canvas repaint here.
+  // The canvas will redraw itself when slotRequestFinished() applies the new
+  // route, avoiding one spurious repaint per request.
   CCanvas* canvas = CMainWindow::self().getVisibleCanvas();
   if (canvas) {
-    canvas->slotTriggerCompleteUpdate(CCanvas::eRedrawGis);
     canvas->reportStatus("BRouter", tr("<b>BRouter</b><br/>Routing request sent to server. Please wait..."));
   }
 
+  delete progress;
```
kiozen, line 601: again one of the smart pointers, either Qt or std would make this much safer

src/qmapshack/gis/rte/router/CRouterBRouter.cpp
```diff
@@ -307,145 +394,279 @@ int CRouterBRouter::synchronousRequest(const QVector<QPointF>& points, const QLi
...
+
+  // Release the mutex and schedule the next pending key on every exit path.
+  auto cleanupGuard = qScopeGuard([this] {
+    if (asyncMutexHeld) {
+      asyncMutexHeld = false;
+      mutex.unlock();
+    }
+    scheduleNextIfPending();
+  });
+  Q_UNUSED(cleanupGuard);
```
kiozen, line 650: one should be enough

kiozen (changes requested):
Well the patch solves a couple of problems for sure. But it introduces a lot of complexity, too. Without diving deep into that code  I can't tell if there is a more elegant way to solve it as a single threaded / reentrant by event loop approach. Or if it is easier to solve by a thread. What I can tell for sure: The progress bar has to go. To get rid of CProgressBar is top priority as it is the root of many problems. 

It would be really nice to rethink the code to make it  easier instead of making it more complex.  Especially this cross method handling of data is painfully known to me. It never ended well in the past.

JFritzle:
@wthaem - about _jumping roads_:

I would overlay a semitransparent OSM Topo map over velo DE map. Roads _should_ be exactly on top of each other. Do they at all for any zoom level? If point p1 deviates from velo DE map's road when zooming but stays on  OSM Topo map's road, then I guess complete velo DE map's road does no longer cover OSM Topo map's road.
-> In my opinion, this is the only possible explanation: Either velo DE map data get inaccurate when zooming or projection of map data onto map canvas gets inaccurate.

wthaem:
@JFritzle: You are almost correct. There is yet another parameter influencing the accuracy (I didn't expect this!): the details level selected for the vector map as you can see from the attached snapshots (same zoom level, different level of detail, snapshots show overlay of 2 maps!).

capture_20260317_195650
capture_20260317_195521

JFritzle:
@wthaem 

I took left screenshot and overlayed semitransparent right screenshot
applied with a slight clockwise twist of about 6 degrees :
grafik
What do we see: roads again exactly match. Road's slight bend to the right near to point p1 fits exactly!
-> Conclusion: changing velo DE map's level of detail applies a twist on projection. Maybe inaccuracy in computing projection?

But this seems to be an issue different from QMS-#1007,

wthaem:
@JFritzle : 

> But this seems to be an issue different from QMS-https://github.com/Maproom/qmapshack/issues/1007,

Sure, I came across this one when preparing the example for the BRouter test and was confused that I did not hit the road with my point.

And: it is no projection issue. It is the (understandable) reduction of the number of points used to draw the road (using DP algorithm or something similar). Load the velo map, look at the Bundesstraße (especially in the nearby location of Haselünne) involved at different detail levels, and see what happens.

For me, it is a feature, not an issue!

kiozen:
Garmin maps have layers. For outer zoom levels and more closer zoom levels. The details factor is manipulating the decision about what layer to use at what zoom level. Each layer uses a different number of bits to code coordinates. And it has a different number of elements like roads, and waypoints. 

This is fine as long as the layer is displayed at a matching zoom level. However if a layer with less resolution is displayed at a closer zoom level, the quantification errors become very obvious. At closest zoom even the error of the 24bit resolution is obvious, e.g.  badly shaped buildings. 

This is also a problem if you use "follow polyline" while drawing tracks. If you do that on an outer zoomlevel and zoom in afterwards the line is not matching the line of the map.

JFritzle:
@wthaem 
I downloaded and installed velomap_germany_14.03.2026.
This is what Garmin BaseCamp shows from left to right - Lowest level of detail - Highest level of detail - Overlay of both:
grafik
-> Roads match exactly. No twist. Independent from zoom level.

Then I created a gmapsupp for QMS -> Roads match exactly for each level of detail too. Independent from zoom level.
And: Roads match exactly with OSM Topo map.

But there are numerous options either for installing downloaded map or for creating gmapsupp later.
Propably these options influence accuracy of levels of detail?

wthaem:
@JFritzle : see kiozen's comment. It explains everything. And we discuss QMS, not Basecamp.

JFritzle:
@wthaem: Sure, we discuss QMS. See my comment: At my site, no inaccuracy depending on level of detail nor zoom level.

JFritzle:
As I have not been able to provide a own suitable solution, I will no longer work on issue.
My PR gets closed to allow someone else to contribute his solution.

Issue #1036, opened 2026-04-07
sqlite db corrupted after having qmapshack opened twice


keenonkites:
### sqlite db was corrupted after having opened qmapshack twice
After running into the problem  the affected db contained only part of the date it was having before the problem. Even a sqlite recover procedure didn't bring everything back

### What have you done to circle down the problem?
There was nothing to be done to narrowing the problem as the reason for it was clear

### To Reproduce
I somehow was able to start two instances of qmapshack in a very short time, probably by multiple clicks in the wrong time. The DB having the problem now contains lots of tracks and one of the folders was shown on the map while opening.

### Expected behavior
QmapShack should prevent multiple instances starting. This would prevent the problem.

### Desktop

 - OS: Windows 11
 - QMapShack Version:  1.19.0

### Additional context / Question
Is there a way to get the data back ? First try of sqlite recovery brought back part, but not everything.

Thanks and regards
Patrik

kiozen:
QMapShack already takes care to have only one instance active. See CSingleInstanceProxy. But this is not an atomic operation and there are race conditions for sure. If you have a better solution you are welcome.

Concerning the corrupted database: The sqlite recovery is our best option. I assume we are talking about the workspace database. As the workspace is temporary the loss shouldn't be to bad. Or are you talking about the real project database?

keenonkites:
Thanks for the feedback.

Looks like I was able to trigger it nevertheless with whatsoever. I really think I clicked it (on the touchpad) too many times in a too short time, so that both instances were thinking that they are alone. Don't know exactly how I was able to 'force' it, but at the end I really had two instances of qmapshack opened, unfortunately.

The workspace itself is also lost, but that's less of a concern. Even if I lost some things there it's not that bad. I always keep too much stuff there anyway ... ;-)
My main concern/problem is a real project database I use for the bookkeeping of my trips (for geocaching). As my last backup of this folder is from early January 2026 (shame on me) I lost quite a few trips in there.... I still have the source files from the GPS, but I would have to do the editing (cleanup,. assigning activities) again.
But if that's the only option, then I can/have to redo it again..... and make more frequent db backups of my project dbs... I was thinking about a wrapper script that does the backup (few generations) before really starting up qmapshack.

I'll play again with sqlite recovery, I was just trying it out last week while on the run. And I also will play around bit more with sqldiff for checking if something is lost in the other project dbs (besides the bookkeeping db another one was affected too, but that one is less problematic).

[...]

PR #1039, opened 2026-04-12
[QMS-1036] Fix: Prevent QMS from opening twice & corrupting sqlite db

JFritzle:

QMS-#1036

### What you have done:

Main issue - QMS opening twice:
Additionally use operating system's native file locking methods to distinguish between primary and secondary QMS instance.

Sub-issue 1 - transferring optional GIS filenames from secondary to primary QMS instance:
Slightly modified data exchange between secondary and primary QMS instance making it reliable.

Sub-issue 2 - primary QMS instance crashing at start due receiving list of GIS filenames from secondary instance:
Now checking if `&CMainWindow::self()` still points to NULL before calling any CMainWindow class function.

1. Verify that it is no longer possible to run more than 1 primary instance of QMS per user. In other words, QMS no longer opens twice. If desired, use debug switch _-d_ to watch output.
2. Run secondary QMS instance supplying additional GIS files on command line. See that GIS files are processed by primary QMS instance.

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/CSingleInstanceProxy.cpp
+++ b/src/qmapshack/CSingleInstanceProxy.cpp
@@ -21,28 +21,37 @@
 #include <QtNetwork>
 
 #include "CMainWindow.h"
+#include "setup/IAppSetup.h"
 
 CSingleInstanceProxy::CSingleInstanceProxy(const QStringList filenames) {
-  serverName = CMainWindow::self().getUser();
+  serverName = CMainWindow::getUser();
   if (serverName != "QMapShack") {
     serverName = "QMapShack-" + serverName;
   }
 
-  QLocalSocket socket;
-  socket.connectToServer(serverName);
-  if (socket.waitForConnected(1000)) {
-    // if the connection is successful another instance
-    // is already running. In that case the list of files to
-    // open is sent to the primary instance. And this instance
-    // will be closed imediately.
-    QDataStream stream(&socket);
-    stream << filenames;
-    socket.waitForBytesWritten(3000);
+  // setting lock failed -> secondary instance
+  if (!IAppSetup::getPlatformInstance()->setLock()) {
+    bool ok = false;
+    QLocalSocket socket;
+    socket.connectToServer(serverName);
+    if (socket.waitForConnected(1000)) {
+      // Another instance is already running. In that case
+      // the list of files to open is sent to the primary instance.
+      // And this secondary instance will be closed imediately.
+      QByteArray sendData;
+      QDataStream sendDataStream(&sendData, QIODevice::WriteOnly);
+      sendDataStream << filenames;
+      socket.write(sendData);
+      socket.waitForBytesWritten(3000);
 
-    // wait for confirmation
-    socket.waitForReadyRead(3000);
-    bool ok;
-    stream >> ok;
+      // wait for confirmation
+      socket.waitForReadyRead(3000);
+      QByteArray recvData = socket.readAll();
+      QDataStream recvDataStream(recvData);
+      recvDataStream >> ok;
+    }
+    socket.close();
+
     qDebug() << "Sent parameters to primary instance. Result" << ok;
     qDebug() << "There can only be one. Exit.";
     exit(0);
@@ -51,8 +60,8 @@ CSingleInstanceProxy::CSingleInstanceProxy(const QStringList filenames) {
   // Looks like we are the first instance.
   // Create a server socket and wait for other instances to connect.
   server = new QLocalServer(this);
-  connect(server, &QLocalServer::newConnection, this, &CSingleInstanceProxy::slotNewConnection);
   server->removeServer(serverName);
+  connect(server, &QLocalServer::newConnection, this, &CSingleInstanceProxy::slotNewConnection);
   if (!server->listen(serverName)) {
     qDebug() << "CSingleInstanceProxy: Failed to start single instance server socket.";
   } else {
@@ -71,20 +80,28 @@ void CSingleInstanceProxy::slotNewConnection() {
   // Each secondary instance will send a QStringList with files to open
   // The list can be empty.
   if (socket->waitForReadyRead(3000)) {
-    QStringList filenames;
-    QDataStream stream(socket);
-    stream >> filenames;
+    // main window might not yet exist
+    bool ok = !CMainWindow::isNull();
 
-    CMainWindow& w = CMainWindow::self();
-    w.loadGISData(filenames);
+    QByteArray recvData = socket->readAll();
+    QDataStream recvDataStream(recvData);
+    QStringList filenames;
+    recvDataStream >> filenames;
 
     // confirm that files are loaded
-    stream << true;
+    QByteArray sendData;
+    QDataStream sendDataStream(&sendData, QIODevice::WriteOnly);
+    sendDataStream << ok;
+    socket->write(sendData);
     socket->waitForBytesWritten(3000);
 
-    // raise the application window to top of desktop
-    w.raise();
-    w.activateWindow();
+    if (ok) {
+      CMainWindow& w = CMainWindow::self();
+      w.loadGISData(filenames);
+      // raise the application window to top of desktop
+      w.raise();
+      w.activateWindow();
+    }
   }
 
   socket->close();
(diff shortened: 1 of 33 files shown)


src/qmapshack/CSingleInstanceProxy.cpp
```diff
@@ -21,28 +21,42 @@
 #include <QtNetwork>
 
 #include "CMainWindow.h"
+#include "setup/IAppSetup.h"
 
 CSingleInstanceProxy::CSingleInstanceProxy(const QStringList filenames) {
   serverName = CMainWindow::self().getUser();
```
kiozen: Can you please remove the self()?  Background: self returns a reference and a reference to a nullptr is undefined. In this case we do not need the self() as getUser() is a static method.

src/qmapshack/CSingleInstanceProxy.cpp
```diff
@@ -21,28 +21,42 @@
...
   QLocalSocket socket;
   socket.connectToServer(serverName);
-  if (socket.waitForConnected(1000)) {
-    // if the connection is successful another instance
-    // is already running. In that case the list of files to
-    // open is sent to the primary instance. And this instance
-    // will be closed imediately.
-    QDataStream stream(&socket);
-    stream << filenames;
+  if (primary != 1 && socket.waitForConnected(1000)) {
```
kiozen: As far as I understand the patch we rely on the lock, now. So, does it make sense to create a socket and try to connect the socket if the lock is gained?

src/qmapshack/CSingleInstanceProxy.cpp
```diff
@@ -21,28 +21,42 @@
...
     socket.waitForReadyRead(3000);
-    bool ok;
-    stream >> ok;
+    QByteArray recvData = socket.readAll();
+    QDataStream recvDataStream(recvData);
+    recvDataStream >> ok;
+  }
+  socket.close();
+
+  if (primary == 0) {
```
kiozen: what happens if primary is -1 and waitForConnected() fails? Imho this code has some potential for clean-up

src/qmapshack/CSingleInstanceProxy.cpp
```diff
@@ -71,20 +85,28 @@ void CSingleInstanceProxy::slotNewConnection() {
   // Each secondary instance will send a QStringList with files to open
   // The list can be empty.
   if (socket->waitForReadyRead(3000)) {
-    QStringList filenames;
-    QDataStream stream(socket);
-    stream >> filenames;
+    // main window might not yet exist
+    bool ok = !QPointer(&CMainWindow::self()).isNull();
```
kiozen: As mentioned earlier self returns a reference and a reference to a nullptr is undefined. The correct way is to add a static method to CMainWindow that compares the private pSelf pointer to a nullptr returning a bool.

kiozen (changes requested):

src/qmapshack/CSingleInstanceProxy.cpp
JFritzle: Ok.

src/qmapshack/CSingleInstanceProxy.cpp
JFritzle: primary == -1 means that something was going wrong with creating new or opening existing lock file.
Currently, in this very very unlikely case detection for primary or secondary instance behaves unreliable as before.
What do you propose to do? Aborting QMS with an error message at all?

src/qmapshack/CSingleInstanceProxy.cpp
JFritzle: Fixed as proposed.

src/qmapshack/CSingleInstanceProxy.cpp
kiozen: Depends a bit on the root cause for the fail. I would like to avoid a broken mechanism due to a stale lock file or any other problem.  Thus if there is some user action required to fix the problem we should inform the user. 

That said maybe it's a good idea to place the log file in the system's temporary directory. At least in Linux this seems to be cleaned up on a restart. 

Anyway, the code is easier to maintain if we decide on a single variable `primary` what to do and then handle each case in a straight forward way.

src/qmapshack/CSingleInstanceProxy.cpp
JFritzle: Fixed: When primary is 1, socket lo longer gets created.

src/qmapshack/CSingleInstanceProxy.cpp
JFritzle: As long as there is no better idea, I will keep behavior:
If primary is -1 and waitForConnected() fails, instance believes to be a primary one.

kiozen: As far as I can see -1 happens if no lock file can be generated. Either because of a bad path or missing access rights. In that case multiple instances can be started permanently. I do not think that is a good condition.

I would suggest something like:

```
  if (primary == 0) {
    bool ok = false;
    QLocalSocket socket;
    socket.connectToServer(serverName);
    if (socket.waitForConnected(1000)) {
      // connected: force secondary instance
      // Another instance is already running. In that case
      // the list of files to open is sent to the primary instance.
      // And this secondary instance will be closed imediately.
      QByteArray sendData;
      QDataStream sendDataStream(&sendData, QIODevice::WriteOnly);
      sendDataStream << filenames;
      socket.write(sendData);
      socket.waitForBytesWritten(3000);

      // wait for confirmation
      socket.waitForReadyRead(3000);
      QByteArray recvData = socket.readAll();
      QDataStream recvDataStream(recvData);
      recvDataStream >> ok;
    }
    socket.close();

    qDebug() << "Sent parameters to primary instance. Result" << ok;
    qDebug() << "There can only be one. Exit.";
    exit(0);
  } else if (primary < 0) {
    QMessageBox::critical(nullptr, "xxxx", "yyyy");
    exit(-1);
  }
```
The message box should point the user to the lock file.

kiozen (changes requested):

src/qmapshack/CSingleInstanceProxy.cpp
JFritzle: A "stale lock file" does never happen as once created, it stays reusable forever. It must never be removed and it does not get removed except by intention and manually.

It is no good idea to put it into temporary directory as temporary folder may get cleaned up manually or by background job. Further on, in a multi-user environment different users may run same application on their own display (screenshot shows 2 independently running QMS primary instances for different users):
grafik
Therefore imho it is savest when each user locks his own user data folder which contains his personal data (e.g. workspace.db) to be protected from being corrupted.

JFritzle:
Following your suggestion, I will modify setLock function as follows:
1. Function returns true if lock gets acquired.
2. Function return false if lock does not get acquired because it is held by another QMS instance.
3. In case of an error else, setLock opens a critical message box showing system error and exits with code -1.
Ok?

JFritzle:
Calling `exit(EXIT_FAILURE);` in `CAppSetup*.cpp` exits from QMS with 2 ugly messages
```
QThreadStorage: entry 2 destroyed before end of thread 0x6115bd937f90
QThreadStorage: entry 1 destroyed before end of thread 0x6115bd937f90
```
How can I avoid that?

kiozen:
Imho they are Qt internal.

JFritzle:
Ugly messages are due to first calling QMessageBox and hit dialog's close button and calling exit after returning from QMessageBox. Messages do not occur when first calling qDebug instead of QMessageBox and the calling exit.

JFritzle:
Definitively QMessageBox is the culprit.
Calling `exit(0);` directly after `QApplication app(argc, argv);` closes application gracefully.
Calling `QMessageBox::critical(nullptr, "title", "message"); exit(0);` directly after `QApplication app(argc, argv); ` closes application after writing these 2 ugly messages.

Might QMessageBox expect to have a main window created and mapped?

Nevertheless, as proposed I call QMessageBox::critical in case of an error and exit application.
As error in function setLock hopefully will never happen, these messages will never be thrown.

kiozen:
exit(-1) is kind of an emergency anyway. I think no one cares of some "mimimimi" on that case.

Issue #1022, opened 2026-03-22
Wrong view active after restart


JFritzle:
### Describe the bug
[comment]: # (A clear and concise description of what the bug is)

Issue 1:
After restart, QMS may not show same view active being active at closing.

Issue 2:
When closing QMS while track editor tab is open, (for non-insiders) it is not predictable which view is active after restart. 

### What have you done to circle down the problem?
[comment]: # (Did you compare to previous versions? What versions? Did you check if others have the same problem?)

Issue is due to saving last active tab number instead of saving last active view's name or unique key.  
After restart, tab number no longer points to last active view.

### To Reproduce
[comment]: # (List the steps to reproduce the behavior)

See sequence of screenshots with description

### Expected behavior
[comment]: # (A clear and concise description of what you expected to happen)

After QMS restart, same map view is expected to be shown as before when closing QMS, regardless of the history.

### Screenshots
[comment]: # (If applicable, add screenshots to help explain your problem)

Issue 1 - Screenshot 1  
Track _Track_ was opened to edit from workspace,  
editor tab was moved to 1st position on the left,  
2nd view's _Cycling_ tab was selected:
<img width="1283" height="555" alt="Image" src="https://github.com/user-attachments/assets/660cb0eb-a3ac-4e7f-96ed-04dec0cde637" />

Issue 1 - Screenshot 2  
QMS was closed and restarted,  
3rd view _Hiking_ is now set as current view:
<img width="1283" height="581" alt="Image" src="https://github.com/user-attachments/assets/c11b681e-8a6c-436a-846b-bc2b3d43475b" />
  

Issue 2 - Screenshot 1  
Tracks _North-South_ and _West-East_ were opened to edit from workspace,  
editor tabs were moved as shown,  
editor tab _North-South_ was selected:
<img width="1283" height="581" alt="Image" src="https://github.com/user-attachments/assets/2734e0b1-a85a-4eef-b966-e77f6d3bbe92" />

Issue 2 - Screenshot 2  
While editor tab _North-South_ was open, QMS was closed and restarted,  
3rd view _Hiking_ is now set as current view:
<img width="1283" height="581" alt="Image" src="https://github.com/user-attachments/assets/c11b681e-8a6c-436a-846b-bc2b3d43475b" /> 

### Attachments
[comment]: # (Can you provide an example file to trigger the problem?)

### Tracebacks
[comment]: # (Add your backtrace below if you have one. If QMapshack crashes the fastest way to get help is a backtrace.
[...]

PR #1042, opened 2026-04-14
[QMS-1022] Fix: Wrong view active after restart

JFritzle:

QMS-#1022

### What you have done:

Save current active map view's unique key to config instead of current tab index.
When closing QMS, current active map view is now taken from current map list tab.
Current map list tab always corresponds to active map view,
whether map view or diary or project editor or ... tab is visible when closing QMS.

Compatibility to previously saved tab index is preserved.

1. Create several map views
2. Set one of these map view as current
3. Create diary and/or project editor and/or ... tabs
4. Arrange these tabs in any arbitrary order
5. Close and re-open QMS, regardless of whether last open tab was a map view or diary or project editor tab
6. See that previous current map view becomes current map view again

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/CMainWindow.cpp
+++ b/src/qmapshack/CMainWindow.cpp
@@ -295,7 +295,7 @@ CMainWindow::CMainWindow() : id(QRandomGenerator::global()->generate()) {
   actionProfileIsWindow->setChecked(cfg.value("profileIsWindow", false).toBool());
   actionLinkMapViews->setChecked(cfg.value("linkMapViews", false).toBool());
   mapFont = cfg.value("mapFont", font()).value<QFont>();
-  tabWidget->setCurrentIndex(cfg.value("visibleCanvas", 0).toInt());
+  tabWidget->setCurrentIndex(getTabIndexForCanvasKey(cfg.value("visibleCanvas", "").toString()));
   cfg.endGroup();  // Canvas
 
   QStatusBar* status = statusBar();
@@ -563,7 +563,8 @@ CMainWindow::~CMainWindow() {
 
   cfg.setValue("canvasOrder", allViewKeys);
   cfg.setValue("gisLayerOpacity", CCanvas::gisLayerOpacity);
-  cfg.setValue("visibleCanvas", tabWidget->currentIndex());
+  const CCanvas* canvas = getVisibleCanvas();
+  cfg.setValue("visibleCanvas", canvas != nullptr ? canvas->getKey() : "");
   cfg.setValue("isGeosearchVisible", actionGeoSearch->isChecked());
   cfg.setValue("isScaleVisible", actionShowScale->isChecked());
   cfg.setValue("isGridVisible", actionShowGrid->isChecked());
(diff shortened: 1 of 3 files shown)


src/qmapshack/CMainWindow.cpp
```diff
@@ -563,7 +571,13 @@ CMainWindow::~CMainWindow() {
...
   cfg.setValue("canvasOrder", allViewKeys);
   cfg.setValue("gisLayerOpacity", CCanvas::gisLayerOpacity);
-  cfg.setValue("visibleCanvas", tabWidget->currentIndex());
+  int index = tabMaps->currentIndex();
+  if (index < 0) {
+    cfg.setValue("visibleCanvas", 0);
+  } else {
+    CMapList* mapList = dynamic_cast<CMapList*>(tabMaps->widget(index));
+    cfg.setValue("visibleCanvas", mapList->getCanvasKey());
+  }
```
kiozen: A dynamic cast should be always checked for nullptr. If the index is out of bound QTabWidget::widget() should return a nullptr. 

```
  const int index = tabMaps->currentIndex();
  CMapList* mapList = dynamic_cast<CMapList*>(tabMaps->widget(index));
  if (mapList != nullptr) {
    cfg.setValue("visibleCanvas", mapList->getCanvasKey());
  } else {
    cfg.setValue("visibleCanvas", "");
  }
``

File: src/qmapshack/CMainWindow.cpp [hunk 2]
```diff
@@ -295,7 +295,15 @@ CMainWindow::CMainWindow() : id(QRandomGenerator::global()->generate()) {
...
-  tabWidget->setCurrentIndex(cfg.value("visibleCanvas", 0).toInt());
+  auto value = cfg.value("visibleCanvas", 0);
+  for (int i = 0; i < tabWidget->count(); i++) {
+    CCanvas* canvas = dynamic_cast<CCanvas*>(tabWidget->widget(i));
+    if (canvas->getKey() == value) {
+      value = i;
+      break;
+    }
+  }
+  tabWidget->setCurrentIndex(value.toInt());
```
kiozen: Working with a QVariant is a bit of overkill: This should be more straight forward:
```
  const QString& key = cfg.value("visibleCanvas", "").toString();
  for (int i = 0; i < tabWidget->count(); i++) {
    CCanvas* canvas = dynamic_cast<CCanvas*>(tabWidget->widget(i));
    if (canvas->getKey() == key) {
      tabWidget->setCurrentIndex(i);
      break;
    }
  }
``

**Review by kiozen — CHANGES_REQUESTED** [review]
Good idea to use the map tab widget to get the key of the last active canvas tab!

I just would suggest some simplification of the code ;)

File: src/qmapshack/CMainWindow.cpp (hunk 2 above)
Comment (JFritzle) [reply src/qmapshack/CMainWindow.cpp]: I encapsulated finding tab index for given unique map view key into a new function
 `int CMainWindow::getTabIndexForCanvasKey(const QString& key)`
returning tab index for given unique map view key, This will simplify above statements to 1 single statement:
`tabWidget->setCurrentIndex(getTabIndexForCanvasKey(cfg.value("visibleCanvas", "").toString()));`

This makes me thinking about why not replacing current function
`CCanvas* CMainWindow::getVisibleCanvas() const { return dynamic_cast<CCanvas*>(tabWidget->currentWidget()); }`
by
```
CCanvas* CMainWindow::getVisibleCanvas() const {
  int index = tabMaps->currentIndex();
  CMapList* mapList = dynamic_cast<CMapList*>(tabMaps->widget(index));
  index = mapList != nullptr ? getTabIndexForCanvasKey(mapList->getCanvasKey()) : -1;
  return dynamic_cast<CCanvas*>(tabWidget->widget(index));
}
```
This would always return map view canvas pointer of currently active map view whether view's canvas is visible or currently hidden by a diary or track editor or something else.
I did a short test and saw no unwanted side effects. But I may have overlooked something.

In addition, it would also simplify saving visible map view at QMS close:
`cfg.setValue("visibleCanvas", getVisibleCanvas()->getKey());`
Maybe checking `getVisibleCanvas()` for null pointer needed.

Before rebasing, I will await your reply.

kiozen: Yes make sense. Assuming that a key might not match or a dynamic cast might fail is mandatory and must be handled.

JFritzle:
Ok, rebased.
Function `CMainWindow::getVisibleCanvas()` now returns canvas pointer of current `tabMaps` map view.
A more suitable name for function would be `CMainWindow::getCurrentCanvas()`.
But this requires function call to be renamed in a lot of files.

PR #1046, opened 2026-04-17
Fix build failure with GDAL 3.13.0.

sebastic:
1.20.1 fails to build with GDAL 3.13.0 due to API changes:
```
error: 'MIN' was not declared in this scope
error: 'MAX' was not declared in this scope
```
See: https://github.com/OSGeo/gdal/commit/1f41ebcb4eaaad8ad5d81a789bf03ce9c20b3beb#diff-5e4656f11ffb922553c3e06f51ce9666886da91cfbc868ba4eac4c7515013c21

### DIFF
--- a/src/qmt_map2jnx/main.cpp
+++ b/src/qmt_map2jnx/main.cpp
@@ -430,7 +430,11 @@ static void printProgress(int current, int total) {
   static int nLastTick = -1;
   int nThisTick = (int)(dfComplete * 40.0);
 
+#if GDAL_VERSION_NUM >= GDAL_COMPUTE_VERSION(3, 13, 0)
+  nThisTick = CPL_MIN(40, CPL_MAX(0, nThisTick));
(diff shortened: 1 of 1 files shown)


kiozen:
Can you please open a ticket (bug report).  Use comment of this PR as content. Add the ticket number to the commit message (see pattern of other commits) and add the line to the change log (keep order of tickets).  Thx.

sebastic:
I have to test and fix a lot of GDAL reverse dependencies, I don't have time to do all the extra work in your PR workflow.

kiozen:
> I don't have time to do all the extra work in your PR workflow.

This statement holds true for everyone committing to QMapShack. And still the workflow is kept. 

And an incompatibility to a GDAL version is worth to be mentioned and documented properly.

sebastic:
No time to follow the desired workflow.

Issue #1079, opened 2026-05-17
DEMs and POIs are not restored on startup after upgrade (regression since commit 38cd1fc9)


Phobooky:
### Describe the bug

After upgrading to a build that includes commit 38cd1fc9 (QMS-1071), all previously
configured DEM and POI files appear in their respective lists but none are activated
on startup. The files and paths are correct — the problem is that the stored
configuration keys no longer match the keys computed by the new binary.

### What have you done to circle down the problem?

I compiled QMapShack from source (Qt 6, Windows) and compared behavior between a
build before and after commit 38cd1fc9. I confirmed that:

- The settings written by the old build store MD5 keys based on the first **1024 bytes**
  of each file.
- The new build computes MD5 keys from the first **4096 bytes**.
- In `CDemDraw::loadDemList()`, the stored key is not found in `demsFound`, the fallback
  creates a new item whose key (4096-byte hash) does not match the stored key
  (1024-byte hash), so the condition `key != dem->getKey()` triggers and the item is
  deleted immediately — the DEM is lost.
- Same issue in `CPoiDraw::restoreActivePoisList()`: the stored key never matches
  `item->key`, so no POI is ever activated.
- Maps are **not** affected because their hash was already effectively 4096 bytes
  (`qMin(0x1000LL, f.size())` = 4096 = 0x1000), so the change was a no-op for maps.

I also verified that copying the old executable (pre-38cd1fc9) back restores the
correct behavior, confirming the regression is in the binary and not in the environment.

### To Reproduce

1. Configure and activate at least one DEM file and one POI file in QMapShack.
2. Replace the executable with a build that includes commit 38cd1fc9 or later
   (keeping all other files and settings unchanged).
3. Start QMapShack.
4. Open the DEM list and the POI list — all items appear as "Unused", none are active.

### Expected behavior

DEMs and POIs previously configured as active should be restored to their active state
on startup, as they were before commit 38cd1fc9.

### Screenshots

N/A

### Attachments

N/A

### Tracebacks

No crash. The issue is silent — items silently fail to activate.

### Desktop

 - OS: Windows 11 Pro
 - QMapShack Version: 1.20.2 (compiled from source, branch `dev`, post-38cd1fc9)
 - Window Manager: N/A (Windows)

### Additional context

I have identified and tested a fix. I am happy to submit a Pull Request if desired.

**Fix summary:**

For DEMs (`CDemDraw::loadDemList`): when a stored key has no match in `demsFound`,
[...]

PR #1082, opened 2026-05-19
[QMS-1079] Fix/dem poi hash backward compatibility

Phobooky:

QMS-#1079

### What you have done:

Commit 38cd1fc9 (QMS-1071) changed the MD5 hash used to identify DEM and POI files
from the first 1024 bytes to the first 4096 bytes. This broke backward compatibility:
settings written by any previous build store the old 1024-byte hash as the config key.
The new binary computes a different key, so all DEMs and POIs appear as "Unused" on
startup even though files and paths are correct.

Maps are not affected (their hash was already effectively 4096 bytes via `qMin(0x1000LL, f.size())`).

Fix: each item now computes both the short (1024-byte) and long (4096-byte) hash in a
single file read. When restoring from config, the stored key is compared against both.
If it matches the short key, it is used temporarily to read settings from the correct
config group, then the long key is restored immediately — so the next save automatically
migrates to the new key.

1. Configure and activate at least one DEM and one POI file with a previous build of QMapShack (pre QMS-1071), so their keys are stored in the config.
2. Replace the executable with this build.
3. Start QMapShack and open the DEM list and POI list.
4. Verify that previously active DEMs and POIs are restored correctly.
5. Restart QMapShack a second time and verify they are still active (migration to new key completed).

- [x] yes

- [x] yes (no user-facing strings added)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,7 +7,9 @@ V1.XX.X
 [QMS-1071] Fix: Valid DEM file may not be loaded
 [QMS-1075] Disable "O", "V" and "T" buttons when drawing a route
 [QMS-1077] Fix: Restoring the workspace crashes if database projects are loaded but their database is missing
+[QMS-1079] Fix: DEM/POI hash backward compatibility
 [QMS-1080] Add *.ts file for Croatian translation
+
 
 V1.20.2
 [QMS-670] Fix: Accessibility issue : font is too small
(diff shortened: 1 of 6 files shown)


kiozen:
I took the freedom to clean up the commit history and a to apply a few changes. Nothing serious, just a minor clean up. Can you please test again to make sure I didn't break anything.

This should update your branch:
```
git pull
git reset --hard origin/fix/dem-poi-hash-backward-compat
```

Phobooky:
I have updated my local branch. I will compile and test now.

Phobooky:
Tested successfully. DEMs and POIs are restored correctly on startup. 
Everything looks good.

Issue #1090, opened 2026-05-28
More project items of delegate entries possibly hardly readable


JFritzle:
### Describe the bug
[comment]: # (A clear and concise description of what the bug is)

While https://github.com/Maproom/qmapshack/issues/1016 and https://github.com/Maproom/qmapshack/pull/1085 fixed bad readability of selected activate/visible map/dem/workspace items after losing input focus, **selected project items** (tracks, routes, waypoints) of an **inactive/invisible project** are still hardly readable after losing focus.

This is particularly the case on platform **Windows** with default color scheme, where background of selected items changes to light gray after losing input focus.

### What have you done to circle down the problem?
[comment]: # (Did you compare to previous versions? What versions? Did you check if others have the same problem?)

### To Reproduce
[comment]: # (List the steps to reproduce the behavior)

1. Run QMS on platform Windows
2. Set a workspace project containing project items to invisible
3. Expand project
4. Select one of it's project items -> item's background becomes blue, foreground becomes white: good readability
5. Click into map canvas -> input focus changes from item to map canvas
6. Project item has lost input focus -> item's background becomes light gray, foreground stays white: bad readability

### Expected behavior
[comment]: # (A clear and concise description of what you expected to happen)

Selected items should still be readable after losing input focus.

### Screenshots
[comment]: # (If applicable, add screenshots to help explain your problem)

Currently bad readability of a selected but set to invisible project item after losing focus:
  <img width="297" height="138" alt="Image" src="https://github.com/user-attachments/assets/d670598c-f457-4b94-95d4-a6daf4af3058" />
 
Expected good readability even after losing fous:
   <img width="298" height="137" alt="Image" src="https://github.com/user-attachments/assets/0948da7c-5388-4c83-a56d-fc1dde803cf3" />

Screenshots taken with Qt style _Fusion_.

### Attachments
[comment]: # (Can you provide an example file to trigger the problem?)

### Tracebacks
[comment]: # (Add your backtrace below if you have one. If QMapshack crashes the fastest way to get help is a backtrace.
For Linux see: https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux.
For Windows it would need a debug build and running QMapshack in Visual Studio.)

### Desktop
[comment]: # (please complete the following information)

 - OS: primarily **Windows**, but may also affect other operating systems depending on color scheme used
 - QMapShack Version:   current
 - Window Manager:  -

### Additional context
[...]

PR #1091, opened 2026-05-28
[QMS-1090] Fix: More project items of delegate entries possibly hardly readable

JFritzle:

QMS-#1090

### What you have done:

Fixed color role of selected project items contained in an inactive/invisible workspace project when input focus has been lost.

1. Run QMS on platform Windows
2. Set a workspace project containing project items to invisible
3. Expand project
4. Select one of it's project items -> item's background becomes blue, foreground becomes white: good readability
5. Click into map canvas -> input focus changes from item to map canvas
6. Project item has lost input focus -> item's background becomes light gray, foreground becomes darker gray: **good readability**

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,6 +11,7 @@ V1.XX.X
 [QMS-1080] Add *.ts file for Croatian translation
 [QMS-1084] Fix: More delegate entries possibly hardly readable
 [QMS-1088] Fix: Mysterious debug output
+[QMS-1090] Fix: More project items of delegate entries possibly hardly readable
 
 V1.20.2
 [QMS-670] Fix: Accessibility issue : font is too small
--- a/src/qmapshack/gis/CWksItemDelegate.cpp
+++ b/src/qmapshack/gis/CWksItemDelegate.cpp
@@ -548,10 +548,13 @@ void CWksItemDelegate::paintDevice(QPainter* p, const QStyleOptionViewItem& opt,
       getRectanglesDevice(opt, item);
 
   const bool isVisible = item.isVisible();
-  QPalette::ColorGroup colorGroup = (opt.state & QStyle::State_HasFocus) ? QPalette::Active : QPalette::Inactive;
-  const QColor& colorName =
-      opt.palette.color(isVisible ? colorGroup : QPalette::Disabled,
-                        opt.state & QStyle::State_Selected ? QPalette::HighlightedText : QPalette::WindowText);
+  const bool isSelected = (opt.state & QStyle::State_Selected) != 0;
+  const bool hasFocus = (opt.state & QStyle::State_HasFocus) != 0;
+
+  // derive strings colors
+  const QPalette::ColorRole colorRole = (isSelected && hasFocus) ? QPalette::HighlightedText : QPalette::WindowText;
+  const QPalette::ColorGroup colorGroup = isVisible ? (hasFocus ? QPalette::Active : QPalette::Inactive) : QPalette::Disabled;
+  const QColor& colorName = opt.palette.color(colorGroup, colorRole);
 
   // draw name
   fontName.setBold(item.hasUserFocus());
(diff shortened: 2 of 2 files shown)


kiozen:
This will be an ever moving target :grin: I really wonder how the code should ever be fit for all themes. Little fun fact on my system:

grafik

Hard to see anything for selected invisible items. But this is not an issue of this patch. It's badly visible even on the develop branch. It's just an example that there will always an example of bad rendering based on the scheme. And there are tons of schemes out there.

JFritzle:
About https://github.com/Maproom/qmapshack/pull/1091#issuecomment-4571556755:

In my opinion there are currently two color problems with delegate items for map, workspace and database.

Problem 1 - color group:
For invisible/inactive items, we choose foreground color = text color from palette's color group `QPalette::Disabled`. But it turns out that `opt.palette.currentColorGroup()` never returns `QPalette::Disabled` but either `QPalette::Active` or `QPalette::Inactive`, depending on input focus = `opt.state & QStyle::State_HasFocus` value. While `QStyledItemDelegate::paint(p, opt, index);` chooses a background according to _current_ color group, we choose foreground color from an _expected_ but different color group. This could be solved by replacing
```
void CWksItemDelegate::paint(QPainter* p, const QStyleOptionViewItem& opt, const QModelIndex& index) const {
  IWksItem* item = indexToItem(index);
  if (item == nullptr) {
    return QStyledItemDelegate::paint(p, opt, index);
  }

  p->save();
  USE_ANTI_ALIASING(*p, true);
  // Standard background & selection
  QStyledItemDelegate::paint(p, opt, index);
  ...
```
using QStyleOptionViewItem's copy constructor and now being able to override color group as _expected_ by
```
void CWksItemDelegate::paint(QPainter* p, const QStyleOptionViewItem& option, const QModelIndex& index) const {
  QStyleOptionViewItem opt = option;
  IWksItem* item = indexToItem(index);
  if (item == nullptr) {
    return QStyledItemDelegate::paint(p, opt, index);
  }

  p->save();
  USE_ANTI_ALIASING(*p, true);
  
  const bool isVisible = item->isVisible();
  const bool hasFocus = (opt.state & QStyle::State_HasFocus) != 0;
  const QPalette::ColorGroup colorGroup = isVisible ? (hasFocus ? QPalette::Active : QPalette::Inactive) : QPalette::Disabled;
  opt.palette.setCurrentColorGroup(colorGroup);

  // Standard background & selection
  QStyledItemDelegate::paint(p, opt, index);
  ...
```
Calling function `opt.palette.currentColorGroup();` after `QStyledItemDelegate::paint(p, opt, index);` now returns _expected_ color group.
If current color group is the one we use for choosing text color according to color role, later on we can simply replace
```
  const QColor& colorName = opt.palette.color(colorGroup, colorRole);
```
by
```
  const QColor& colorName = opt.palette.color(colorRole);
```

Problem 2 - background vs foreground colors:
While we choose foreground = text color by color role according to `opt.state & QStyle::State_Selected` value, function `QStyledItemDelegate::paint(p, opt, index);` chooses background color according to unpredictable internal rules and fills item's rect. I tried to find out which color without success. It is even not guaranteed that fill color is one of current color group's predefined palette colors.
If we do not rely on unpredictable background color we can make sure to have a good contrast between background and foreground. This can be achieved by inserting before `QStyledItemDelegate::paint(p, opt, index);`:
```
  const bool isSelected = (opt.state & QStyle::State_Selected) != 0;
  if (isSelected) {
    const QBrush& backgroundBrush = hasFocus ? opt.palette.highlight() : opt.palette.base();
    p->fillRect(opt.rect, backgroundBrush);
  }
```
Now all brushes are taken from current color group, same color group as color role.
Perhaps a different palette brush than base() may suit better.

For testing purpose, I attach my modified but not perfect _CWksItemDelegate.cpp_ file: [CWksItemDelegate.zip](https://github.com/user-attachments/files/28421864/CWksItemDelegate.zip)

Issue #1093, opened 2026-05-31
Track segment's endpoint relocates after left mouse click fix


JFritzle:
### Describe the bug
[comment]: # (A clear and concise description of what the bug is)

Despite fixing the endpoint of a track segment with a left mouse click, endpoint _may_ still relocate afterwards.

### What have you done to circle down the problem?
[comment]: # (Did you compare to previous versions? What versions? Did you check if others have the same problem?)

### To Reproduce
[comment]: # (List the steps to reproduce the behavior)

1. Using auto-routing, generate a long distance and time consuming track segment from point A to point B
2. Fix endpoint B by left-click -> segment from point A to point B is recalculated by router and redrawn
3. Left-click next endpoint C _before_ redraw of segment from point A to point B has finished
    (Redraw has finished when endpoint B becomes marked by a green-bordered square dot.)
4. Endpoint B previously fixed by left mouse click relocates to a new position B' towards point C
    and redrawn segment from point A to point B is extended by a straight line
    from endpoint B position to relocated endpoint B' position. 

### Expected behavior
[comment]: # (A clear and concise description of what you expected to happen)

I expect track segment's endpoint not relocating anymore after being fixed by left-click.

### Screenshots
[comment]: # (If applicable, add screenshots to help explain your problem)

To generate track, points marked by digits 1 to 6 have been left-clicked.
While point 5 has not been marked by black square as waypoint (= track segment endpoint) although left-clicked, point ? is being marked as waypoint although not left-clicked.

<img width="903" height="627" alt="Image" src="https://github.com/user-attachments/assets/28636190-c66f-44f7-b48c-5ab07ec829c3" />
 
### Attachments
[comment]: # (Can you provide an example file to trigger the problem?)

### Tracebacks
[comment]: # (Add your backtrace below if you have one. If QMapshack crashes the fastest way to get help is a backtrace.
For Linux see: https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux.
For Windows it would need a debug build and running QMapshack in Visual Studio.)

### Desktop
[comment]: # (please complete the following information)

 - OS: any
 - QMapShack Version:   current
 - Window Manager:  -

### Additional context
[comment]: # (Add any other context about the problem here)

PR #1094, opened 2026-05-31
[QMS-1093] Fix: Auto-routed track segment endpoint jumps when next point is clicked before routing finishes

kiozen:

QMS-#1093

### What you have done:

Root cause: When the user left-clicks to fix point B while auto-routing is active, slotTimeoutRouting() calls calcRoute() which spins a nested Qt event loop (progress dialog). During that event loop, mouseMove events update points[idxFocus].coord to wherever the mouse drifts (toward C). When routing finishes, the coord has moved to B' — so storeToHistory saves the wrong position, causing the endpoint to jump.

  Three-part fix applied to both CLineOpAddPoint and CLineOpMovePoint:

  1. Pin before routing — set points[idxFocus].coord to the clicked position right before calling slotTimeoutRouting(), so the router uses the correct target even if mouseMove fires during the event loop
  2. Restore after routing — reset points[idxFocus].coord back to the clicked coordinate after slotTimeoutRouting() returns, undoing any drift
  3. Re-entrancy guard — check isRouting at the top of leftClick() to silently drop any click that arrives during the nested event loop (added to ILineOp as a shared protected member)

See ticket

- [x] yes

- [x] yes

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,112 @@
+# QMapShack — Claude project notes
+
+QMapShack is a Qt/C++ desktop application for planning and analysing GPS tracks, routes, and waypoints. It loads map tiles, vector maps, and DEM data, and supports online and offline routing engines.
+
+---
+
+## Code style
+
+**All C++ is formatted with clang-format.** After editing any `.cpp` or `.h` file run:
+
+```bash
+clang-format -i <file> [<file> ...]
+```
+
+The style is defined in `.clang-format` in the project root (Google base, 120-column limit).
+
+**Every control-flow block requires braces**, even single-liner bodies:
+
+```cpp
+// correct
+if (condition) {
+  doSomething();
+}
+
+// wrong — never do this
+if (condition)
+  doSomething();
+```
+
+**Documentation comments use `/** */` doxygen blocks.** Follow the style of existing blocks (`@brief`, `@param`, `@return`). Inline member docs use `/**< */`. Plain `//` comments are for non-doxygen annotations.
+
+---
+
+## Building
+
+```bash
+cmake --build build --target qmapshack -j$(nproc)
+```
+
+The build directory is `build/` in the project root.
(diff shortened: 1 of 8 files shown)


kiozen:
@JFritzle please test

JFritzle:
Track segment endpoint no longer jumping. 👍

It _appears_ that this PR also resolves the issue where track segments could be generated as straight lines, despite the auto-routing feature being enabled. At least during my tests, I never encountered any straight lines. 👍👍

kiozen:
Hm, don't know. Looking at the root cause and the fix I would say it's unrelated. But this is a very complex topic and it's hard to comprehend everything. In fact I solved the last two bugs with AI - the good way. 

Some interesting facts about this:

I fed  your tickets into the AI as a start and expected to give it some guidance and hints. Amazingly enough it analyzed the ticket, jumped right at the code and came up with a pretty understandable root cause analyzes.  Imho your brilliant way to write very good tickets played a major part in this.

My next step was to request a suggestion for a fix. The suggested code is in the 1st commit of this PR. It's  a minimum impact fix, easy to understand. The final version in  this PR had 2 iterations to simplify and cleanup the code. 

Compared to the past AI approach to fix that stuff, introducing buffers and caches and all kind of other peculiar code, this was really a huge help. I can't tell why. Maybe because the used AI (Claude Code) is better. Maybe because the ticket was in place. Maybe because I applied more control in the review of the code. I can't tell. 

Anyway: This is the way AI adds value to a project.

The other two commits are just a general code cleanup and documentation.  The first commit is the gold.

After fixing the bug I became venturesome and asked the AI to fix the issue with this stupid event loop in the progress bar for good,  introducing threaded code. This had quite an impact and the final result had several glitches. And it it was late in the evening, so I aborted the experiment. 

Maybe I pick it up again the next days. On the other hand side it's working now and I already wasted enough time on this topic.

Issue #1095, opened 2026-06-01
Invalid track ascent/descent at track creation


JFritzle:
### Describe the bug
[comment]: # (A clear and concise description of what the bug is)

Issue 1:
When creating a track where only parts of track are covered by elevation (DEM) data, _Track Metrics_ box shows huge and invalid ascent and/or descent values.

Issue 2:
After saving track, workspace project as well as project's track item show these invalid values too.

### What have you done to circle down the problem?
[comment]: # (Did you compare to previous versions? What versions? Did you check if others have the same problem?)

### To Reproduce
[comment]: # (List the steps to reproduce the behavior)

1. Creating a track where only parts of track are covered by elevation (DEM) data
2. See huge invalid ascent/descent values in _Track Metrics_ box
3. Save track in project
4. See same huge invalid ascent/descent values in project
5. See same huge invalid ascent/descent values in project's track item

### Expected behavior
[comment]: # (A clear and concise description of what you expected to happen)

I expect a hint about missing/invalid DEM data instead of displaying invalid values. 

### Screenshots
[comment]: # (If applicable, add screenshots to help explain your problem)

DEM data applied to map view is Sony's GeoTIFF _Germany_.
DEM data covers whole Germany but ends at German/Polish border.
Track is crossing German/Polish border multiple times.
Each time border is crossed, a huge value is added to ascent and/or descent. 

<img width="340" height="394" alt="Image" src="https://github.com/user-attachments/assets/422205ef-ec90-411e-9165-be469da8ad40" />

After saving track, workspace project as well as project's track item show these invalid values too,
while tooltip correctly reports _Invalid elevations!_:

<img width="312" height="241" alt="Image" src="https://github.com/user-attachments/assets/05ac319a-e520-4ddb-9715-1c76e9af06c3" />

### Attachments
[comment]: # (Can you provide an example file to trigger the problem?)

### Tracebacks
[comment]: # (Add your backtrace below if you have one. If QMapshack crashes the fastest way to get help is a backtrace.
For Linux see: https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux.
For Windows it would need a debug build and running QMapshack in Visual Studio.)

### Desktop
[comment]: # (please complete the following information)

 - OS: any
 - QMapShack Version:   current
 - Window Manager:  -

### Additional context
[comment]: # (Add any other context about the problem here)

I will commit this fix for issue 1:

[...]

PR #1096, opened 2026-06-01
[QMS-1095] Fix: Invalid track ascent/descent at track creation (issue 1)

JFritzle:

QMS-#1095

### What you have done:

Report text string **Invalid elevations!** instead of invalid ascent/descent values in _Track Metrics_ box (issue 1 of #1095).

1. Create a track where only parts of track are covered by elevation (DEM) data
2. See **Invalid elevations!** instead of invalid ascent/descent values in _Track Metrics_ box

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/locale/qmapshack.ts
+++ b/src/qmapshack/locale/qmapshack.ts
@@ -393,79 +393,79 @@
 <context>
     <name>CDBItemDelegate</name>
     <message>
-        <location filename="../gis/CDBItemDelegate.cpp" line="329"/>
-        <location filename="../gis/CDBItemDelegate.cpp" line="331"/>
-        <location filename="../gis/CDBItemDelegate.cpp" line="333"/>
+        <location filename="../gis/CDBItemDelegate.cpp" line="332"/>
+        <location filename="../gis/CDBItemDelegate.cpp" line="334"/>
+        <location filename="../gis/CDBItemDelegate.cpp" line="336"/>
         <source>since: </source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../gis/CDBItemDelegate.cpp" line="329"/>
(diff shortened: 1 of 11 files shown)


kiozen:
Will this PR fix issue 2, too. Or will this be another PR. In other words: Ready to merge?

JFritzle:
I tried to fix issue 2 in file _CWksItemDelegate.cpp_ too. Without success.
_CWksItemDelegate.cpp_ gets ascent/descent by {project|trk}->getTotal{Ascent|Descent} and does some checks (which I do not understand) to decide if ascent/descent should be shown or not. 
To fix issue 2, function getTotal{Ascent|Descent} should report invalid elevations.
-> To fix issue 2 too, another PR will be needed. Ready to merge for issue 1 fix.

kiozen:
```
    qreal ascent = project->getTotalAscent();
    if (ascent < NOFLOAT && !qFuzzyIsNull(ascent) && itemStatusControl.prj.ascent) {
      QString unit, val;
      IUnit::self().meter2elevation(ascent, val, unit);
      status += QString("%1%2%3 ").arg(QChar(0x2197)).arg(val, unit);
    }
```
It checks for the value to be less than NOFLOAT in the assumption that any value less than NOFLOAT is a valid value. However the mixed sum of valid and in valid elevations seems to be something less but still bogus. Probably the best fix is to set the total values to either 0 or NOFLOAT the moment invalid elevations are detected.

JFritzle:
... or to return a certain negative value, let's say -1., to indicate invalid elevations are detected since valid ascent/descent values are always ≥ 0.

Thus we can still distinguish between NOFLOAT, i.e. no elevation data at all, and -1, i.e. invalid elevations are detected.
And we are able to skip output of ascent/descent in case of NOFLOAT or report string _Invalid elevations!._  if either total ascend and/or total descend is -1.

Do you agree?

Issue #1097, opened 2026-06-02
Incomplete routing when editing a track points on a faster mouse movement speed


kkarsten62:
### Describe the bug
This is a bug report related to Issue [#1093](https://github.com/Maproom/qmapshack/issues/1093) and pull request [#1094](https://github.com/Maproom/qmapshack/pull/1094).

In edit track point mode:
If you modify an existing track point or add a new track point between two existing track points, incorrect routing may occur if you move the mouse slightly faster when clicking for the new track point position.

### What have you done to circle down the problem?
This issue has occurred in previous versions of QMS.
The fix [#1094](https://github.com/Maproom/qmapshack/pull/1094) ensures that at least one new track path is routed correctly. However, the second track path is a straight line.

**Example on editing a track**
<img width="1810" height="1216" alt="Image" src="https://github.com/user-attachments/assets/0274f1ea-52be-4d5c-a1a6-21d40eacfe21" />

**Slower (normal) mouse move speed when picking the new track point position ==> OK**
<img width="1810" height="1216" alt="Image" src="https://github.com/user-attachments/assets/a5de396c-19e1-41f1-a2c7-44179ad2747e" />

**Faster mouse move speed when picking the new track point position ==> NOT OK**
<img width="1810" height="1216" alt="Image" src="https://github.com/user-attachments/assets/7e5ca15a-2fa8-4e33-a650-4a6aca492661" />

### To Reproduce
1. Create a new track
2. Select start and endpoint
3. Right mouse click
4. Choose `Add point (+)` 
5. Select track line 
6. Try to pick the new track point position with a slow, normal and faster mouse movement speed
7. Check the results
 
### Screenshots
See above

### Attachments
NA

### Tracebacks
NA

### Desktop
 - OS: Linux Fedora 44 GNOME Wayland
 - QMapShack Version:   latest development commit
 - Window Manager:  GNOME

### Additional context
NA

PR #1099, opened 2026-06-03
[QMS-1097] Fix: Incomplete routing when editing a track points on a f…

kiozen:
…aster mouse movement speed

QMS-#1097

### What you have done:

Root cause: finalizeOperation calls tryRouting(idx-1) first. That call runs calcRoute, which shows a progress dialog that spins a nested Qt event loop. During that event loop, mouseMove fires (fast mouse) and overwrites  points[idx].coord with the current cursor position. When tryRouting(idx) is then called, it snapshots the now-drifted coord as its start point —routing from the wrong location, producing a straight line or wrong path.

Fix: Snapshot points[idx].coord before entering the first event loop, then restore it before the second tryRouting call. The three-line addition is minimal and doesn't affect the abort path (if the user right-clicks during the first routing, points is reallocated and the size check on idx < points.size() prevents an out-of-bounds write before the restore, and the second tryRouting is skipped by the idx < points.size() - 1 guard).

The fix applies equally to both CLineOpAddPoint and CLineOpMovePoint since they both go through finalizeOperation.

See ticket

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,6 +15,7 @@ V1.XX.X
 [QMS-1090] Fix: More project items of delegate entries possibly hardly readable
 [QMS-1093] Fix: Auto-routed track segment endpoint jumps when next point is clicked before routing finishes
 [QMS-1095] Fix: Invalid track ascent/descent at track creation
+[QMS-1097] Fix: Incomplete routing when editing track points on a faster mouse movement speed
 
 V1.20.2
 [QMS-670] Fix: Accessibility issue : font is too small
--- a/src/qmapshack/mouse/line/CLineOpAddPoint.cpp
+++ b/src/qmapshack/mouse/line/CLineOpAddPoint.cpp
@@ -103,6 +103,9 @@ void CLineOpAddPoint::leftClick(const QPoint& pos) {
 void CLineOpAddPoint::mouseMove(const QPoint& pos) {
   ILineOp::mouseMove(pos);
   if (isDragging) {
+    if (isRouting) {
+      return;
+    }
     QPointF coord = pos;
     gis->convertPx2Rad(coord);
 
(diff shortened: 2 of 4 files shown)


kiozen:
@kkarsten62 @JFritzle please test

kkarsten62:
I compiled the branch QMS-1097 from your repo and did some tests.

Overall, it feels better, and I’ve been making fewer mistakes. However, in some cases, I can still trigger the issue by moving the mouse a little bit faster while releasing the left mouse button at the same time.

image

kiozen:
Is the progress dialog popping up when this happens?

kkarsten62:
> Is the progress dialog popping up when this happens?

No, a progress dialog does not appear.

There are three possible scenarios when you move the mouse and release the left mouse button at the same time.
1. The track is routed correctly
2. The point is not accepted, and you can continue dragging the point with the mouse
3. An incorrect track path, as shown in the image above

JFritzle:
When inserting an intermediate point C into track segment from A to B, connection from A to C still might be a straight line.
To reproduce:

1. Switch _Add new points_ on
2. Drag an intermediate point to it's new position C
3. Release left mouse button and move mouse **immediately** to a different position
4. Point C does not jump but line A to C is a straight line
5. Progress dialog has not popped up

grafik

kiozen:
next try, please.

kkarsten62:
> next try, please.

After next test, three scenarios:
1. Move mouse to the new position, wait for routing, click left mouse button ==> OK
2. **Slow** moving the mouse to the new position and clicking the left mouse button while moving ==> routing as expected, OK
3. **Fast** moving the mouse to the new position and clicking the left mouse button while moving ==> NO routing, still in dragging mode, OK

Thanks @kiozen, seems to be fixed :-)

JFritzle:
Ok on my end. 👍

kiozen:
Hallelujah, hope this is the last one.

Issue #1115, opened 2026-06-12
Improve performance for high resolution DEMs


jmlzr:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

Currently QMS does the various shadings for DEMs at the full resolution of the DEM and only afterwards scales it to the view. This makes using high resolution DEMs extremely slow, especially when zoomed out.
Even with a decent PC a DEM with 1m horizontal resolution is unstably slow at anything but the closest zoom levels.

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

Scale the DEM data before computing the shadings and take advantage of available overviews.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

N/A

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

N/A; I'm working on this myself, just filing this issue to satisfy the "commit guide".

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

PR #1116, opened 2026-06-12
[QMS-1115] Improve performance of DEM rendering

jmlzr:

QMS-#1115

### What you have done:

Rewrite most of CDemVRT::draw to dramatically speed up rendering of DEMs.

Calculate an appropriate (down-)scale factor from the DEM according to current zoom level. Then use GDALs ReadRaster function to retrieve the data already scaled which will also automatically use overviews if available.
Read all of the data in one call to make things simpler and more efficient as reading them in chunks inside multiple threads was pointless since GDAL is not threadsafe. Then calculate the shadings in parallel on chunks of the data.
Also fix prev incorrect datatypes in the signatures of shading/coloring functions.

With those changes even very high resolution DEMs will be really fast now, especially when they have overviews.

1. Load a DEM (with very high resolution and overviews)
2. Activate hillshading, slopeshading, ...
3. Enjoy

- [x] yes

*Note:* There are no configs for clang-tidy and clazy provided making usage of those tools pretty much pointless.

- [x] yes

### DIFF
--- a/src/qmapshack/dem/CDemVRT.cpp
+++ b/src/qmapshack/dem/CDemVRT.cpp
@@ -19,7 +19,9 @@
 
 #include "dem/CDemVRT.h"
 
+#include <gdal.h>
 #include <gdal_priv.h>
+#include <gdalwarper.h>
 
 #include <QtWidgets>
 
@@ -91,12 +93,60 @@ CDemVRT::CDemVRT(const QString& filename, CDemDraw* parent) : IDem(parent), file
   noData = pBand->GetNoDataValue(&hasNoData);
   qDebug() << "no data:" << hasNoData << noData;
 
+  // ------- setup warped VRT ---------------
+  // if the projection of the dataset is different then that of the drawing context we wrap the dataset in a virtual
+  // warped dataset to transparently resample it into the drawing contexts projection.
+  OGRSpatialReference targetSRS;
+  OGRErr rv = targetSRS.SetFromUserInput(dem->getProjection().toUtf8());
+  const OGRSpatialReference* sourceSRS = dataset->GetSpatialRef();
+  if (rv == OGRERR_NONE && sourceSRS != nullptr && !sourceSRS->IsSame(&targetSRS)) {
+    srcDataset = dataset;
+
+    GDALWarpOptions* psOptions = GDALCreateWarpOptions();
+    psOptions->pProgressArg = dem;
+    psOptions->pfnProgress = [](double dfc, const char* msg, void* arg) -> int {
+      auto dem = reinterpret_cast<CDemDraw*>(arg);
+      return !dem->needsRedraw();
+    };
+
+    dataset = GDALDataset::FromHandle(GDALAutoCreateWarpedVRT(
+        GDALDataset::ToHandle(srcDataset), nullptr, targetSRS.exportToWkt().c_str(), GRA_Bilinear, 0.1, psOptions));
+
+    GDALDestroyWarpOptions(psOptions);
+
+    if (dataset == nullptr) {
+      GDALClose(srcDataset);
+      srcDataset = nullptr;
+      QMessageBox::warning(CMainWindow::getBestWidgetForParent(), tr("Error..."),
+                           tr("Failed to create Warp for:") % '\n' % filename);
+      return;
+    }
+
+    if (hasOverviews) {
+      // to make GDAL take advantage of overviews in the original dataset we need to add "virtual overviews" with the
+      // same decimation factors to the WarpedVRT so first build a list of the overviews the original dataset
+      // contains...
+      QVector<qint32> overviews(pBand->GetOverviewCount());
+      for (int i = 0; i < overviews.size(); ++i) {
+        GDALRasterBand* overview = pBand->GetOverview(i);
+        qreal decimationFactor = (qreal)pBand->GetXSize() / overview->GetXSize();
+        overviews[i] = qRound(decimationFactor);
+      }
+      // ...and attach them as virtual overview
+      CPLSetConfigOption("VRT_VIRTUAL_OVERVIEWS", "YES");
+      dataset->BuildOverviews("NONE", overviews.size(), overviews.data(), 0, nullptr, nullptr, nullptr);
+      CPLSetConfigOption("VRT_VIRTUAL_OVERVIEWS", "NO");
+    }
+  }
+
   // ------- setup projection ---------------
   proj.init(dataset->GetProjectionRef(), "EPSG:4326");
 
   if (!proj.isValid()) {
     GDALClose(dataset);
     dataset = nullptr;
+    GDALClose(srcDataset);
+    srcDataset = nullptr;
     QMessageBox::warning(CMainWindow::getBestWidgetForParent(), tr("Error..."),
                          tr("No georeference information found:") % '\n' % filename);
     return;
@@ -146,6 +196,7 @@ CDemVRT::CDemVRT(const QString& filename, CDemDraw* parent) : IDem(parent), file
 CDemVRT::~CDemVRT() {
   threadPool.waitForDone();
   GDALClose(dataset);
+  GDALClose(srcDataset);
 }
 
 void CDemVRT::slotNeedsRedraw() { threadPool.clear(); }
@@ -322,6 +373,17 @@ void CDemVRT::draw(IDrawContext::buffer_t& buf) {
   QVector<float> data(static_cast<qsizetype>(w_buf) * h_buf);
   {
     QMutexLocker lock(&mutex);
+
+    // add a temporary GDAL error handler to filter out errors that are caused by intentionally aborted reads
+    // automatically removed at end of scope
+    CPLErrorHandlerPusher oCurrentHandler(
+        [](CPLErr eErrClass, CPLErrorNum errNo, const char* msg) {
+          if (errNo == CPLE_UserInterrupt || errNo == CPLE_AppDefined) {
+            return;
+          }
+          CPLDefaultErrorHandler(eErrClass, errNo, msg);
+        },
+        nullptr);
 
     // by requesting a different size than the size of the buffer GDAL will automatically do scaling for us and use
     // overviews
--- a/src/qmapshack/dem/CDemVRT.cpp
+++ b/src/qmapshack/dem/CDemVRT.cpp
@@ -243,11 +243,18 @@ void CDemVRT::draw(IDrawContext::buffer_t& buf) {
     return;
   }
 
-  // get pixel offset of top left buffer corner
-  QPointF pp = buf.ref1;
-  dem->convertRad2Px(pp);
+  // use bufferScale (and therefore the zoom level) and the pixel scale of the DEM to caluclate a downsampling factor
+  qreal buf_scale_x = qAbs(bufferScale.x() / xscale);
+  qreal buf_scale_y = qAbs(bufferScale.y() / yscale);
+  // <1 would mean GDAL does upscaling which is pointless
+  if (buf_scale_x < 1.0) {
+    buf_scale_x = 1.0;
+  }
+  if (buf_scale_y < 1.0) {
+    buf_scale_y = 1.0;
+  }
 
-  // calculate area to read from file
+  // corners of the area we shall draw
   QPointF pt1 = buf.ref1;
   QPointF pt2 = buf.ref2;
   QPointF pt3 = buf.ref3;
@@ -263,42 +270,77 @@ void CDemVRT::draw(IDrawContext::buffer_t& buf) {
   pt3 = trInv.map(pt3);
   pt4 = trInv.map(pt4);
 
-  qint32 left, right, top, bottom;
-  left = qRound(pt1.x() < pt4.x() ? pt1.x() : pt4.x());
-  right = qRound(pt2.x() > pt3.x() ? pt2.x() : pt3.x());
-  top = qRound(pt1.y() < pt2.y() ? pt1.y() : pt2.y());
-  bottom = qRound(pt4.y() > pt3.y() ? pt4.y() : pt3.y());
+  // bounds of the area to draw in the coordinate space of the DEM
+  qreal left = pt1.x() < pt4.x() ? pt1.x() : pt4.x();
+  qreal right = pt2.x() > pt3.x() ? pt2.x() : pt3.x();
+  qreal top = pt1.y() < pt2.y() ? pt1.y() : pt2.y();
+  qreal bottom = pt4.y() > pt3.y() ? pt4.y() : pt3.y();
 
-  if (left <= 0) {
-    left = 1;
-  }
-  if (left >= xsize_px) {
-    left = xsize_px - 1;
+  if ((top > ysize_px) || (left > xsize_px) || (bottom < 0) || (right < 0)) {
+    // current view is entirely outside the bounds of the DEM so there is nothing to draw
+    return;
   }
 
-  if (top <= 0) {
-    top = 1;
+  // clip bounds
+  if (left < buf_scale_x) {
+    left = buf_scale_x;
   }
-  if (top >= ysize_px) {
-    top = ysize_px - 1;
+  if (top < buf_scale_y) {
+    top = buf_scale_y;
   }
-
-  if (right >= xsize_px) {
-    right = xsize_px - 1;
+  if (right + buf_scale_x > xsize_px) {
+    right = xsize_px - buf_scale_x;
   }
-  if (right <= 0) {
-    right = 1;
-  }
-
-  if (bottom >= ysize_px) {
-    bottom = ysize_px - 1;
-  }
-  if (bottom <= 0) {
-    bottom = 1;
+  if (bottom + buf_scale_y > ysize_px) {
+    bottom = ysize_px - buf_scale_y;
   }
 
-  const qint32 w = 2048 / xscale;
-  const qint32 h = 2048 / xscale;
+  // guard against degenerate/inverted bounds (e.g. extreme geotransforms or
+  // projections where the corner pairing above doesn't yield left<right/top<bottom)
+  if (right <= left || bottom <= top) {
+    return;
+  }
+
+  // the shading algorithms need a extra 1px border
+  // that border needs to be scaled as well
+  qreal x = left - buf_scale_x;
+  qreal y = top - buf_scale_y;
+  qreal w = right - left;
+  qreal h = bottom - top;
+
+  // dimensions we request from the DEM
+  qreal w_dem = w + 2.0 * buf_scale_x;
+  qreal h_dem = h + 2.0 * buf_scale_y;
+
+  // dimensions of the buffer GDAL will read into
+  quint32 w_buf = qRound(w_dem / buf_scale_x);
+  quint32 h_buf = qRound(h_dem / buf_scale_y);
+  if (w_buf < 3 || h_buf < 3) {
+    return;
+  }
+
+  QVector<float> data(static_cast<qsizetype>(w_buf) * h_buf);
+  {
+    QMutexLocker lock(&mutex);
+
+    // by requesting a different size than the size of the buffer GDAL will automatically do scaling for us and use
+    // overviews
+    CPLErr err = dataset->GetRasterBand(1)->ReadRaster(
+        data.data(), static_cast<size_t>(w_buf) * h_buf, x, y, w_dem, h_dem, w_buf, h_buf, GRIORA_Bilinear,
+        [](double dfc, const char* msg, void* arg) -> int {
+          auto dem = reinterpret_cast<CDemDraw*>(arg);
+          return !dem->needsRedraw();
+        },
+        dem);
+
+    if (err != CE_None) {
+      return;
+    }
+  }
+
+  // get pixel offset of top left buffer corner
+  QPointF pp = buf.ref1;
+  dem->convertRad2Px(pp);
 
   // start to draw the map
   QPainter p(&buf.image);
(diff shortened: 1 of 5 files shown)


kiozen:
I just gave it a quick smoke test on my fairly weak laptop and experienced a complete stall. 

My guess:

My DEM has no overviews. This applies probably to all QMS users, as overviews had no gain in the past. Without overviews  the function call into GDAL to read data takes quite long. As it is not interruptible the thread "stalls". Well, probably it will return in some time but from UX point of view it stalls. I had to kill the process and restart.

This is kind of bad as most users will not understand what to do, why it happens. And some might not even have enough disk space for overviews. 

I dug deep into my memories trying to recall implementing the DEM stuff. There is a weak memory that I tried reading scaled chunks many years ago and abandoned the idea for that reason. Right now it's slow for larger areas, but it's interruptible. By that the user can zoom closer, find the sweet spot and limit the visibility range.  For outer zoom levels DEM with less resolution kick in. They are fairly small compared to the 1x1m ones.

Additionally I made an automated review. It found 2 issues. But I need more time to verify and understand the findings. They might be false positives.

jmlzr:
I assume by "complete stall" you mean that it's not interruptible, not that it's slower? Because I can't see how this could be slower under any circumstances and tested with a 10m DEM that does not have overviews myself which felt at least as fast if not faster than before. Though I'm not saying that uninterruptible is not bad as well.

Overviews are pretty common, especially high resolution models may already come with them (they can be embedded in GeoTIFFs). But even if not the new code should be faster as the the rather slow shading algorithms have to handle way fewer pixels.

I saw that one can attach a progress callback to `ReadRaster` which is able to cancel the operation. So I will look into that!

> And some might not even have enough disk space for overviews.

That is a very unlikely scenario. Even a full set of overviews will only add around 30-40% of storage size to a DEM in practice.

> By that the user can zoom closer, find the sweet spot and limit the visibility range. For outer zoom levels DEM with less resolution kick in. They are fairly small compared to the 1x1m ones.

But this is just a workaround to archive what overviews are supposed to do with a lot of extra steps. Plus - besides the "stall" issue making range tuning annoying - you can still do that as the `outOfScale` logic is untouched.

jmlzr:
Okay, this was actually even easier than I thought. :grin: 

Now it aborts the GDAL read operation when a redraw is needed. Could you please try again and confirm?

wthaem:
I have some file-based DEM tests in various formats (TIFF, ASC, HGT, XYZ, ...). Some of these files are rather big. I didn't check if there are overviews in the files. My visual impression from testing this PR: DEM data is loaded quite fast. Loading online maps at the same time as the DEM takes more time than the DEMs do. Without this PR it is quite often the opposite.

src/qmapshack/dem/CDemVRT.cpp
```diff
@@ -309,77 +349,60 @@ void CDemVRT::draw(IDrawContext::buffer_t& buf) {
...
-    }
+  quint32 w_used = w_buf - 2;
+  quint32 h_used = h_buf - 2;
 
-    for (qint32 x = left - 1; x < right; x += w) {
+  // run the shadings in paralell on equal sized chunks
+  quint32 n_x = 4;
+  quint32 n_y = 4;
+  for (quint32 i = 0; i < n_y; ++i) {
+    for (quint32 j = 0; j < n_y; ++j) {
```
kiozen: j < n_y ? probably n_x

src/qmapshack/dem/IDem.cpp
```diff
@@ -213,22 +213,21 @@ int IDem::getFactorHillshading() const {
...
 #define AZ (315 * DEG_TO_RAD)
-  for (unsigned int m = 1; m <= h; m++) {
-    unsigned char* scan = img.scanLine(m - 1);
-    for (unsigned int n = 1; n <= w; n++) {
+  for (unsigned int m = 0; m < h; m++) {
+    unsigned char* scan = img.scanLine(m);
+    for (unsigned int n = 0; n < w; n++) {
       float win[eWinsize3x3];
-      fillWindow(data, n, m, wp2, win);
+      fillWindow(data, n + x, m + y, dx, win);
```
kiozen: dx will be shadowed in line 234

src/qmapshack/dem/CDemVRT.cpp
```diff
@@ -263,42 +270,75 @@ void CDemVRT::draw(IDrawContext::buffer_t& buf) {
...
   }
-  if (right <= 0) {
-    right = 1;
+  if (bottom + buf_scale_y > ysize_px) {
+    bottom = ysize_px - buf_scale_y;
   }
 
-  if (bottom >= ysize_px) {
-    bottom = ysize_px - 1;
+  // the shading algorithms need a extra 1px border
```
kiozen, line 355: I suggest a guard before that

```
  // guard against degenerate/inverted bounds (e.g. extreme geotransforms or
  // projections where the corner pairing above doesn't yield left<right/top<bottom)
  if (right <= left || bottom <= top) {
    return;
  }
```

src/qmapshack/dem/CDemVRT.cpp
```diff
@@ -263,42 +270,75 @@ void CDemVRT::draw(IDrawContext::buffer_t& buf) {
...
+  {
+    QMutexLocker lock(&mutex);
+
+    // by requesting a different size than the size of the buffer GDAL will automatically do scaling for us and use overviews
+    CPLErr err = dataset->GetRasterBand(1)->ReadRaster(data.data(),
+                                                       static_cast<size_t>(w_buf) * h_buf,
+                                                       x, y,
+                                                       w_dem, h_dem,
+                                                       w_buf, h_buf,
+                                                       GRIORA_Bilinear,
```
kiozen: This will destroy noData values compared to near neighbor resampling. By that DEMs with holes will produce funny results a s the noData guards will not work.  I am a bit undecided. Bi-linear will produce better results on good data for sure.

kiozen (changes requested):

kiozen:
> Okay, this was actually even easier than I thought. 😁
> 
> Now it aborts the GDAL read operation when a redraw is needed. Could you please try again and confirm?

Yes, now it behaves nicely.

mitxel-m:
This sounds good, and it will reduce problems for newer users who aren't aware that they can stop rendering by zoom level.

Reading the comments has brought a question to mind. Apologies if the answer is obvious or if I'm missing something but on theese days I don't have a computer with all the necessary stuff to test the PR so I can't see for myself.

About this:

>This will destroy noData values compared to near neighbor resampling. By that DEMs with holes will produce funny results a s the noData guards will not work. I am a bit undecided. Bi-linear will produce better results on good data for sure.

Does this mean that pixels with `nodata` will now be filled with interpolated values, and they will return a false but apparently valid value?

If true, I think that's a big mistake. `nodata` is `nodata`, and it should remain exactly as it is.

kiozen:
I don't think it is making much difference:

Current dev:

grafik


This patch:

grafik

kiozen:
Another observation while testing:  Find the peak of a small hill, place the mouse on top of it and zoom in and out. It looks like the hill shading is jumping around. Compare to dev.

src/qmapshack/dem/CDemVRT.cpp
jmlzr: Why do you think so? Since GDAL knows about nodata I'm pretty sure it handles those points correctly when scaling and I do have DEMs with missing data (cut at country borders as well as missing tiles) and do not see anything remotely "funny".

src/qmapshack/dem/IDem.cpp
jmlzr: I just took the `dx` naming from the signature of `fillWindow` since it only ends up there. Are you fine with me renaming it to `stride` which would be a much clearer naming for this meaning anyways?

kiozen: much cleaner naming is always good :smile:

src/qmapshack/dem/CDemVRT.cpp
kiozen: seem my two screenshots below. It's fine.

jmlzr:
> Another observation while testing: Find the peak of a small hill, place the mouse on top of it and zoom in and out. It looks like the hill shading is jumping around. Compare to dev.

I see what you mean but for me dev does similar. I think it's mostly just more noticeable with this PR since it loads much faster after the zoom. Might be worse but hard to quantify. And I have no idea what could cause that. Maybe some reprojection issue that's now worse as under some conditions (xscale > 1, large viewport) the tiles this PR draws are larger than dev?
I'm still looking around for a spot where my map and DEM both show a very clear and matching feature to check if there is actually an alignment issue at certain zoom levels but in general use I did not see any mismatch.

Does your DEM need a lot of reprojection? Mine do, so to check if it's a projection issue I think I'll try to warp one into map projection and see if it's better then.

Btw, I do see much more noticeable borders between tiles/chunks with this PR. Is it the same for you?

kiozen:
Usually this hints some rounding issue. (e.g. truncation vs rounding) 


> Btw, I do see much more noticeable borders between tiles/chunks with this PR. Is it the same for you?

yes, that's also a hint of loosing accuracy because of rounding and quantification.  But even in the original code this issue is not 100% solved. One main reason is that the code does not a proper re-projection. It's just doing an approximation (scale and rotate). On small areas this is negligible. The larger the area the more error compared to a real re-projection.

jmlzr:
Implemented all the change requests.

> Usually this hints some rounding issue. (e.g. truncation vs rounding)

I do a lot less rounding than the original code but I'll debug a bit more to see if I spot anything.

> yes, that's also a hint of loosing accuracy because of rounding and quantification. But even in the original code this issue is not 100% solved. One main reason is that the code does not a proper re-projection. It's just doing an approximation (scale and rotate). On small areas this is negligible. The larger the area the more error compared to a real re-projection.

Yeah, one of the things I wanted to look at next was see if I can come up with some clever way to round the quads in a way that makes them not (or less often) overlap or have gaps but that's out of scope of this PR.

In the process of writing this PR I also made a version of CDemVRT that configures GDAL so it automatically does a proper reprojection/resampling on loading, but that comes with a noticeable performance impact so I removed it again. We could maybe add it in the future as a optional feature that can be toggled in the GUI but I didn't want to have it always on considering it's impact.

Edit: Do you think it would be worthwhile to force the use of drawTileHQ here to get smaller quads?

kiozen:
Yes performance was always an issue. The code was develop for a CPU with 4 threads max back in those days :smile: 

The blocking GDAL API was always a major drawback. But as I learned some APIs have a callback to interrupt. Can't tell if that slipped by attention when doing this code or it was introduced later. If the re-projection API has the same feature now it would get into scope again.

> Edit: Do you think it would be worthwhile to force the use of drawTileHQ here to get smaller quads?

You can give it a try.

kiozen:
Another small observation: "ReadRaster error" is printed quite frequently. I guess this is because the read process has been aborted while moving/zooming the map. It would be better to just print a debug out put if there is a real error.

jmlzr:
There indeed was a bug: The position and dimension of each chunk in the buffer are calculated as integers while the coordinates in DEM space were done separately as qreals. So if `w_used` was not integer divisible by `n_x` it would lead to different results, plus it would miss the last few pixels.
Fixed that and now I do not notice any more jumping than with dev and the gaps also improved to be the same.

> Another small observation: "ReadRaster error" is printed quite frequently. I guess this is because the read process has been aborted while moving/zooming the map. It would be better to just print a debug out put if there is a real error.

Yes, I removed that debug line again.

kiozen:
I still have a lot of wobble:

https://github.com/user-attachments/assets/b5f0cb04-b3e4-459c-b760-88c7acb53f23

kiozen:
compared to current dev:


https://github.com/user-attachments/assets/b34c91ea-62a7-4714-9f54-0cb02ebe0a29

jmlzr:
This is a preexisting bug/limitation and just a result of chunk/tile size. Applying the following patch to dev to make it use chunks of the same size as this PR gives the same wobble behavior:

```
diff --git a/src/qmapshack/dem/CDemVRT.cpp b/src/qmapshack/dem/CDemVRT.cpp
index 6c5a0186..ef54e084 100644
--- a/src/qmapshack/dem/CDemVRT.cpp
+++ b/src/qmapshack/dem/CDemVRT.cpp
@@ -297,8 +297,8 @@ void CDemVRT::draw(IDrawContext::buffer_t& buf) {
     bottom = 1;
   }
 
-  const qint32 w = 2048 / xscale;
-  const qint32 h = 2048 / xscale;
+  const qint32 w = (right - left) / 4;
+  const qint32 h = (bottom - top) / 4;
 
   // start to draw the map
   QPainter p(&buf.image);
```

I recorded this with dev and this patch applied:

https://github.com/user-attachments/assets/26bfda35-cb18-4318-acbf-108aaefa4cc0

My guess is that it's discontinuities between the transforms of the different chunks caused by the pseudo-reprojection the existing code does. In my opinion the correct thing to do would be to draw the whole canvas in one go with a single transformation. That would probably only have imperceptibly more error than the current solution but would be continuous and get rid of the ugly lines as well.
In the meantime I "fixed" it for this PR by dramatically decreasing the size of the chunks to roughly match your example.

kiozen:
Meehhh. That's not nice. This approximation is really just good for small areas.

I agree that reading the full area re-projected would be the best. Much simpler code. But it must be interruptible at any stage to avoid these stalls we had in the beginning.

jmlzr:
Yeah, it is meh. With the approximation it's always gonna be a tradeoff, either discontinuities or accumulating errors and jumps on zooming.
Played around with having the whole view as a single quad with a DEM that I warped into the map projection beforehand and that works flawlessly though.

> I agree that reading the full area re-projected would be the best. Much simpler code. But it must be interruptible at any stage to avoid these stalls we had in the beginning.

Sounds good! I can work on that. Shall I open a new PR (and issue?) for that or shall I just do it in another commit on this one?

kiozen:
> Played around with having the whole view as a single quad with a DEM that I warped into the map projection beforehand and that works flawlessly though.

Yes but this is seldom the case. One key requirement right from the beginning of QMapShack was, to do reprojection on the fly. 

> Sounds good! I can work on that. Shall I open a new PR (and issue?) for that or shall I just do it in another commit on this one?

As you prefer. Maybe it's a good idea to keep this one until the next fix turns out successful. You never know :sweat_smile:

jmlzr:
I think we're mostly there now! Hope I didn't miss something again.

There might be a edge case where interruption isn't as fast as would be ideal but I did not find a way to reproduce it.

What do you think?

kiozen:
A quick smoke test is looking good. Need to test in depth this evening with a slower PC and a larger DEM database.

src/qmapshack/dem/CDemVRT.cpp
```diff
@@ -91,12 +93,58 @@ CDemVRT::CDemVRT(const QString& filename, CDemDraw* parent) : IDem(parent), file
   noData = pBand->GetNoDataValue(&hasNoData);
   qDebug() << "no data:" << hasNoData << noData;
 
+  // ------- setup warped VRT ---------------
+  OGRSpatialReference targetSRS;
+  targetSRS.SetFromUserInput(dem->getProjection().toUtf8());
```
kiozen: check return value and abort on error.

kiozen (changes requested):

kiozen:
Performance on a weak PC with 1x1m data and full set of overview is reasonable fast.  Just fix the error guard and we are ready for merge.

kiozen:
Ping! Just waiting for the last review comment to be fixed. Else ready to merge.

jmlzr:
Nice!

Sorry, just didn't have time the last days. Added the requested error check.

kiozen:
No worries. Just wanted to make sure my feedback was noticed.

Thanks very much for this patch. This is a major improvement.  

I will spend the next days on refactoring and cleaning up the DEM code. While doing the review I just noticed this stuff needs some love. The code hasn't been touched for years.

I also will try to apply the re-projection via GDAL to the VRT map handling. I do not expect performance gains, but maps also suffer artifacts due to the approximation. They are just not that visible. And if GDAL does the job properly it's always the preferred way.

jmlzr:
Sure, my pleasure!

> I will spend the next days on refactoring and cleaning up the DEM code. While doing the review I just noticed this stuff needs some love. The code hasn't been touched for years.

When touching `IDem` I really wished for C++26's `std::submdspan`. Will probably be a while though till that has enough wide spread library support.
I only looked at the rendering part plus anything I had to interact with but that code wasn't too bad. Decent readability wise for sure... otherwise I wouldn't have attempted this patch. haha

kiozen:
Yeah, it's not that bad. But it's not really good either. The code has been developed a long time ago and over the years features were added. So a bit of cleaning a sweeping does not hurt.

JFritzle:
I have noticed a worse performance of DEM rendering.

Perhaps an unusual scenario:
- QMS is running on Linux VM guest 
- DEM VRT combines 5 Sonny's 20m DTM GeoTIFFs (A, CH, D, F, I) of Central Europe
- GeoTIFFs reside on filesystem mounted from Windows VM host

->Hillshading of "Kleinwalsertal" previously took about 4 seconds
->Performance improved hillshading now takes about 18 seconds

grafik

kiozen:
Do the files have overview layers?

jmlzr:
Does your DEM use a different projection then you have configured for the map in QMS? Does it have overviews? (both can be checked with `gdal raster info dem.vrt`)

Needing a warp and not having overviews on a slow device/vm is the only scenario this can be slower. That is expected though, it's the price of having a proper reprojection.

To create overviews you can do something like:
```
gdaladdo -r cubic --config COMPRESS_OVERVIEW=ZSTD --config ZSTD_LEVEL_OVERVIEW=9 --config PREDICTOR_OVERVIEW=3 --config NUM_THREADS=ALL_CPUS --config GDAL_NUM_THREADS=ALL_CPUS dem.vrt
```

I would bet that improves the situation already a lot.

If you want maximum performance on slow hardware you should resample the DEM beforehand with `gdalwarp` into the map projection you use in QMS.

kiozen:
@JFritzle if you have overviews: Are they calculated for the vrt or do the DEM files themselves define overviews? 

Why I ask:  I am just in discussion about overview levels in the new CMapVrt enhancement.  There is a point in having a vrt that covers different files and those files have different to none overview levels.  Here is a rough summary.I hope it makes sense without following the complete discussion.


Mixed overviews across a VRT mosaic's sub-files:

When a mosaic VRT's underlying source files don't all have the same overview structure, here's what actually happens (verified against the system's GDAL via gdalbuildvrt):

- Different depths (e.g. tile A has overview factors [2,4], tile B has [2,4,8,16]): the VRT only advertises the intersection — [2,4]. Tile B's deeper levels exist on disk but are invisible to anything reading the VRT; they're silently wasted.
- One file with zero overviews (the rest fully built): the intersection with an empty set is empty — the VRT advertises no overviews at all. A single un-gdaladdo'd tile in an otherwise well-prepared mosaic disables overview-based acceleration for the entire mosaic, not just that tile.

This is GDAL's own behavior (confirmed in the generated <OverviewList> XML), not something CMapVRT/CDemVRT compute — both just read whatever the master VRT band reports (pBand->GetOverviewCount()) and gate their own warped-VRT virtual-overview registration on that single number. So they inherit this all-or-nothing-per-factor cliff directly: a large, mostly-well-prepared mosaic loses all warp-read acceleration because of one incompletely-prepared tile, with nothing in the code today to detect or work around that — it just silently reads slower.

JFritzle:
- Currently, I am using original GeoTIFFs downloaded from [Sonny' DTMs](https://sonny.4lima.de/).
- Using projection _EPSG:3857_ in QMS, I created a VRT file for each GeoTIFF file by
`gdalwarp -r cubic -t_srs epsg:3857 -of vrt "$tif" "$vrt"`.
- Then I concatenated these VRT file names into a list file by `echo *.vrt > vrt.list` ...
- ... and created a combined VRT by
 `gdalbuildvrt -r cubic -resolution highest -input_file_list vrt.list central_europe.vrt`
- DEM file _central_europe.vrt_ then is used by QMS.

Being no projection expert this is possibly not a good approach.

kiozen:
How many files does the vrt combine? <100 is ok. If it's more I would suggest to use gdal to group files into larger files. 

If your target SRS is EPSG:3857 I would also warp those files into that SRS using as much quality as possible. No need to do that over and over again on the fly.

And of course add overviews. Either the same set to each individual file (they should have the same resolution, too). Or on the vrt. Later requires to recalculate overviews if you add remove files from the vrt.

I use vrts per country, combining several files for a country. The overviews are calculated  on the vrt. With an 8 years old laptop (8 threads) I can use Bavarian 1x1m, Austrian 5x5m, Italian 10x10m and Swiss 20x20m at the same time on all zoom levels. Rendering takes max 10s.

JFritzle:
Combined VRT references 5 country VRT files:
- DTM Austria 10m v2 by Sonny
- DTM France 20m v3 by Sonny
- DTM Germany 20m v3b by Sonny
- DTM Italy 20m v2b by Sonny
- DTM Switzerland 10m v2 by Sonny

kiozen:
and what does gdalinfo tell about overview levels for the vrt?

JFritzle:
gdalinfo does not report overviews neither for original country GeoTIFFs, nor for country VRTs, nor for combined VRT.

kiozen:
And that is probably the main reason for the performance hit. And the point where we hit this discussion: https://github.com/Maproom/qmapshack/pull/1126#issuecomment-4777515285

JFritzle:
After translating original country GeoTIFFs into _Cloud Optimized GeoTIFFs_ by
gdal_translate "$tif" "$cogtif" -of COG -co BIGTIFF=YES,
each translated GeoTIFF has these overviews:
`Overviews: 17672x12043, 8836x6021, 4418x3010, 2209x1505, 1104x752, 552x376, 276x188`
Translated GeoTIFFs are about twice as large as their originals.

Performance improved hillshading now takes about 3 seconds, even on Linux VM guest,
and about 1.5 seconds on Windows host.

kiozen:
Yes it's a trade off between hard disk size and speed.  But good to see that it is up to speed now.

JFritzle:
Courageous I tried to apply hillshading using Cloud Optimized GeoTIFFs to an area reaching from Nuremberg at NW to Maribor at SE. Green bar of DEM was flashing and wheel was spinning. After minutes, I tried to interrupt process. Not knowing how, I hit DEM button to inactivate. QMS became unresponsible for quite a long time.

This leads to the question: Is it possible at all to interrupt hillshading? If yes, how to achieve?

kiozen:
Ouch that is bad. It should be interruptible. If it is not it's likely a hanging GDAL API not calling the progress handler to interrupt it. 

If you want you can try several combinations:

* Each DEM in it's own vrt. 
  * overviews for vrt
  * overviews for DEM
* A combined vrt
  *  overviews for vrt
  * overviews for DEM

There is a command line in gdaladdo to enforce overviews in parallel to your DEM file. You do not want to write into your orginals.

jmlzr:
Making them COG most likely has zero advantage here (unless your disk is incredibly slow maybe), the actual advantage comes from the overviews that were generated as a side effect.
The massive size increase is probably because the overviews aren't compressed. With zstd compression on level 9 overviews only add around ~1/3 of the original size for me.

JFritzle:
@jmlzr 
>With zstd compression on level 9 overviews only add around ~1/3 of the original size for me.

Verified. Without performance impacts.

Issue #963, opened 2026-02-14
Action to move a map to the top in map list.


mitxel-m:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

One of the most powerful features of QMS is the ability to overlay maps. It is desirable to have the maps you are using at the top of the map list so that you can order them according to the redrawing order you want. Previously, when you activated a map, it automatically moved to the top of the list, but since PR935, with the redesign of the map panel, this has changed and when you activate a map, it does not jump to the top of the list. 

This new behaviour is fine for me because just as you want to keep the map settings, you also want to keep its position in the list. However, this has brought about an inconvenience: now, after activating the map, the user has to manually drag it to the desired position, and when the list of maps is very long, this is frustrating. In addition, to search for a map, we can use a very powerful option in the panel, the filter, but if I use the filter, I will not be able to move the map to the desired position.

There is also another drawback: if a user forgets an activated map at the bottom of the list, it will overlap all the others, and you will have to scroll through the entire list to find it.

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

Add an action to move the map to the top of the list whenever the user wants:

1- In the context menu (right button) add an action: ‘Move to top’.

2- Modify the behaviour of the activate/deactivate map button with CTRL+click:  

   - For inactive or not used maps: CTRL+click activates the map and moves it automatically to the top
   - For active maps: CTRL+click keeps the map active and moves it automatically to the top

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

An alternative would be to add a button in the map filter line to show only active maps, but this would not solve all cases.

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

The CTRL+click behaviour on the ‘activate/deactivate map’ button would greatly improve the user experience, and since we have a new button, it would be an opportunity to take full advantage of it without adding other buttons.

<img width="112" height="53" alt="Image" src="https://github.com/user-attachments/assets/c3ba1a56-36f5-4687-a957-63f082fbce53" />

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

wthaem:
[...]

PR #1121, opened 2026-06-19
[QMS-963] Add action to move a map to the top in map list

JFritzle:

QMS-#963

### What you have done:

For current _selected_ map or DEM list item, following keyboard actions have been added:
- If _Shift+Home_ is simultaneously pressed, the _selected_ list item is moved to the top of the list.
- If _Shift+End_ is simultaneously pressed, the _selected_ list item is moved to the bottom of the list.
- If _Shift+Up_ is simultaneously pressed, the _selected_ list item is moved one row up.
- If _Shift+Down_ is simultaneously pressed, the _selected_ list item is moved one row down.

In addition,  _selected_ item's context menu has been expanded
- with shortcuts showing
- and tootips enabled:
grafik

1. Apply one of above actions to _selected_ list item
2. See that item has moved as mentioned above
3. Open item's context menu
4. See expanded menu

- [x] yes

- [x] yes

### DIFF
--- a/src/qmapshack/dem/IDemList.ui
+++ b/src/qmapshack/dem/IDemList.ui
@@ -171,7 +171,7 @@
     <string>Shift+Home</string>
    </property>
    <property name="menuRole">
-    <enum>QAction::MenuRole::TextHeuristicRole</enum>
+    <enum>QAction::MenuRole::NoRole</enum>
    </property>
   </action>
   <action name="actionMoveUp">
@@ -188,6 +188,9 @@
    <property name="shortcut">
     <string>Shift+Up</string>
    </property>
+   <property name="menuRole">
+    <enum>QAction::MenuRole::NoRole</enum>
+   </property>
   </action>
   <action name="actionMoveDown">
    <property name="icon">
@@ -202,6 +205,9 @@
    </property>
    <property name="shortcut">
     <string>Shift+Down</string>
+   </property>
+   <property name="menuRole">
+    <enum>QAction::MenuRole::NoRole</enum>
    </property>
   </action>
   <action name="actionMoveEnd">
@@ -219,7 +225,7 @@
     <string>Shift+End</string>
    </property>
    <property name="menuRole">
-    <enum>QAction::MenuRole::TextHeuristicRole</enum>
+    <enum>QAction::MenuRole::NoRole</enum>
    </property>
   </action>
   <action name="actionReloadDem">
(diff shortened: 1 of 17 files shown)


JFritzle:
Not yet ready for merge.

JFritzle:
Ready for merge 😄.

kiozen (dismissed):
I simplified the code a bit and removed some old stupid stuff. See last two commits.

Only one peculiar thing remaining:

To move an item to top or bottom of list I can use shift and page up/ page down or shift and  pos1/end in parallel. This seems a bit "doppelgemoppelt". 

And the key binding from the actions in the menu is not really used. Nothing serious. But let's hope no one is changing something in the key handler forgetting to fix the action and vice versa.

JFritzle:
👍Nice simplification. I also tried to simplify the code but I lacked the brilliant idea.

I do not agree that _shift+page up/page down_ and _shift+pos1/end_ is redundant ("doppelgemoppelt"). Both behave totally different:
- _Shift+pos1/end_ moves **position** of selected item in list to top/bottom of list while keeping item's **selection** state.
This function has not yet been available. Changing item's position in list may change item's visibility priority if item is _active_.
- _Shift+page up/page down_ moves current **selection** to item one page above/page below while keeping previous item's **postion** in list.
This function has already been available. _Shift+page up/page down_ behaves exactly as _page up/page down_ and this treewidget's built-in behavior has not been touched. Changing which item is selected does never change any visibility priority.
Further more, if list is longer than 1 visible page, position of _one page above/page below_ is different from position _top/bottom_.

Do you agree?

Yes, the key binding from the actions in the menu is not really used while context menu is open. But it is a reminder which keys to use without need to open context menu. If desired and needed, keys could be made working too while context menu is open.

JFritzle:
Done.

src/qmapshack/dem/IDemList.ui
```diff
@@ -167,18 +185,42 @@
...
+   <property name="toolTip">
+    <string>Show DEM on top of others</string>
+   </property>
+   <property name="shortcut">
+    <string>Shift+End</string>
+   </property>
+   <property name="menuRole">
+    <enum>QAction::MenuRole::TextHeuristicRole</enum>
+   </property>
   </action>
```
kiozen, line 230: I think this has to be set to NoRole like on all other actions.  This was done ages ago to circumvent some issues on MacOS (using the menu text to create key bindings)

kiozen (changes requested):

kiozen:
Ready for merge?

src/qmapshack/dem/IDemList.ui
JFritzle: _IDemList.ui_ & _IMapList.ui_: Replaced each _TextHeuristicRole_ by _NoRole_.
_CDemList.cpp_ & _CMapList.cpp_: Ultimate simplification.

Issue #1125, opened 2026-06-22
Use GDAL's warp and scale API to draw *.vrt maps


kiozen:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

The current code reads small tiles and transforms them (rotate and scale) into the target projection. This is only an approximation. And a lot of code in QMapShack to maintain.

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

GDAL has an API that does reprojection and scaling including the use of overlays. This has been used for DEM data already.  The goal is to use the same for VRT maps.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

Leave the code as is.

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

I do it.

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

Issue #903, opened 2025-12-06
Code housekeeping


kiozen:
This is a developer ticket.  Changes in the code that do housekeeping should use that ticket. Add a short line to that ticket to keep track of what was done. 

Commits on that ticket do not need to be part of the changelog.

PR #1126, opened 2026-06-22
[QMS-1125] Use GDAL's warp and scale API to draw *.vrt maps

kiozen:

QMS-#1125

### What you have done:

Used the same pattern as in CDemVrt to read re-projected and scaled map data via the GDAL API

Test the PR on all kinds of raster maps, enclosed in a GDAL VRT file.

- [x] yes

- [x] yes

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -66,6 +66,8 @@ if (condition)
 
 **Prefer `QFileInfo::exists(path)` (static) over `QFileInfo(path).exists()`** when you only need an existence check — no need to construct a full `QFileInfo`.
 
+**Prefer Qt's fixed-width typedefs over plain C++ types**: `qint32`/`quint32` over `int`/`unsigned int`, `qreal` over `double`, `qsizetype` over `size_t`/`ptrdiff_t` for sizes coming from Qt containers. Exception: when a parameter type is dictated by an external API (e.g. GDAL's `int*` out-param, or `size_t` in a `ReadRaster()` signature), match that API's type instead of casting.
+
 ---
 
 ## Building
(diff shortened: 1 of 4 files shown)


kiozen:
@mitxel-m I guess you have quite a collection of vrt maps. Can you give it a test.

@jmlzr this has the enhanced resolving of overview levels (CMapVRT::collectOverviewFactors). Maybe this is something for the CDemVrt, too. 

Probably DEM and maps are sharing some functionality here. This is a topic for #903 Code housekeeping.

jmlzr:
What's enhanced about the resolving of overview levels? Crawling through the tiles is not necessary, GDAL will detect internal overviews by itself. For example I have a VRT that does not have overviews but the geotiffs it consists of each have overviews and my CDemVRT code sees and uses these overviews without problems.

kiozen:
> What's enhanced about the resolving of overview levels? Crawling through the tiles is not necessary, GDAL will detect internal overviews by itself. For example I have a VRT that does not have overviews but the geotiffs it consists of each have overviews and my CDemVRT code sees and uses these overviews without problems.

Yes that is the good case. But if you have a collection of DEMs with different resolution and different number of overviews things get tricky. Especially if one DEM has no overviews at all.  The GDAL API can handle quite a lot of this stuff properly, as long as overviews are registered properly.

jmlzr:
GDAL can really handle such messy combinations? Did you test that?
When I was writing the code for DEM I tried something quite similar to coax GDAL into using external overviews of a VRT in addition to the internal ones of the tiles (to potentially further speed up large downscale factors where multiple tiles are visible by saving GDAL from having to get data from multiple files) but GDAL started to segfault with that. And that was a much simpler scenario then a mixture of resolutions and incomplete overviews.

kiozen:
Yes, did some test to check what overview levels are reported for mixed files and different combinations. Haven't tested how well the API handles all kind of evil. 

But as we learn from that post: https://github.com/Maproom/qmapshack/pull/1116#issuecomment-4777574126 mixed files are not that far from imagination.  

But it's surely debatable how far the code should support every craziness imaginable.

jmlzr:
I was just about to comment there... I don't see how this is related as this wouldn't have helped him at all since he didn't have overviews at any level. In that case the best way is to create them as external overviews for the VRT at the highest level anyways.

Yeah, imho it is really not worth the added code (and therefore maintenance burden) to support such adventurous constructions plus I'd fear to invite bugs on lots of weird edge cases.

kiozen:
> Yeah, imho it is really not worth the added code (and therefore maintenance burden) to support such adventurous constructions plus I'd fear to invite bugs on lots of weird edge cases.

Valid reasoning. On the other hand of side I do not know how users handle their DEM data. No usage statistics in QMapShack. But I think the scenario of @JFritzle is not that far off track. 

> In that case the best way is to create them as external overviews for the VRT at the highest level anyways.

The easiest way. Not the best. If you add another DEM you have to build the overviews again. That is why  I use one vrt per DEM.  Especially as in the past each DEM needed fine tuning for the visible zoom levels. 

However now all DEMs can be shown at each zoom level. Therefore calculating overviews per DEM and combining them into a vrt is really an option.

jmlzr:
Ah, you misunderstood me there. I think too that having separate VRTs for different DEMs is certainly the proper way to handle it. Especially with different resolutions or if they aren't clean tiles. (How would a VRT with overlapping sub-datasets behave?)
What I meant was just that if the subfiles don't have overviews already, adding them at the highest level is whats probably best for the average user. The only advantage of adding them to the subfiles would be that if you change part of them you don't have to recompute the overviews for everything. Which would be annoying for an agency that's keeping a DEM for a country up to date but I doubt the average QMS user patches their DEMs haha

kiozen:
One idea that's in my head is to measure the time maps or DEM take to be rendered. If it's too long a message box pops up with instruction how to enhance performance. Therefore finding best practice is kind of important. 

But that's worth another ticket.

kiozen:
Btw I applied the fix to CDemVrt, and created a Frankenstein VRT:

```
File              | CRS         | Size (px)      | Pixel size | Overview levels         | Overviews (external)
------------------|-------------|----------------|------------|--------------------------|---------------------
austria.tif       | EPSG:32633  | 58127 x 29852  | 10 m       | 2, 4, 8, 16, 32, 64      | austria.tif.ovr
slovenia.tif      | EPSG:32633  | 12404 x 8141   | 20 m       | 3, 9, 27, 81, 243        | slovenia.tif.ovr
swiss_utm33.tif   | EPSG:32633  | 18601 x 13402  | 20 m       | 5, 10, 25, 50, 100, 200  | swiss_utm33.tif.ovr
mosaic.vrt        | EPSG:32633  | 52102 x 24219  | 16.667 m   | (none — fallback target) | n/a

Expected vs. actual collectOverviewFactors() result on mosaic.vrt:

Predicted: 1, 2, 4, 5, 6, 10, 11, 12, 19, 30, 32, 38, 60, 97, 119, 238, 286
Actual:    1, 2, 4, 5, 6, 10, 11, 12, 19, 30, 32, 38, 60, 97, 119, 237, 286
```
 Works like a charm. I will create a ticket and a PR.

jmlzr:
How does GDAL handle the missing overview levels? Like when you're reading data from say Slovenia at a scale that would make a overview level 5 the ideal target to read, what would happen?
Will it go to level 3 and downscale it more (will it downscale twice if the read operation requires a downscale on top of the overview?) or go to the full resolution and downscale that?

kiozen:
Hard to tell what GDAL does exactly internally without analyzing the code. But performance is bad for sure.  The Frankenstein VRT takes ~10s to render the complete area on the appropriate zoom level.  A proper setup should render in ~1s.

But that is fine. This is not meant to fix completely broken VRTs.  But it handles VRTs with mixed sources and halfway sensible overviews per file nicely.  Together with the next step to show an advisory dialog if rendering takes too much time this should help users to get their VRTs done "right".

jmlzr:
Ah, yeah if it's just slow but works it should be fine together with an advisory. I was just concerned of bugs and users building such stupid contraptions thinking it's a good idea and then wonder why its slow.

And I guess kudos to the GDAL dev's. Their API and documentation suck but I was already impressed what that lib can do and now I'm even a lot more if it can handle something as crazy as that. haha

kiozen:
Yes GDAL and Proj are real jewels. QMapShack wouldn't be that powerful without them.

Btw the advisory dialog is already pushed. You can give it a try. I just make some minor improvements right now.

Issue #902, opened 2025-12-06
Maps, text and icons are "blurred" in Windows 10 version (1.19.0) - Resolution problem?


Dret42:
### Maps, text and icons are "blurred" in Windows 10

Maps, text and icons are "blurred" in Windows 10 version (1.19.0) - Resolution problem?

**The problem is more evident on .IMG vecotr map format.**

### What have you done to circle down the problem?
I tryed to find some setting about QMapShack resolution but with no luck.

### To Reproduce
Install on Window 10 and compare to 1.17.0 version

### Expected behavior
Map should be not "blurred"

### Screenshots

Here two screenshots (form both version) to compare.

Map used is in Garmin .IMG format downloaded form https://www.wanderreitkarte.de/index.php?lang=en
(but problem is visible in all .IMG format maps)

**1.17**

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/bfd4f3a1-5e59-40e8-ae9c-3813298ecbcd" />

**1.19 (blurred)**

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/2f799ab3-bf7b-4e8b-9743-86bcbf1660ec" />

### Desktop
[comment]: # (please complete the following information)

 - OS: Windows 10
 - QMapShack Version:  1.19.0
 - Window Manager:  -

Dret42:
> Open _View -> Setup Map View_ . In _Projection & Datum_ window hit _Grid symbol_ button at upper right corner and select _World Mercator (OSM)_ for projection. Result will be _EPSG:3857_. That solved at least blur issue in OSM based maps after moving to QMS 1.18.x+.

**Unfortunately it is already in EPGS:3857** :(

Even changing to UTM the situation does not change!

Any other Idea?

Dret42:
Maybe I found a possible issue...

One years ago I got a similar problem on some maps into QGIS.
I that case the situation was generated by the default resolution of Windows on my Lenovo laptop that is 125% and not 100%.
This kind of settings is very common in many laptop Windows evironment..

In that case the QGIS developer added a function to adjust map resolution based also on this king of setting.

Unfortunatly is not possible to turn resolution at 100% because all other app (and SO too!) would broke!!!

I hope this could be an idea for a fix!

Thanks in advance!

JFritzle:
@kiozen 
_The only big change is switching from Qt5 to Qt6. That could have some fallout._

[...]

PR #1127, opened 2026-06-23
[QMS-902] Add basic HiDPI support

jmlzr:

QMS-#902

### What you have done:

Add handling for device pixel ratios to canvas and draw context. This makes IMG maps behave correctly on scaled displays. Some minor tweaks to raster based maps that make them scale properly at least under some conditions and plattforms.

Try different types of maps with different scaling ratios configured in the OS. Move window between displays with different scaling ratios.

See also discussion in #902 where changes where tested already. This is just my "hidpi" branch rolled into a PR.

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,5 @@
 V1.XX.X
+[QMS-902] Add basic HiDPI support
 [QMS-963] Add action to move a map to the top in map list
 [QMS-1105] Fix: Context menu shortcuts not shown (macOS)
 [QMS-1111] Add percentage values to track range
--- a/src/qmapshack/canvas/CCanvas.cpp
+++ b/src/qmapshack/canvas/CCanvas.cpp
@@ -1183,6 +1183,14 @@ bool CCanvas::setDrawContextSize(const QSize& s) {
   return done;
 }
 
+bool CCanvas::setDrawContextPixelRatio(qreal ratio) {
+  bool done = true;
+  for (IDrawContext* context : std::as_const(allDrawContext)) {
+    done &= context->setPixelRatio(ratio);
+  }
+  return done;
+}
+
 void CCanvas::print(QPainter& p, const QRectF& area, const QPointF& focus, bool printScale) {
   const QSize oldSize = size();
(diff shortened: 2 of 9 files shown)


kkarsten62:
I just checked this commit in my system environment:
image

Screen is on 200% scale:
image

For vector maps, ex. "Freizeitkarte" it looks now more brillant and sharp, **text can be read ==> ok**:
image

but I see a regression for raster maps, ex. "OpenCycleMap", **text to small to read ==> not ok**: 
image

Can somebody else confirm this? Or do we have some settings to increase text font size?

kkarsten62:
And a next observation and for me a regression:
Zooming and paning with OpencycleMap (TMS) is slower. Each action (zoom, pan) needs 2-5 seconds and and many tiles must be reloaded, cache size is on maximum (1000 MB). It makes a fluid working (ex. creating a new track) for me impossible.
Bildschirmfoto vom 2026-06-23 20-00-49
And I have already a fast computer with a 1GBit/s internet download rate.

jmlzr:
I think the corresponding issue (#902) is a better place for this discussion than here.

Though your first complaint is not a bug but working exactly as expected. That is just a limitation of raster maps as discussed in #902 already. You need HiDPI tiles if you want the text to be an appropriate size.

Thunderforest  provides hidpi tiles by just adding a `@2x` to the URL, see: https://www.thunderforest.com/docs/map-tiles-api/

For example this would be the default OpenCycleMap.tms with that change applied:

```
<TMS>
 <Title>Opencyclemap</Title>
 <MinZoomLevel>1</MinZoomLevel>
 <MaxZoomLevel>1024</MaxZoomLevel>
 <Layer idx="0">
  <ServerUrl>http://a.tile.thunderforest.com/cycle/%1/%2/%3@2x.png</ServerUrl>
 </Layer>
 <Copyright>Map data: (c) OpenStreetMap contributors, ODbL | Rendering: (c) OpenCycleMap , CC-BY-SA</Copyright>
</TMS>
```

Issue #1128, opened 2026-06-24
Enhance VRT handling for maps and DEM


kiozen:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

A lot of changes in the code from #1115 and #1125. Some code is duplicated. And using overview on DEM is now strongly encouraged.

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

Let's fix following aspects:

* Check if code unification is worth it
* Try to simplify / clean up the code in CDemVrt
* Try to speed up rendering in CDemVrt
* Add timeouts to both CDemVrt and CMapVrt. If rendering takes too long give the user guidance how to create overview levels.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

n/a

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

I will take care of it

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

n/a

PR #1129, opened 2026-06-24
[QMS-1128] Enhance VRT handling for maps and DEM

kiozen:

QMS-#1128

### What you have done:

* Read available overviews from sub-files, aka allow Frankenstein VRTs
* Speed up rendering, by optimizing the code for speed, without dropping readability too much.
* Add an advisory dialog if rendering takes too long

Simply use your DEMs with all kind of shading options (e.g. enable all).

Expectation: Everything works as expected plus a small speed up.

Also challenge the advisory dialog. If rendering takes more than 10s the dialog should show up.

- [x] yes

- [x] yes

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -250,4 +250,44 @@ isn't already axis-aligned with the target SRS. What fills those corners depends
   never check warp coverage explicitly. Hasn't been revisited — only matters for non-axis-aligned
   DEM sources.
 
+### CDemVRT/IDem rendering-speed punch list
+
+Working through this one item at a time (tracked as session tasks #1-8). Status as of
+2026-06-24: items 1-5 done (full low-risk batch complete); items 6-8 still deferred.
+
+**Active batch (low-risk, do first):**
+1. **DONE.** `IDem::maxElevationInWindow()` (`IDem.cpp:359-371`, used by `elevationLimit()`/
+   `elevationShading()`, called once per output pixel) goes through `IUnit::meter2elevation()`'s
+   *virtual* call plus an unused `QString` assignment (`unit = elevationUnit`) that's discarded
+   immediately. Since shading runs across multiple `threadPool` threads in parallel, every pixel
+   bumps the same shared `QString`'s atomic refcount - real cross-core contention, not just
+   wasted work. Fix: use the public `qreal IUnit::elevationFactor` member directly, skipping the
+   virtual call and the `QString` entirely. Preserve the `NOFLOAT` passthrough.
+2. **DONE.** `IDem::hillshading()` (`IDem.cpp:206-243`) computes `aspect = atan2(dy,dx)` then
+   `sin(aspect - azimuth)` per pixel. Since `azimuth` is fixed, use the identity
+   `r*sin(aspect-azimuth) = dy*cos(azimuth) - dx*sin(azimuth)` (`r = sqrt(xx_plus_yy)` already
+   computed) to drop `atan2`/`sin` entirely - 2 multiplies instead. Verify numerically equivalent,
+   including the `dx=dy=0` degenerate case (works out to 0 either way since `r` cancels
+   algebraically).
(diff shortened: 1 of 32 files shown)


kiozen:
@wthaem as you have a slow PC, can you test the branch and compare to current dev if there is a noticeable performance increase?

wthaem:
Sorry, my slow PC has gone. Now, I'm working on a much faster one. In addition to this: in the next few days i'll be on an excursion and thus no chance to test (except the heat in Germany).

kiozen:
ok, have fun!

kiozen:
@jmlzr @JFritzle @kkarsten62 @mitxel-m or to whom it may be of concern. The intended functionality of this PR is reached. Please test if you feel like it. 

The code still needs major review, testing and fine tuning.

wthaem:
My first quick test results of this PR:

* **DEM tests with different off- and online sources:**
    * DEM loading seems to be rather fast in general
    * WCS sources don't work as well as before:
        * BKG (Bundesamt für Kartographie) source shows the attached warning which didn't appear with earlier QMS versions. DEM data from this source are completely missing.

           WCS_BKG

        * Italy source seems to block due to extremely slow loading with this version (not noticed with earlier versions)
        * Other WCS sources are working as expected.
* **Map tests with various off- and online sources:**
    * The previously used zoom levels (the map region displayed in the map window) aren't reestablished
    * Fonts seem to be smaller
    * Loading of maps is pretty fast, especially for a PDF, MrSID, and MBTiles maps which used to have rather long loading times.
    * A KMZ map is now extremely blurry. Not noticed with earlier versions.

mitxel-m:
I'm very interested in this topic, but for a while now I've been working away from home on a laptop that doesn't have all the necessary dependencies to compile QMS, so I'll have to find some time to test it on my desktop computer when I get back.

I'll test it as soon as possible and also try to test it with different WCS services, as well as with VRTs that call more than one file.

kiozen:
@mitxel-m understood. Just open tickets if you find issues after the merge.

kiozen:
@wthaem did a wild guess on WCS. Please test again. If nothing changed I need those files.

However I think performance will decrease on any shading using WCS as outer zoom levels need more data from the server. This will hit performance issues or blockage. On the other hand side using online sources for shading was a broken concept right from the beginning.  This only works half well if you have a high speed connection to a high speed server..

wthaem:
@kiozen: attached you find 2 WCS files which I used for my tests and where I have found some issues. I don't have a clue whether the XML files with the same name are used by QMS/GDAL or not. Anyway, I included them.

* **BKG.wcs:** Works smoothly with the latest QMS release and without any error message. With latest version of this PR: there is still the above mentioned error message and the WCS doesn't work. Both QMS versions are started with the same GDAL! At the end of the WCS there is the line `<BandType>Float64</BandType>` what makes the error message understandable.
`
* **vDXfU.100.wcs:** (no idea where I found the strange name, should be Italy): Obviously, the server is quite slow. Waiting time for data (blocking the QMS GUI) is much smaller with the latest QMS release compared with this PR.

If I remember well, I started the tests with different WCS files which where overwritten at their first call by GDAL. The attached files are the overwritten ones, the original ones don't exist anymore. 

[DEM.zip](https://github.com/user-attachments/files/29339593/DEM.zip)

kiozen:
@wthaem I could fix the BKG.wcs. 

For the Italian wcs I couldn't find a difference. Imho it's a server side issue. After a few request in a row it blocks. Same behavior on a pre QMS-1115 binary and the current branch.

wthaem:
@kiozen: I can confirm: BKG.wcs working again (without error message).

kiozen:
From my point of view the PR is finished.  The code base is cleaned up. Doesn't mean that this code is easy to understand :wink: This whole GDAL and coordinate system bashing is a beast. However I hope the refactoring improved the readability and helps to maintain the code.

For the user the speed up and guidance when overviews are missing is a big plus. Let's see how this performs.

Issue #1130, opened 2026-06-24
Improve performance of IMG renderer


jmlzr:
### Is your feature request related to a problem? Please describe.
[comment]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] )

The rendering of IMG maps is quite slow. Especially on big screens panning and zooming is quite sluggish even on powerful hardware.
Since the DEM performance improvements the IMG renderer has became the bottleneck.

### Describe the solution you'd like
[comment]: # (A clear and concise description of what you want to happen.)

Improve performance of `CMapIMG::draw` and the functions below it. Possibly switch to a vector graphics engine with better performance.

### Describe alternatives you've considered
[comment]: # (A clear and concise description of any alternative solutions or features you've considered.)

n/a

### Give a reason why someone else should spend time on that request
[comment]: # (Keep in mind that you ask others to spend their spare time for your request. Good reasons are if you are able to prove that the request solves a fundamental problem for everyone, or the request is of general interest for many users. Bad reasons are you just want it to have that way, or argue that software xyz does it.)

n/a

### Additional context
[comment]: # (Add any other context or screenshots about the feature request here.)

n/a

kiozen:
Oh, now you are tackling the real nasty stuff. 😄 I apologize for the code. Most of it is over 20 years of age.

PR #1131, opened 2026-06-24
Replace QPainter with Blend2D in CMapIMG

jmlzr:

QMS-#1130

### What you have done:

Added Tracy as a profiler and instrumented the IMG rendering to look for where to focus my performance optimization efforts. That showed that Qt's rasterizer is the bottleneck that will dwarf any gains that could be made in the code. With the biggest bottlenecks being `drawPolylines` by and followed by `drawPolygons`.
So I decided to let Claude replace Qt's very slow rasterizer with Blend2D which is probably the fastest software rasterizer/vector graphics library in existence.

The new code seems to result in the same output but runs ~3 times faster!

##### Performance comparison

To make a comparison I recorded traces of me panning and zooming around the same area before and after the commit that switched to Blend2D and then used Tracy's compare feature.

`drawPolylines` saw a massive 5x improvement, dropping from 569ms mean to 117ms:

image

`drawPolygons` saw a also impressive 2.5x improvement, from 170ms to 67ms:

image

Looking at `draw` is deceiving as the results of that are diluted by early exists from redraw requests.

#### Render output

I took comparison screenshots before of the same view before and after, map used is a Freizeitkarte IMG.

Before:

qms-1130-kufstein-qt

After:

qms-1130-kufstein-blend2d

As you can see there is virtually no visible difference! The only thing I can notice on this view is the POIs shifting a little but they seem to shift by a similar amount with the previous renderer as well when panning minimally so I'm not sure if that is even a regression.

#### Next steps

Since this is a rather large change and I haven't checked if Claude did stupid things I made this just a draft for now. I do encourage people to try it out though!

I think now that the biggest bottleneck is gone I could also squeeze out probably at least another ~20% from optimizing the rest of the code.

Btw the massive amount of `ISO C++ prohibits anonymous structs` warnings comes from the fact that our CMakeLists.txt unfortunately sets global compile options with `add_compile_options` that then affects subprojects as well. The better thing would be to set them only for our own targets with `target_compile_options`. But we can fix that if we get close to merging this, for now I didn't want to bother with that.

Open a IMG.

#### Tracy

If anyone wants to play with Tracy:

1. Reconfigure QMS with `-DTRACY_ENABLE=1 -DTRACY_ON_DEMAND=1`
2. Build Tracy, in the QMS build dir do:
```
cd _deps/tracy-src
cmake -B profiler/build -S profiler -DCMAKE_BUILD_TYPE=Release
cmake --build profiler/build --config Release -j$(nproc)
```
3. Start Tracy from the same dir with `./profiler/build/tracy-profiler` and connect to the running QMS instance.

- [x] yes

- [x] yes

No as only a draft for now.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -215,6 +215,21 @@ preview curve.
 - Below ~80 nodes the basis, not the penalty, limits the curve. That is where the quality setting
   changes the result.
 
+### IGarminStrTbl label cache (the `get`/`decode` split)
+
+Profiling (Tracy) showed label decoding (`strtbl.get`) was the single largest slice of
+`loadVisibleData` — an `mmap` (via `CFileExt::data()`) plus a codepage decode **per labelled object,
+re-done every frame** even though label content is immutable. To fix that, `IGarminStrTbl::get()` is a
+**non-virtual** cache wrapper over a bounded LRU `QCache<quint64, QStringList>` (key `(type << 32) |
+offset`); on a miss it delegates to the protected pure-virtual **`decode()`** (the per-coding work,
+implemented by `CGarminStrTbl6/8/Utf8`). Do not re-merge `get`/`decode` or make `get` virtual again —
+the split exists so the cache lives in exactly one place. The cap (`labelCacheMaxEntries`) bounds
+memory when panning huge gmapsupp files; a hit needs no `mmap` at all, which also shrinks the
+per-subfile `CFileExt::free()` `munmap` loop. The cache is touched only from the draw thread
+(`get()` ← `loadSubDiv` ← `loadVisibleData` ← `draw`, all serialized), so it needs no lock.
+Tracy zones: `strtbl.get` = wrapper (hits+misses), `strtbl.decode` = misses only — compare their call
+counts to read the hit rate.
+
 ---
 
 ## Architecture: tree item delegates
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -236,6 +236,49 @@ if(APPLE)
 endif(APPLE)
 
 
+include(FetchContent)
+
+###############################################################################################
+# Get Blend2D and it's dependency AsmJit
+###############################################################################################
+# asmjit does not do official releases so pinned it to what was HEAD at time of writing this
+set(ASMJIT_GIT_TAG  "0bd5787b54b575ed94bf32ac452153b34385c514" CACHE STRING "AsmJit git revision to fetch")
+# blend2d does releases but not set git tags; the commit set here referese to release v0.21.2
+set(BLEND2D_GIT_TAG "def0d1238c3e5d0983bb848e5676049d829e435b" CACHE STRING "Blend2D git revision to fetch")
+
+# asmjit is Blend2D's JIT backend. Fetch its sources only: SOURCE_SUBDIR points at a
+# directory with no CMakeLists.txt so FetchContent_MakeAvailable populates without calling
+# add_subdirectory(). Blend2D adds asmjit itself (via ASMJIT_DIR) and, because we build it
+# statically, embeds it directly into the blend2d library.
+FetchContent_Declare(asmjit
+    GIT_REPOSITORY https://github.com/asmjit/asmjit.git
+    GIT_TAG        ${ASMJIT_GIT_TAG}
+    GIT_SHALLOW    FALSE # FALSE required when using a commit hash
+    GIT_PROGRESS TRUE
+    SOURCE_SUBDIR  __populate_only__
+)
+FetchContent_MakeAvailable(asmjit)
+set(ASMJIT_DIR "${asmjit_SOURCE_DIR}" CACHE PATH "AsmJit source directory" FORCE)
+
+set(BLEND2D_STATIC TRUE)
+FetchContent_Declare(blend2d
+    GIT_REPOSITORY https://github.com/blend2d/blend2d.git
+    GIT_TAG        ${BLEND2D_GIT_TAG}
+    GIT_SHALLOW    FALSE
+    GIT_PROGRESS TRUE
+)
+FetchContent_MakeAvailable(blend2d)
+
+# Blend2D's public headers use constructs our warning flags reject (anonymous structs vs
+# -Wpedantic, ...). Re-publish its interface include dirs as system to supress any warnings.
+# Once we bump our minimum cmake version to >= 3.25 we can instead just declare it `SYSTEM`
+# in the `FetchContent_Declare` call and get rid of this.
+get_target_property(BLEND2D_IFACE_INCLUDE_DIRS blend2d INTERFACE_INCLUDE_DIRECTORIES)
+if(BLEND2D_IFACE_INCLUDE_DIRS)
+    set_target_properties(blend2d PROPERTIES
+        INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${BLEND2D_IFACE_INCLUDE_DIRS}")
+endif()
+
 ###############################################################################################
 # Create library from Garmin's FIT SDK
 ###############################################################################################
--- a/changelog.txt
+++ b/changelog.txt
@@ -9,6 +9,7 @@ V1.XX.X
 [QMS-1123] Build error on platform Linux Ubuntu 24.04.4
 [QMS-1125] Use GDAL's warp and scale API to draw *.vrt maps
 [QMS-1128] Enhance VRT handling for maps and DEM
+[QMS-1130] Improve performance of IMG renderer + Fix: patterned areas shifting when panning
 [QMS-1132] Add action to move a map to the top in map list (macOS)
 [QMS-1135] Add guidance to improve overviews for DEM and map VRT
 [QMS-1139] Reject non-UTF-8 encoded VRT files with a clear error message
--- a/src/qmapshack/CMakeLists.txt
+++ b/src/qmapshack/CMakeLists.txt
@@ -1068,6 +1068,7 @@ target_link_libraries(${APPLICATION_NAME} PRIVATE
     ROUTINO::ROUTINO
     QuaZip::QuaZip
     GarminFit
+    blend2d::blend2d
 )
 
 if(APPLE)
--- a/src/qmapshack/CMakeLists.txt
+++ b/src/qmapshack/CMakeLists.txt
@@ -590,6 +590,7 @@ set( HDRS
     helpers/Platform.h
     helpers/Signals.h
     helpers/Tristate.h
+    map/Blend2dUtil.h
     map/CMapDraw.h
     map/CMapGEMF.h
     map/CMapIMG.h
--- a/src/qmapshack/map/Blend2dUtil.h
+++ b/src/qmapshack/map/Blend2dUtil.h
@@ -1,0 +1,64 @@
+/**********************************************************************************************
+    Copyright (C) 2026 Jens Mölzer <moelzerjens@gmail.com>
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+**********************************************************************************************/
+
+#ifndef BLEND2DUTIL_H
+#define BLEND2DUTIL_H
+
+#include <blend2d/blend2d.h>
+
+#include <QImage>
+#include <cstring>
+
+/**
+   @brief Deep-copy a QImage into an owning premultiplied BLImage.
+
+   @param img source image in any format; inputs are converted on the fly if necessary
+   @return the owning image, or an empty BLImage on failure
+ */
+inline BLImage toOwnedBLImage(const QImage& img) {
+  BLImage out;
+  if (img.isNull()) {
+    return out;
+  }
+
+  QImage storage;
+  const QImage& src = (img.format() == QImage::Format_ARGB32_Premultiplied)
+                          ? img
+                          : (storage = img.convertToFormat(QImage::Format_ARGB32_Premultiplied));
+
+  if (out.create(src.width(), src.height(), BL_FORMAT_PRGB32) != BL_SUCCESS) {
+    return out;
+  }
+
+  BLImageData dst;
+  if (out.make_mutable(&dst) != BL_SUCCESS) {
+    return BLImage();
+  }
+
+  const qsizetype bpl_src = src.bytesPerLine();
+  const qsizetype bpl_dst = static_cast<qsizetype>(dst.stride);
+  const qsizetype rowBytes = qMin(bpl_src, bpl_dst);
+  const uchar* const s = src.constBits();
+  uchar* const d = static_cast<uchar*>(dst.pixel_data);
+  for (int y = 0; y < src.height(); ++y) {
+    memcpy(d + bpl_dst * y, s + bpl_src * y, static_cast<size_t>(rowBytes));
+  }
+  return out;
+}
+
+#endif  // BLEND2DUTIL_H
--- a/src/qmapshack/map/CMapIMG.cpp
+++ b/src/qmapshack/map/CMapIMG.cpp
@@ -123,23 +123,13 @@ namespace {
 /// Convert a QColor to a Blend2D non-premultiplied 0xAARRGGBB colour.
 inline BLRgba32 toBLColor(const QColor& c) { return BLRgba32(c.rgba()); }
 
-/// Build a Blend2D path from a polyline. When @p close is set the path is closed
-/// so it can be filled and its closing edge stroked (matching QPainter::drawPolygon).
-BLPath polyToPath(const QPolygonF& poly, bool close) {
-  BLPath path;
-  const int n = poly.size();
-  if (n == 0) {
-    return path;
-  }
-  path.move_to(poly[0].x(), poly[0].y());
-  for (int i = 1; i < n; ++i) {
-    path.line_to(poly[i].x(), poly[i].y());
-  }
-  if (close) {
-    path.close();
-  }
-  return path;
-}
+// QPointF and BLPoint are both two contiguous doubles, so a QPolygonF's storage can be
+// handed to Blend2D's bulk polygon/polyline calls without copying into a BLPath.
+static_assert(sizeof(QPointF) == sizeof(BLPoint), "QPointF must be layout-compatible with BLPoint");
+static_assert(sizeof(QPointF) == 2 * sizeof(double), "QPointF must be two doubles");
+
+/// View a QPolygonF as a contiguous array of BLPoint for Blend2D bulk geometry calls.
+inline const BLPoint* toBLPoints(const QPolygonF& poly) { return reinterpret_cast<const BLPoint*>(poly.constData()); }
 
 BLStrokeCap toBLCap(Qt::PenCapStyle cap) {
   switch (cap) {
@@ -1428,9 +1418,9 @@ void CMapIMG::loadVisibleData(bool fast, polytype_t& polygons, polytype_t& polyl
 
       map->convertRad2Px(poly);
 
-      ctx.setStrokeWidth(2);
-      ctx.setStrokeStyle(BLRgba32(0xFFFF00FFu));  // magenta
-      ctx.strokePath(polyToPath(poly, true));
+      ctx.set_stroke_width(2);
+      ctx.set_stroke_style(BLRgba32(0xFFFF00FFu));  // magenta
+      ctx.stroke_polygon(toBLPoints(poly), static_cast<size_t>(poly.size()));
 #endif  // DEBUG_SHOW_SECTION_BORDERS
     }
 
@@ -1447,9 +1437,9 @@ void CMapIMG::loadVisibleData(bool fast, polytype_t& polygons, polytype_t& polyl
 
     QPolygonF poly;
     poly << p1 << p2 << p3 << p4;
-    ctx.setStrokeWidth(1);
-    ctx.setStrokeStyle(BLRgba32(0xFF000000u));  // black
-    ctx.strokePath(polyToPath(poly, true));
+    ctx.set_stroke_width(1);
+    ctx.set_stroke_style(BLRgba32(0xFF000000u));  // black
+    ctx.stroke_polygon(toBLPoints(poly), static_cast<size_t>(poly.size()));
 #endif  // DEBUG_SHOW_SUBDIV_BORDERS
 
 #ifdef Q_OS_WIN32
@@ -1697,10 +1687,31 @@ void CMapIMG::drawPolygons(BLContext& ctx, polytype_t& lines) {
   // QPainter::drawPolygon fills using the odd-even rule; Blend2D defaults to non-zero.
   ctx.set_fill_rule(BL_FILL_RULE_EVEN_ODD);
 
-  const int N = polygonDrawOrder.size();
-  for (int n = 0; n < N; ++n) {
-    quint32 type = polygonDrawOrder[(N - 1) - n];
-    const CGarminTyp::polygon_property& property = polygonProperties[type];
+  // Bucket the polygon indices by type once, so each draw-order pass visits only the
+  // polygons of its type instead of scanning the whole list once per type.
+  QHash<quint32, QVector<qsizetype> > byType;
+  for (qsizetype i = 0; i < lines.size(); ++i) {
+    byType[lines[i].type].push_back(i);
+  }
+
+  // Fallback for types that have geometry but no matching property (magenta hatch,
+  // known == false). Static so we never mutate polygonProperties while drawing.
+  static const CGarminTyp::polygon_property unknownProperty;
+
+  for (qsizetype n = polygonDrawOrder.size() - 1; n >= 0; --n) {
+    const quint32 type = polygonDrawOrder[n];
+    const auto bucket = byType.constFind(type);
+    if (bucket == byType.constEnd()) {
+      continue;
+    }
+
+    const auto propIt = polygonProperties.constFind(type);
+    const CGarminTyp::polygon_property& property =
+        (propIt != polygonProperties.constEnd()) ? propIt.value() : unknownProperty;
+    if (!property.known) {
+      qDebug() << "unknown polygon" << Qt::hex << type;
+    }
+
     const QBrush& brush = night ? property.brushNight : property.brushDay;
 
     // Configure the fill style once per type. Solid colours map directly; texture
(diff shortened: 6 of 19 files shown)


kiozen:
First quick impression:

* I am naturally a bit uneasy about another huge dependency. But let's see what the package maintainers say.
* On my fairly weak private laptop I hardly notice any big difference. Maybe because reading the data  takes the most of the time. Or belnd2d can't get much out of the Intel graphics adapter. Can't tell

jmlzr:
From a package maintainance perspective those shouldn't be a big thing.
They are pulled in via cmake automatically and we manage their version by setting a git tag. Almost no distro packages these so we don't even have to handle the option to build against system versions.

Strange... For me it's the difference between often >1s per draw to ~300ms which is quite a large difference in usability.
Your GPU should not make a difference. Blend2D is a pure software rasterizer so runs only on CPU. With the profiler you could easily check what impact reading the data has
as the loading step will show up as it's own zone.

wthaem:
My attempt to build the PR jmlzr:QMS-1130 with my default CMake parameters failed on Windows 11:

~~~
FAILED: [code=2] src/qmaptool/CMakeFiles/qmaptool.dir/canvas/CCanvas.cpp.obj
C:\PROGRA~1\MICROS~1\18\COMMUN~1\VC\Tools\MSVC\1451~1.362\bin\Hostx64\x64\cl.exe  /nologo /TP -DAPPLICATION_NAME=qmaptool -DHELPPATH=doc/HTML -DQT_CORE_LIB -DQT_GUI_LIB -DQT_HELP_LIB -DQT_NETWORK_LIB -DQT_NO_DEBUG -DQT_SQL_LIB -DQT_WIDGETS_LIB -DUNICODE -DVER_MAJOR=1 -DVER_MINOR=20 -DVER_STEP=3 -DVER_TWEAK="" -DWIN32 -DWIN64 -D_CRT_SECURE_NO_WARNINGS -D_ENABLE_EXTENDED_ALIGNED_STORAGE -D_UNICODE -D_WIN64 -ID:\QtProjects\QMS\QMS4Qt6\build-ninja\src\qmaptool -ID:\QtProjects\QMS\QMS4Qt6\src\qmaptool -ID:\QtProjects\QMS\QMS4Qt6\build-ninja\src\qmaptool\qmaptool_autogen\include -ID:\QtProjects\QMS\QMS4Qt6\src\qmaptool\. -ID:\QtProjects\QMS\QMS4Qt6\src\qmaptool\..\common -external:ID:\QtProjects\QMS\QMS4Qt6\build-ninja -external:Id:\QtProjects\QMS\GISInternals\1944_260614\release-1944-x64\include -external:Id:\QtProjects\QMS\GISInternals\1944_260614\release-1944-x64\include\proj9 -external:IC:\Qt\6.11.0\msvc2022_64\include\QtWidgets -external:IC:\Qt\6.11.0\msvc2022_64\include -external:IC:\Qt\6.11.0\msvc2022_64\include\QtCore -external:IC:\Qt\6.11.0\msvc2022_64\mkspecs\win32-msvc -external:IC:\Qt\6.11.0\msvc2022_64\include\QtGui -external:IC:\Qt\6.11.0\msvc2022_64\include\QtNetwork -external:IC:\Qt\6.11.0\msvc2022_64\include\QtHelp -external:IC:\Qt\6.11.0\msvc2022_64\include\QtSql -external:W0 /wd4996 /wd4101 /wd4005 /wd4003 /EHsc /O2 /Ob2 /DNDEBUG /Zi /GL -std:c++20 -MD /MP -Zc:__cplusplus -permissive- -utf-8 /showIncludes /Fosrc\qmaptool\CMakeFiles\qmaptool.dir\canvas\CCanvas.cpp.obj /Fdsrc\qmaptool\CMakeFiles\qmaptool.dir\ /FS -c D:\QtProjects\QMS\QMS4Qt6\src\qmaptool\canvas\CCanvas.cpp
D:\QtProjects\QMS\QMS4Qt6\src\qmaptool\canvas\CCanvas.cpp(22): fatal error C1083: Datei (Include) kann nicht geöffnet werden: "tracy/Tracy.hpp": No such file or directory
[234/606] Building CXX object src\qmaptool\CMakeFiles\qmaptool.dir\overlay\gridtool\CGridSelArea.cpp.obj
ninja: build stopped: subcommand failed.
~~~

There are lots of new CMAKE variables. I used their default settings. More info is certainly needed to build successfully on Windows.

Must there any additional packages be downloaded before building?

Sorry for the message layout - I copied it from the output log.

jmlzr:
@wthaem Yes cmake will need to download a few packages via FetchContent. Can you check at the start of cmake output if that failed or something? If you run cmake directly it should normally already fail at that point though... If you had the build directory already before switching to this branch you could maybe try cmake with `--fresh` to force a reconfigure.

kiozen:
same problem on a Linux machine. Even with cmake from scratch. The tracy header path is not set properly for the target.

wthaem:
I'm starting the build always with `cmake --fresh`. Now, I deleted in a first step everything from the build directory, too. This changed the situation a bit: Some dependencies were obviously found (asmjit, blend2d, tracy). Now there appears `-- Could NOT find rocprofiler-sdk (missing: rocprofiler-sdk_DIR)` as new message. Configuring and generating is ok. But building fails again with a message similar to the one communicated yesterday:

~~~
FAILED: [code=2] src/qmaptool/CMakeFiles/qmaptool.dir/canvas/CCanvas.cpp.obj
C:\PROGRA~1\MICROS~1\18\COMMUN~1\VC\Tools\MSVC\1451~1.362\bin\Hostx64\x64\cl.exe  /nologo /TP -DAPPLICATION_NAME=qmaptool -DHELPPATH=doc/HTML -DQT_CORE_LIB -DQT_GUI_LIB -DQT_HELP_LIB -DQT_NETWORK_LIB -DQT_NO_DEBUG -DQT_SQL_LIB -DQT_WIDGETS_LIB -DUNICODE -DVER_MAJOR=1 -DVER_MINOR=20 -DVER_STEP=3 -DVER_TWEAK="" -DWIN32 -DWIN64 -D_CRT_SECURE_NO_WARNINGS -D_ENABLE_EXTENDED_ALIGNED_STORAGE -D_UNICODE -D_WIN64 -ID:\QtProjects\QMS\QMS4Qt6\build-ninja\src\qmaptool -ID:\QtProjects\QMS\QMS4Qt6\src\qmaptool -ID:\QtProjects\QMS\QMS4Qt6\build-ninja\src\qmaptool\qmaptool_autogen\include -ID:\QtProjects\QMS\QMS4Qt6\src\qmaptool\. -ID:\QtProjects\QMS\QMS4Qt6\src\qmaptool\..\common -external:ID:\QtProjects\QMS\QMS4Qt6\build-ninja -external:Id:\QtProjects\QMS\GISInternals\1944_260614\release-1944-x64\include -external:Id:\QtProjects\QMS\GISInternals\1944_260614\release-1944-x64\include\proj9 -external:IC:\Qt\6.11.0\msvc2022_64\include\QtWidgets -external:IC:\Qt\6.11.0\msvc2022_64\include -external:IC:\Qt\6.11.0\msvc2022_64\include\QtCore -external:IC:\Qt\6.11.0\msvc2022_64\mkspecs\win32-msvc -external:IC:\Qt\6.11.0\msvc2022_64\include\QtGui -external:IC:\Qt\6.11.0\msvc2022_64\include\QtNetwork -external:IC:\Qt\6.11.0\msvc2022_64\include\QtHelp -external:IC:\Qt\6.11.0\msvc2022_64\include\QtSql -external:W0 /wd4996 /wd4101 /wd4005 /wd4003 /EHsc /O2 /Ob2 /DNDEBUG /Zi /GL -std:c++20 -MD /MP -Zc:__cplusplus -permissive- -utf-8 /showIncludes /Fosrc\qmaptool\CMakeFiles\qmaptool.dir\canvas\CCanvas.cpp.obj /Fdsrc\qmaptool\CMakeFiles\qmaptool.dir\ /FS -c D:\QtProjects\QMS\QMS4Qt6\src\qmaptool\canvas\CCanvas.cpp
D:\QtProjects\QMS\QMS4Qt6\src\qmaptool\canvas\CCanvas.cpp(22): fatal error C1083: Datei (Include) kann nicht geöffnet werden: "tracy/Tracy.hpp": No such file or directory
[320/831] Building CXX object src\qmaptool\CMakeFiles\qmaptool.dir\overlay\gridtool\CGridSelArea.cpp.obj
ninja: build stopped: subcommand failed.

~~~

The file `Tracy.hpp` can be found in the directory `...\build-ninja\_deps\tracy-src\public\tracy\Tracy.hpp`.

Edit: ChatGPT info: "Does ROCprofiler-SDK exist on Windows?

Officially: no, not as a supported Windows package."


How to switch off the use of tracy?

My settings:

~~~
TRACY_ENABLE:BOOL=OFF

TRACY_ON_DEMAND:BOOL=OFF
~~~

jmlzr:
I'm sorry I should have read your error message more carefully! Turns out I accidentally added an include to a qmaptool file (and qmaptool is not linked with Tracy). Since I always just built the qmapshack target I did not notice.

I fixed that by removing the include and confirmed that now all our targets build successfully.

You can ignore that missing rocprofiler. That's just an optional dependency of Tracy we don't need.

wthaem:
Now building is ok and nothing special noticed with a quick first test (KMz now again without blurry).

jmlzr:
@wthaem did you maybe comment on the wrong PR? Because the blurry KMZ maps was something you noted in #1129 so it won't affect this branch yeah...

Did you try IMG maps? Noticed a performance improvement with those? Or visual issues?

wthaem:
Sure: the KMZ topic belongs to the other PR. But I noticed again the difference in quality with this PR.

I tried IMG maps and many types of raster maps. Loading, zooming, panning of IMGs was pretty quick. Image quality on my notebook screen quite good. My notebook has a good CPU!

JFritzle:
Currently, I am trying to build on macOS. I get tons of warnings like these:
```
/Users/user/qms1130/build_QMapShack/_deps/blend2d-src/blend2d/compression/deflatedecoder_p.h:281:5: warning: anonymous structs are a GNU extension [-Wgnu-anonymous-struct]
  281 |     struct {
      |     ^
/Users/user/qms1130/build_QMapShack/_deps/blend2d-src/blend2d/compression/deflatedecoder_p.h:285:5: warning: anonymous structs are a GNU extension [-Wgnu-anonymous-struct]
  285 |     struct {
      |     ^
/Users/user/qms1130/build_QMapShack/_deps/blend2d-src/blend2d/compression/deflatedecoder_p.h:281:5: warning: anonymous types declared in an anonymous union are an extension [-Wnested-anon-types]
  281 |     struct {
      |     ^
/Users/user/qms1130/build_QMapShack/_deps/blend2d-src/blend2d/compression/deflatedecoder_p.h:285:5: warning: anonymous types declared in an anonymous union are an extension [-Wnested-anon-types]
  285 |     struct {
      |     ^
```
Might it be possible to expand cmake configuration not to throw these warnings?

jmlzr:
@JFritzle Pretty sure that's the same issue I already described in the PR description and Apple's clang just uses a different wording for that warning then GCC:
>Btw the massive amount of ISO C++ prohibits anonymous structs warnings comes from the fact that our CMakeLists.txt unfortunately sets global compile options with add_compile_options that then affects subprojects as well. The better thing would be to set them only for our own targets with target_compile_options. But we can fix that if we get close to merging this, for now I didn't want to bother with that.

JFritzle:
Did not notice a performance improvement with map _Reit_Wanderkarte_Italia_11_25.img_ on my Windows 11 PC:
Intel Core i5-2400 CPU, 4 logical processors, 22 GB RAM.

kiozen:
Haven't had the guts to deal with Tracy yet. But my hands-on experience is similar. On my high performance work laptop I can notice a small difference in performance. On my old private laptop no noticeable  gain.

But on both machines I have to increase the details level quite a lot to get a noticeable lag. More than I usually would do. So the benefit is on the lower side

I am a bit undecided. Pulling in new dependencies is always a risk and makes me feel uneasy. On the other hand side I highly appreciate and acknowledge every work donated to QMapShack. This places me in between chairs. It's always a hard decision based on the estimated future of the feature. Will it cause more benefit than future trouble? 

There is already quite some stuff I would like to get rid of (alglib, quazip) either by kicking out the feature or replacing it with native code. Dropping features is always trouble because of this one user popping up actually using it. And replacing code is another problem on it's own. For example what happens if I have to decide that blend2d can't be used anymore for whatever reason. I have to revert to Qt. Users might notice the performance loss and complain.  Restoring the former code is possible. But might be more trouble than expected.

So there must be a benefit worth it. For example replacing the QMapShack  native code with the Garmin SDK was a tough decision, too.  Apart from a not 100% clarified license situation, it's now depending on Garmin's lib. On the other hand side, the native code was done ages ago by someone not reachable any more. The newer devices implement new features that require to actually notice the change (reported by users), understanding the rather complex code and fixing the stuff without breaking something. To get rid of this maintenance burden picking the Gamrin Lib was a choice.

So long story short: I am still undecided. Even after writing all of this.

jmlzr:
That's very strange. I guess you guys must be either using maps only in areas with very little features or have screens with low resolution if you don't have noticeable lag with the previous renderer.
On a 3840x2160 screen and with a AMD 5600G with details on 0 already the map feels sluggish on areas without that much going on. On areas with much going on - worst case a big city - it get very annoying (>=1s per update). If I turn details up it gets pretty much unusable.
I checked a Wanderreitkarte and that has even worse performance than a Freizeitkarte.

@kiozen I don't get your point tbh. If you ever wanna get rid of Blend2D you can just revert. It will only be used in the IMG renderer and that one is unmaintained since a while anyways, so not even like there is much of a chances of there being changes that then might need back porting. Most importantly: Why would a user complaining about performance in the future after a hypothetically possible revert matter more than one complaining now about current performance?

kiozen:
Tested on a 2560x1440 screen. With Freizeitkarte and a big City like Rome.  I am not really that much of a fan of very large screens. I appreciate multiple  (cheap) screens. But that's just personal taste.

I passed the problem of a speed up vs dependency to Claude, asking if a performance gain could be achieved with Qt, too. 

This is the first answer.  Of course this has to be taken with a grain of salt. This is not a proven enhancement with measured performance gain. But it seems to be interesting enough to give it a try.  Let's see how much development time I can spare on that the next week.

PR #1131 — Replace QPainter with Blend2D in CMapIMG

The PR substitutes Blend2D for Qt's software rasterizer in CMapIMG's geometry pipeline (drawPolygons, drawPolylines, drawPoints, drawPois) while keeping text on QPainter. The benchmark numbers are real (5× polyline, 2.5× polygon), but the implementation has several correctness gaps and the core architectural question — is Blend2D the right fix — deserves scrutiny.

Findings:

1. applyPen silently drops standard Qt dash styles → solid-line regression
applyPen handles only Qt::SolidLine and Qt::CustomDashLine. Qt::DashLine, Qt::DotLine, Qt::DashDotLine, Qt::DashDotDotLine fall through and Blend2D gets no dash array, so all pre-defined dashed road/area-border types render as solid lines. Garmin TYP files use these heavily. This is a visible rendering regression on any map tile that uses preset pen styles.
2. Format_ARGB32 premultiply round-trip is lossy
draw() calls buf.image.convertTo(QImage::Format_ARGB32_Premultiplied) at entry and convertTo(QImage::Format_ARGB32) at exit. If any previous layer already painted semi-transparent content into buf.image, the multiply-then-divide round-trip loses low bits of the RGB channels. The correct fix is to allocate buf.image as Format_ARGB32_Premultiplied from the start (IDrawContext.cpp:109,113) so no in-place conversion is needed — this also speeds up all other layers' QPainter work.
3. brushToTile creates a new QImage per polygon per frame for textured fills
Hatch/texture brushes call brushToTile() on every drawPolygon() invocation. For a lake or forest area, this allocates and renders a tile image on every frame. A per-type cache keyed on the brush would avoid all but the first allocation.
4. polyToPath has undefined behaviour on empty polygons
poly[0] is accessed unconditionally before the loop that iterates segments. A zero-point polygon (degenerate data in some Garmin tiles) causes UB.

---
Can Qt alone close most of this gap?

Yes — significantly. The Blend2D speedup is real, but it's partly chasing the symptom (slow rasterizer) while some of the root causes are algorithmic:

1. The pointAtPercent O(n²) problem — biggest single win
In drawPolylines (line 1617–1676), the pixmap-along-line code builds a QPainterPath from the polygon just to call path.pointAtPercent(t) for each segment. Each pointAtPercent() call walks the full path from the start to measure arc length — O(n) per call, O(n²) total. The PR itself fixes this by using poly[i]/poly[i+1] directly with std::atan2. This fix requires zero new dependencies and is likely where most of the "5× drawPolylines" comes from for longer roads. It can be cherry-picked without Blend2D.

2. Commented-out simplifyPolyline() — potentially even bigger
Lines 1546, 1611, 1743, 1772 all have // simplifyPolyline(...) commented out. If that function implements Douglas-Peucker with a ~0.5 px threshold, enabling it would reduce vertex counts by 70–90% at typical zoom levels, which speeds up both convertRad2Px() and everything QPainter does downstream. Do you know why it was disabled?

3. Buffer format: Format_ARGB32 → Format_ARGB32_Premultiplied
IDrawContext.cpp:109,113 allocates the draw buffer as Format_ARGB32. QPainter compositing onto premultiplied surfaces skips a per-pixel multiply step — the Qt docs explicitly note this as the fast path. Switching these two lines benefits every map layer, not just CMapIMG, at zero cost. (This also fixes the PR's lossy round-trip if Blend2D is ever merged.)

4. p.save()/translate/rotate/drawImage/p.restore() per pixmap segment
The pixmap-along-line loop does 5 QPainter calls per road segment. Computing the rotation transform directly (QTransform::fromTranslate + rotate) and calling p.setTransform() once, or better, pre-rotating the pixmap tile and using drawImage(QPointF, ...) with no transform, removes the save/restore overhead.

5. Batching drawPolygon per type
drawPolygons calls p.drawPolygon(poly) individually for every polygon of the same type (same pen+brush). Accumulating them into a single QPainterPath with addPolygon() and calling p.fillPath() / p.drawPath() once per type reduces QPainter state machine overhead.

Bottom line: Items 1–3 alone could realistically recover 50–70% of Blend2D's measured gain with no new dependency, no premultiply round-trip hazard, and no hybrid-API surface to maintain. I'd suggest doing those first (especially re-enabling simplification if it's safe), profiling again, and then deciding whether the remaining gap justifies pulling in Blend2D.

---

jmlzr:
A grain of salt is underselling it. You have to be really careful with letting Claude think where it can make performance gains. It's wrong a lot of time cause it misjudges the costs of things. For example I can already tell you that this pixel format conversion is way too cheap to matter. Also I wanted to look into if changing the format of the double buffers in IDrawContext would be possible without regressions, as even the Qt docs say that for composition work the premultiplied is preferred.

I'm very doubtful on the other points as well as I spent quite some time before the port to Blend2D with letting it optimize the Qt renderer and I got to only around 20%. Plus I could see in Tracy that almost all time was spent in Qt's rasterizer. If you look at benchmarks you will also see that Qt is just very very slow: https://blend2d.com/performance.html

Maybe seeing the Blend2D port gave it ideas that indeed would make a difference with Qt so please try it if you'd like but I'd be very surprised if it got anywhere close.

I found that Claude gets way better at performance optimization if you let it instrument all it's suspicions and run the program with Tracy recording it. Then you ether tell it what the expensive zones are or even just feed it `tracy-csvexport` output and let it implement what it things makes most sense from that. Then you do another Tracy run to check if you actually gained anything and guide the next optimization step. Loop as long as you wish.

I have a few more commits for this branch local that I still need to review and clean up before I push. Got some more performance out of it, made some refactoring plus I fixed that bug (that was in the original Qt version and been bugging me since forever) of patterned fills shifting around.

kiozen:
Ok, agreed. Fixing it on the Qt side does not hit the same performance by quite a gap compared to Blend2d.  So the only option to speed up rendering is to rely on non Qt code. Go ahead.

jmlzr:
Rebased on current dev and added most of the commits I mentioned that I had by now reviewed and cleaned up.

I don't really like the [cache focus-in-metres in IDrawContext for convertRad2Px](https://github.com/Maproom/qmapshack/pull/1131/commits/a95a053b1ddffa321e189ad81f378a2821f31e0a) commit, might remove that again. Also gave only 1-2% performance. `convertRad2Px` really needs a refactor though. It most likely has a bug too: It checks the y axis for wrap-around but then applies the fix to the x axis. (that bug exists in current dev and this commit preserves that behavior)

kiozen:
If you change anything in those central methods like convertRad2Px please do it in it's own PR. I want this completely isolated from anything else to revert it easily if anything breaks

jmlzr:
Did roughly the following:

* Removed the commit that touched convertRad2Px
* Got rid of the flood of warnings that was introduced by Blend2D
* Pin the new dependencies to release versions
* Take advantage of Blend2D's multithreaded renderer
* Do text rendering now also in Blend2D

The last 2 points each provide another significant performance boost. Combined I now see a nice ~5x speedup for a full draw compared to dev.

I'm quite happy so far! The change in subjective performance feel is huge and for those who didn't care so much about a smooth experience the fix to patterned fills alone should be worth it.
I might experiment with some further tweaks but I think it's mostly done now and the next step will be the final polish to get this ready to be merged.
Would love to hear some feedback especially if anyone notices any regressions.

kiozen:
The polygon fills do not shift anymore! :laughing: 

Tell me when you have cleaned up the code and everything else and I will run a deep review. 

If, by any chance, you have a good idea how to fit the strings nicely to the polylines I would appreciate it very much. That is one of the topics I gave up.

jmlzr:
> Tell me when you have cleaned up the code and everything else and I will run a deep review.

The code in the existing commits is already cleaned up and ready for a deep review. If anything I was gonna add additional commits but thinking about it this PR is already big enough so I shouldn't keep adding more anyways.

So I just dropped the commit that conflicted with the changes from #1179 and rebased onto dev.

> If, by any chance, you have a good idea how to fit the strings nicely to the polylines I would appreciate it very much. That is one of the topics I gave up.

Uh, yeah that is a non-trivial problem - possibly one of the most complex aspects of a map renderer. 😅
Adding support for non-latin script might also be nice someday but would complicate that even more.
I think both is way out of scope for this PR though.

wthaem:
@jmlzr: I failed to compile this PR on Windows 11. Here is the relevant part of the CMake log:

~~~
[0/7] Performing update step for 'blend2d-populate'
-- Fetching latest from the remote origin
fatal: unable to read tree (def0d1238c3e5d0983bb848e5676049d829e435b)
CMake Error at D:/QtProjects/QMS/QMS4Qt6/build-ninja/_deps/blend2d-subbuild/blend2d-populate-prefix/tmp/blend2d-populate-gitupdate.cmake:202 (execute_process):
  execute_process failed command indexes:

    1: "Child return code: 128"



FAILED: [code=1] blend2d-populate-prefix/src/blend2d-populate-stamp/blend2d-populate-update D:/QtProjects/QMS/QMS4Qt6/build-ninja/_deps/blend2d-subbuild/blend2d-populate-prefix/src/blend2d-populate-stamp/blend2d-populate-update
C:\WINDOWS\system32\cmd.exe /C "cd /D D:\QtProjects\QMS\QMS4Qt6\build-ninja\_deps\blend2d-src && "C:\Program Files\Microsoft Visual Studio\18\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" -Dcan_fetch=YES -DCMAKE_MESSAGE_LOG_LEVEL=VERBOSE -P D:/QtProjects/QMS/QMS4Qt6/build-ninja/_deps/blend2d-subbuild/blend2d-populate-prefix/tmp/blend2d-populate-gitupdate.cmake"
ninja: build stopped: subcommand failed.

CMake Error at C:/Program Files/Microsoft Visual Studio/18/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-4.3/Modules/FetchContent.cmake:1931 (message):
  Build step for blend2d failed: 1
Call Stack (most recent call first):
  C:/Program Files/Microsoft Visual Studio/18/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-4.3/Modules/FetchContent.cmake:1622 (__FetchContent_populateSubbuild)
  C:/Program Files/Microsoft Visual Studio/18/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-4.3/Modules/FetchContent.cmake:2158:EVAL:2 (__FetchContent_doPopulation)
  C:/Program Files/Microsoft Visual Studio/18/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-4.3/Modules/FetchContent.cmake:2158 (cmake_language)
  C:/Program Files/Microsoft Visual Studio/18/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-4.3/Modules/FetchContent.cmake:2399 (__FetchContent_Populate)
  CMakeLists.txt:289 (FetchContent_MakeAvailable)

-- Configuring incomplete, errors occurred!
~~~

I don't have a clue how to handle these errors. Are there any special steps required to handle Blender correctly?

jmlzr:
@wthaem No special steps required, will be downloaded and compiled automatically. Weird seems like it fails attempting to get the Blend2D source via git. Especially the "fatal: unable to read tree" sounds weird.

How did you call cmake? For me (on Linux) `cmake -B build -S .` works fine.

If you haven't done so already please try with a completely new build dir. If you reused the build dir from your previous tests of this PR that might cause issues as in the meantime I changed `GIT_SHALLOW` for blend2d to false and set a commit instead of a branch.

wthaem:
@jmlzr: Thanks for the hint. After deleting the build dir,  handling the Blender part in the config run seems to be ok. But due to kiozen's `GDAL::GDAL` issue discussed in the other thread the configuration still fails.

kiozen:
Do you really want to drag Tracy into a release build? I hope not.

jmlzr:
> Do you really want to drag Tracy into a release build? I hope not.

What do you mean? Unless you define `TRACY_ENABLE=1` all Tracy calls are compiled to no-ops.

kiozen:
Can we please remove all unnecessary stuff from the source and cmake files. In half a year no one will remember and just wonder what all those lines of peculiar, dead code are good for.

jmlzr:
Ah so you mean if I "really want to merge Tracy". That's a discussion that can be had and I too think it does not make much sense keeping that dependency if no other developer cares about the performance of QMS.
I'll remove Tracy before we merge this but I'd like to keep it until the last moment. If there are any code changes necessary I'd like to still have instrumentation available and I sure do not want to bother with fixing the merge conflicts caused by dropping the tracy commits more than once...

So - except for removing Tracy - are we ready to merge this PR?

kiozen:
> I'll remove Tracy before we merge this but I'd like to keep it until the last moment. 

:+1: 

Still need to check the review findings. So far all the "bugs" are either no real defect or a defect already in the code before the PR.

There are some performance issues pointed out, I would like to check. If you want I can just post the findings unfiltered.

kiozen:
There seems to be a valid point concerning performance of drawing points and POIs. 

```
drawPoints/drawPois call blitQImage() once per point, per frame. It runs toOwnedBLImage(), which deep-copies the icon into a new BLImage — and because POI icons are Format_Indexed8 (CGarminTyp.cpp:1061), it also runs a full convertToFormat first.

So per point, per frame: 2 allocations + 2 passes over the icon. The old p.drawImage(x, y, icon) did none.

Regression from this branch. The copy exists for a good reason (multithreaded renderer, create_from_data views dangle) — it's just done per blit instead of once per distinct icon.

Fix: cache the BLImage keyed on QImage::cacheKey(), thread_local, same as the glyph cache. A TYP has a few hundred distinct icons.
```
This might be worth looking at. I am not sure how much this really hits performance. Maybe a topic for a follow up PR

kiozen:
Ok. So far so good. At least from my point of view non of the findings is a serious show stopper.  Maybe the bug section deserves a second look. But I haven't seen anything serious that haven't been kind of broken before. I think CMapIMG deserves a bug fixing session in general. But this is not the scope of this PR.

I attached the list. Maybe you find one or the other point worth fixing. You are deeper in that code than me. 

[QMS-1130-review-findings.md](https://github.com/user-attachments/files/30901374/QMS-1130-review-findings.md)

jmlzr:
The `blitQImage` thing sounded valid indeed. Although the suggested fix you got is strange (I'm starting to notice that Claude seems a little obsessed with caches). The cleaner fix is having `CGarminTyp` hold `BLImages` instead of the `QImages`.

Unfortunately my tests show no performance improvement but I'd still keep it. (that's why I wanted to keep Tracy for now)

> I attached the list. Maybe you find one or the other point worth fixing. You are deeper in that code than me.

Did that `FALLBACK_TO_SYNC` even though it probably does not really matter. Otherwise it sounds like nit picking that's not worth the effort.

> Ok. So far so good. At least from my point of view non of the findings is a serious show stopper.

Okay then I'll remove Tracy next so we can merge this.

jmlzr:
Tracy is gone. From my side this is now ready to merge.

kiozen:
Ok, let's merge it before it starts to get smelly and grows tiny white hairs.

Issue #585, opened 2023-02-13
GPX no more compatible with Osmand


XR-01:
### Describe the bug
When I open a GPX with Osmand (Android navigation App), I can't see some track, randomly. I found that every time I save the same GPX with QMapShack, will be overwrited the header of the GPX (with all the xmlns). Without the string (xmlns:osmand="https://osmand.net/") Osmand can't open correctly the GPX. I think is a problem with the last QMapShack version (1.16.1) because with older version I never had this problem.

### What have you done to circle down the problem?
I have to put this string:

xmlns:osmand="https://osmand.net/"

at the end of the head of the GPX, and with this tip the GPX is perfectly readed in Osmand.

### To Reproduce
Open the attached GPX and save it: the string (xmlns:osmand="https://osmand.net/") will be overwritted.

### Attachments
[prova.zip](https://github.com/Maproom/qmapshack/files/10721190/prova.zip)
### Desktop
[comment]: # (please complete the following information)

 - OS: Windows 10
 - QMapShack Version:  1.16.1

### Additional context

Thank you for your work!

kiozen:
First glance into the problem: This is really an issue of QMapShack. The attached file is using extensions with the osmand namespace therefore it has to provide it in the list of extensions to create a valid GPX file.

kiozen:
Can you please test the PR #589 if it fixes the problem

XR-01:
I have done the test, when I save the GPX from osmand the Osmand xmlns will be overwritten. Osmand can open the file, but if there are a lot of tracks Osmand don't show some tracks. If I manually add the xmlns to the GPX Osmand can open it without problems

robert7k:
It seems I still have the same issue with QMapShack 1.20.3

PR #1151, opened 2026-07-10
[QMS-585] Preserve extra XML namespaces in GPX files

robert7k:

QMS-#585

### What you have done:
Previously, QMapShack would strip existing namespaces when saving a GPX file, but keep the XML tags using these namespaces, resulting in invalid XML.

This patch makes sure to retain these namespaces.

1. Open [the GPX file attached to QMS-#585](https://github.com/Maproom/qmapshack/files/10721190/prova.zip) in QMapShack
2. Save the GPX file
3. Compare the file with the old version: make sure all namespaces are retained (they might be in a different order). Look out for `xmlns:osmand="https://osmand.net"`  in the `<gpx>` tag.

Additional test:
1. run xmllint on the test file: `xmllint --noout prova.gpx`
2. Open the file in QMapShack, save it
3. run xmllint again, confirm there are no errors.

- [x] yes

- [x] yes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,5 @@
 V1.XX.X
+[QMS-585] Preserve extra XML namespaces in GPX files
 [QMS-902] Add HiDPI support
 [QMS-963] Add action to move a map to the top in map list
 [QMS-1105] Fix: Context menu shortcuts not shown (macOS)
--- a/src/qmapshack/gis/gpx/CGpxProject.cpp
+++ b/src/qmapshack/gis/gpx/CGpxProject.cpp
@@ -18,6 +18,7 @@
 
 #include "gis/gpx/CGpxProject.h"
 
+#include <QSet>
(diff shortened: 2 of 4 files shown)


kiozen (commented):
The only thing that might be worth considering:

If the file uses one of the well known internal namespaces it will override the internal namespace with what ever is defined. I would say such an input is broken. But on the other hand side it's easily fixed with a small guard. 

Up to you. Just give feedback.

robert7k:
> The only thing that might be worth considering:
> 
> If the file uses one of the well known internal namespaces it will override the internal namespace with what ever is defined. I would say such an input is broken. But on the other hand side it's easily fixed with a small guard.
> 
> Up to you. Just give feedback.

Yes, good point. However, I think we should leave the namespace as it is: it might refer to a newer version and use tags defined in it, so if reverted the namespace to the previous value we might end up with invalid XML again.

Issue #1171, opened 2026-08-03
Formated text is rendered badly in dark theme


kiozen:
### Describe the bug
[comment]: # (A clear and concise description of what the bug is)

<img width="662" height="419" alt="Image" src="https://github.com/user-attachments/assets/54bd9985-1e65-4c1e-a5f4-b3def49320f1" />

I think this explains it best. Probably this happens everywhere html or rich text is displayed

### What have you done to circle down the problem?
[comment]: # (Did you compare to previous versions? What versions? Did you check if others have the same problem?)

These widgets do not handle theming properly.

### To Reproduce
[comment]: # (List the steps to reproduce the behavior)

* switch to dark mode
* open the overlay info dialog.

### Expected behavior
[comment]: # (A clear and concise description of what you expected to happen)

Read text without getting eye cancer

### Screenshots
[comment]: # (If applicable, add screenshots to help explain your problem)

n/a

### Attachments
[comment]: # (Can you provide an example file to trigger the problem?)

n/a

### Tracebacks
[comment]: # (Add your backtrace below if you have one. If QMapshack crashes the fastest way to get help is a backtrace.
For Linux see: https://github.com/Maproom/qmapshack/wiki/TroubleShooting#create-a-backtrace-of-a-crash-on-linux.
For Windows it would need a debug build and running QMapshack in Visual Studio.)

### Desktop
[comment]: # (please complete the following information)

 - OS: any
 - QMapShack Version:   any
 - Window Manager:  any

### Additional context
[comment]: # (Add any other context about the problem here)

PR #1172, opened 2026-08-03
[QMS-1171] Theme rich text colours so they work on a dark palette

kiozen:
Rich text, label stylesheets and setTextColor() hardcoded one half of the colour pair and took the other from the palette, which inverts on a dark scheme. CUiTheme supplies both halves per role for qmapshack and qmaptool, and CHelpBrowser applies it to the packaged help CSS.

QMS-#1171

### What you have done:

Add a theming engine for text resources. Use that engine everywhere rich formatted text is used.

* Open overlay info dialog
* Open help (F1)
* open any other widget with text

They all should render well in dark and light theme (fingers crossed).

Note: some widgets can't update on the fly. they need to be reloaded.

- [x] yes

- [x] yes

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -229,6 +229,39 @@ Do not run `cmake --build` or any build command — Oliver builds externally.
 
 Do not add `Co-Authored-By` lines to commit messages.
 
+### Rich text colours — `CUiTheme`
+
+`CUiTheme` (`src/common/theme/`) is the only source of status colours for rich text, label
+stylesheets and `setTextColor()`. Roles `Neutral/Ok/Warn/Error/Info/Code`, each a fixed light/dark
+pair, light arm as authored. Tune there, never per site.
+
+- **Set both colours or neither.** A hardcoded background inherits the palette's text colour and
+  inverts on dark. Normal text takes palette colours.
+- Pick the entry point by what the widget *is*:
+  - permanently a status message, only shown and hidden → `markLabel(label, role)`
+  - a value that is *sometimes* a status → `span()`/`spanBold()`, so the colour travels with the
+    string (`CSetupExtTools`, where one label alternates between "not found" and a path)
+  - fills a background (table cell, banner) → `css()`
+  - text on the widget's own background → `cssForeground()`/`foreground()`
+- Never bake a colour into a `tr()` string or a `.ui` `<string>` — the translation carries it.
+- `paletteIsDark()` is `inline` in `CUiTheme.h` so `CSvgtIconEngine::roleColor()` can share it with
+  nothing to link; the plugin needs only `target_include_directories(svgticonengine PRIVATE ..)`.
+  Every consumer of the light/dark split uses it — a second copy of the threshold drifts silently.
+- Re-render on a scheme switch from `QEvent::ApplicationPaletteChange || QEvent::PaletteChange`
+  (`CSvgtIcon`, `CHelpBrowser`, `CListTrkPts`), not from `QStyleHints::colorSchemeChanged`: the
+  colours resolve through `QGuiApplication::palette()`, and a palette event is delivered *because*
+  that palette changed. A widget may get both events, so a re-render can run twice per switch.
+- Printing: `const CUiTheme::ForceLight paperColours(printable)` in `CDetailsPrj::draw()`, beside
+  the two palette branches it completes. It does not touch `QPalette`, so palette colours still
+  need their own paper branch.
+- `QTextBrowser` applies a `<link>`ed stylesheet (`loadResource`, `StyleSheetResource`), so
+  `CHelpBrowser::loadResource()` appends a themed `code, pre` rule to whatever the packaged help
+  ships. That CSS lives outside this repo; `Role::Code`'s light arm is picked to match its `code`
+  background, so light mode renders unchanged.
+- Untouched on purpose: `CShell`/`IToolShell` colour each line as it is appended, so lines already
+  in the log keep the scheme they were written in; `IPlot`'s plot sheet, the halo
+  under `CWksItemDelegate`'s progress bar and `CIconGrid`'s tiles are white by design.
+
 ### IDrawContext — logical vs device pixels (HiDPI)
 
 `convertRad2Px()`/`convertPx2Rad()` work in **logical** viewport pixels (built from `center` and
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,6 +1,7 @@
 # QMapShack — Claude project notes
 
-QMapShack is a Qt/C++ desktop application for planning and analysing GPS tracks, routes, and waypoints. It loads map tiles, vector maps, and DEM data, and supports online and offline routing engines.
+Qt/C++ desktop app for planning and analysing GPS tracks, routes and waypoints. Loads map tiles,
+vector maps and DEM data; supports online and offline routing engines.
 
 ---
 
@@ -9,8 +10,9 @@ QMapShack is a Qt/C++ desktop application for planning and analysing GPS tracks,
 - **Language:** C++20
 - **GUI / framework:** Qt 6.8+
 - **Key libs:** GDAL, PROJ 8+, Routino, QuaZip-Qt6
-- **Build:** CMake 3.20+, Ninja; Debug build in `build/`; binaries in `build/bin/`
+- **Build:** CMake 3.20+, Ninja; Debug build in `build/`, binaries in `build/bin/`
 - **Bundled 3rdparty:** alglib, Garmin FIT SDK
+- **Minimum GDAL:** 3.10
 
 ---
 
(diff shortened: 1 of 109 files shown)


kiozen:
@wthaem @JFritzle @kkarsten62 I think this is a good point to search for more quirks. I tried my best ;)

kkarsten62:
> @wthaem @JFritzle @kkarsten62 I think this is a good point to search for more quirks. I tried my best ;)

I can help to test. But, please can you give me a short guide line:
- Waiting for your merge into `dev` or your branch `kiozen:QMS-1171`?
- How can I enable dark mode?
- What to test?
- Use cases?
- ...
Thanks.

wthaem:
Now, icons in light and dark mode look ok compared to the ones I saw a few days ago. No obvious contrast issues noticed anymore.

kiozen:
In Linux dark mode can be selected in the system setup of your window manager. For windows it's a command line parameter https://doc.qt.io/qt-6/qguiapplication.html#platform-specific-arguments

Test as much of the UI as possible. Are all text readable? Especially in all the on-screen stuff of the canvas. Anything else bad because colors are not switched properly and you get dark elements on a dark background. 

Issues I fond and fixed:

* Point table in the track detail dialog
* Point table in the range selection on screen toolbox.
* Project summary
* The table in overview advisory 
* The F1 help

And I found a new one: Waypoint Info Bubble. -> will fix

wthaem:
What about non-square icons in real-time window (GPS TCP/IP part and others)?

capture_20260804_095849

*Remark:* With "No contrast issues noticed" I mean not only icons but everything I can somewhere see in the GUI.

wthaem:
I do not yet understand why: but I get 2 different toolbar layouts - one is ok, the other one not:

Calling "qmapshack.exe" without any parameters (ugly variant!):

capture_20260804_102428


Starting QMS with some parameters (INI file, special workspace!) from a shortcut (nice variant):

capture_20260804_102348

kkarsten62:
To activate **Dark Mode** in Fedora 44, GNOME 50, Wayland: `GTK_THEME=Adwaita:dark qmapshack`

First issue. Text boxes are not really readable:
image

kkarsten62:
Next issue:
Icon menu is flickering when change text font and select a text (general problem already in previous versions):
image

kkarsten62:
Next issue:
Hard to identify which option is currently enabled:
image

kkarsten62:
> To activate **Dark Mode** in Fedora 44, GNOME 50, Wayland: `GTK_THEME=Adwaita:dark qmapshack`
> 
> First issue. Text boxes are not really readable

Sorry, I used the wrong binary QMS file. Already fixed.

kkarsten62:
I checked my uses cases and that is all I found on issues in dark mode theme.

JFritzle:
> In Linux dark mode can be selected in the system setup of your window manager. For windows it's a command line parameter ...

In _Windows_ dark mode can be selected in _Settings_:
_Settings -> Personalization -> Colors -> Choose your default app mode -> Dark_
grafik

In _macOS_ dark mode can be selected in _System Settings_:
_System Settings -> Appearance -> Dark_
grafik

JFritzle:
On platform Windows:

- Icons in _Set Map Icon Sizes_ dialog are hard to recognize:
  grafik
  -> potentially not replaced by SVG icons and/or not yet dark mode aware.

- Check boxes in map layer selection seem to be not yet dark mode aware:
  grafik

- Check boxes in POI selection seem to be not yet dark mode aware:
  grafik

JFritzle:
Very nice:
Switching _Windows_ dark mode to light mode while QMS is running, QMS switches from dark mode to light mode too and vice versa.

But:
Color of URL links does not switch between navy and cyan on the fly. URLs keep their color at creation time.
Thus, URL links in a **new** view _Welcome_ dialog newly created get color corresponding to currently active light/dark mode.,
while BRouter's copyright URL links only once created keep their initial color at QMS startup, either navy or cyan:
grafik
grafik

JFritzle:
On platform macOS:

While application's top toolbar and toolbar pull downs follow _System Settings -> Appearance -> Dark_,
QMS main window and subwindows do not:
grafik

Web search tells:
- _Enabling dark mode for Qt applications on macOS is automatic by default using native styles, but requires handling Info.plist settings, style choices, or custom palettes for full control._
- _Ensure NSRequiresAquaSystemAppearance is not set to true in your app bundle's Info.plist file, as this explicitly opts out of dark mode._

-> Application's _Info.plist_ indeed contains line `NSRequiresAquaSystemAppearance = true;`.
This line currently is added intentionally by build script _MacOSX/build-all.sh_ to prevent application from dark mode.

After removing line from _Info.plist_, QMS shows almost perfect:
grafik

Almost perfect but not completely perfect, because icon toolbar does not follow dark mode.
This is certainly due to the fact, that icon toolbar has it's on style sheet, which has propably to be adjusted according to light/dark mode.

I will
- remove  line `NSRequiresAquaSystemAppearance = true;` from build script _MacOSX/build-all.sh_
- investigate how to adjust style sheet

Questions:
- how do I query light/dark mode at QMS start by Qt?
- how can I get notified by Qt if mode changes while QMS running?

wthaem:
@kiozen: the above mentioned differences in the toolbar layout seem to have as origin the use of the command line parameter `--style Fusion`. Without this parameter, the "ugly" variant is shown, with this parameter, the "nice" variant is shown. 

Is this the wanted behavior? If yes, then the use of the additional parameter is compulsory in order to avoid the ugly variant!?

edit: additional info from Copilot:

The default Qt widget style is platform‑native, meaning:

    If you do not pass --style, Qt uses the operating system’s native style for the current platform.

On Windows, this means:
⭐ Default style on Windows

Qt will use:

    WindowsVista style if available (Qt5, Qt6 on modern Windows)

    otherwise it falls back to Windows style

In practice, on Windows 10/11 with Qt5/Qt6, the default is almost always:

    WindowsVista


and the WindowsVista style is the ugly one! (my QT is up-to-date!)

JFritzle:
@wthaem
> Is this the wanted behavior?

This is the Qt built-in default behavior. Obviously wanted by Trolltech stylists.

> On Windows, this means:
⭐ Default style on Windows
 In practice, on Windows 10/11 with Qt5/Qt6, the default is almost always: WindowsVista

On Windows 10, Qt6 default style is _Windows_.
On Windows 11, "ugly" Qt6 default style is _Windows11_.

Style _Windows_ is available on Windows 11 too and may be forced by `--style Windows` . And it doesn't look that ugly:
grafik

wthaem:
@JFritzle: my main concern is that someone launching qmapshack.exe without any parameters (why not?) or with (default?) style Windows11 on a Windows 11 system like mine gets the "ugly" toolbar. Styles Window and fusion toolbars are ok. I don't have a clue whether it is possible to avoid the "ugly" variant of the toolbar. Would be nice, if yes.

JFritzle:
@wthaem: I regard style _Windows11_ ugly too and would never use it voluntarily except for testing purpose.

Have you ever tested work around (sorry for German)?

_Aktivierung des Windows 11 Stils ... per Konfigurationsdatei (qt.ini):
Ab Qt 6.7 können Sie den Standardstil ohne Code-Änderungen erzwingen:_
```
[Qt]
style=windows11
```
Replace _windows11_ by _fusion_ to get _fusion_ as standard style.

Update:
Either I did not find where to store file or it is not working at all. Brave new AI world.

wthaem:
@JFritzle: Copilot tells me: *No, you cannot preset the --style parameter in any Qt config file.
But you can preset the style using the environment variable QT_STYLE_OVERRIDE.*

kiozen:
> On platform Windows:
> 
>     * Icons in _Set Map Icon Sizes_ dialog are hard to recognize:
>       -> potentially not replaced by SVG icons and/or not yet dark mode aware.
> 

Fixed, good catch

>     * Check boxes in map layer selection seem to be not yet dark mode aware:
> 

can't confirm

>     * Check boxes in POI selection seem to be not yet dark mode aware:
>      

can't confirm.

Those check box items are pure Qt. They should be themed by the Qt framework properly. This is how they are rendered on my system (Linux):

grafik

kiozen:
> Very nice: Switching _Windows_ dark mode to light mode while QMS is running, QMS switches from dark mode to light mode too and vice versa.
> 
> But: Color of URL links does not switch between navy and cyan on the fly. URLs keep their color at creation time. Thus, URL links in a **new** view _Welcome_ dialog newly created get color corresponding to currently active light/dark mode., while BRouter's copyright URL links only once created keep their initial color at QMS startup, either navy or cyan: 

Fixed.

kiozen:
Some AI generated answer about the style problem on windows. Note: This has been derived on Linux, so statements are not 100% confirmed:

AI content ----->

@wthaem @JFritzle I ran the style-selection mechanisms against Qt 6.10.2 instead of
guessing, here is what actually holds:

| mechanism | result |
|---|---|
| `QApplication::setStyle()` in code | wins over everything |
| `-style Fusion` / `--style` | beats the environment variable |
| `QT_STYLE_OVERRIDE=Fusion` | works, lowest precedence of the three |
| `qt.conf` / `qt.ini` with `[Qt] style=…` | **ignored** |

So the `qt.ini` route is a dead end, @JFritzle — the `[Qt] style=` key does nothing, which
is why you could not get it to fire. It is not a matter of the file being in the wrong
place. @wthaem's `QT_STYLE_OVERRIDE` is the only way to preset a style without passing a
command line argument.

One trap if anyone still tries the config file: dropping a `qt.conf` next to the
executable re-roots Qt's plugin search path. My test run aborted with "could not find the
Qt platform plugin" until I added an explicit `[Paths] Plugins=…`. On Windows that would
break the application for a setting that has no effect anyway.

**A hypothesis on the two check box reports I could not confirm.** Those indicators are
drawn entirely by the active style, QMS paints nothing custom there — no delegate, no
style sheet, no icons. My screenshot was Linux, where the default style here is Fusion.
Both of your reports are Windows 11, where the default is `windows11` — the same style
this toolbar discussion is about.

@JFritzle could you re-check the map layer and POI check boxes with `--style Fusion` and
with `--style Windows`? If they render correctly there, all three reports share one root
cause, the `windows11` style, and none of them are QMS bugs.

**On forcing a style in QMS:** Note row 1 of the table — a
`setStyle("Fusion")` in `main.cpp` would take precedence over the user's own `--style`,
so @wthaem could then never select `windows11` even deliberately. Making that conditional
means parsing `argv` and `QT_STYLE_OVERRIDE` by hand, since Qt exposes no "was a style
requested" API. Picking a style is the user's call, not the application's.

<------------ end AI content

Last paragraph is a possible fix in the code. However it's another Lex Windows. As most Windows users simply use what the installer provides it might be a better solution to set the style in the installed Windows shortcut.

JFritzle:
> could you re-check the map layer and POI check boxes with --style Fusion and
with --style Windows?

Above screenshots with (nearly) invisible check boxes had been actually taken with `--style Fusion`:
grafik
-> Check box border can be seen with a magnifying glass:
grafik

Corresponding screenshot with `--style Windows`:
grafik

Corresponding screenshot with `--style Windows11`:
grafik

JFritzle:
> ... forcing a style in QMS ...

This is probably no good idea. Linux with KDE desktop manager might no longer be able to set KDE's own style.

kiozen:
> > ... forcing a style in QMS ...
> 
> This is probably no good idea. Linux with KDE desktop manager might no longer be able to set KDE's own style.

Well this should live in the Windows specific part of the code. After all it's "Lex Windows."

kiozen:
> Next issue: Hard to identify which option is currently enabled: 

should be better now.

JFritzle:
>  ... it might be a better solution to set the style in the installed Windows shortcut

Why not create 2 short cuts, one forcing style Fusion, other using default style?

kiozen:
I think I fixed all UI/Theming issues. I ignored the ones like button sizing etc.  

Anything left?

kiozen:
> > ... it might be a better solution to set the style in the installed Windows shortcut
> 
> Why not create 2 short cuts, one forcing style Fusion, other using default style?

I would tend to pin it to the style that works best. I mean theming and styling isn't really what Windows users are used to.

JFritzle:
> I would tend to pin it to the style that works best. I mean theming and styling isn't really what Windows users are used to.

Same with iMac users. 

On iMac, style _Fusion_ works best too. But unfortunately, default style is _macOS_.
Installing QMS application either in global or in user's local _Applications_ folder automatically adds a shortcut to _Launchpad_ ...
grafik
... without chance to force style _Fusion_.

JFritzle:
> Anything left?

Nothing from my side. Beside from check boxes.

kkarsten62:
> > Next issue: Hard to identify which option is currently enabled:
> 
> should be better now.

For me ok. Current selected can be seen by blue border on Linux Fedora 44, GNOME 50, Wayland.
image

kiozen:
ok, another try on the toggle buttons/actions.  Especially in menus. It's now a bit brute but very visible. I am still undecided if that is not too much

kiozen:
> > Anything left?
> 
> Nothing from my side. Beside from check boxes.

As this is rooted in the Fusion style that is shipped with Windows I fear we are out of luck if Fusion must be used. Unless there is another idea how to fix that on windows.

wthaem:
Shortcuts created with the Windows installer coming from my repo fork (another one I don't know) are configured with `--style fusion`. @JFritzle: Maybe, there could be a hint in the `qmapshack.exe -h` dialog with the recommendation to use this style on Windows.

JFritzle:
@wthaem: 
> Maybe, there could be a hint in the qmapshack.exe -h dialog ...

👍 Good proposal.
The help dialog is black box. Only program arguments can be defined. Neither `-h` nor `--help-all`
shows hidden option `--style` which is silently processed by Qt itself.
I added an argument `--style` which _help_ actually shows. Although command line parser does not return it's value, value is indeed processed correctly.
`qmapshack -h` on Windows 11 returns:
```
Usage: C:\Program Files\qmapshack.exe [options] files

Options:
  -?, -h, --help        Displays help on commandline options.
  --help-all            Displays help, including generic Qt options.
  -v, --version         Displays version information.
  -n, --no-splash       Do not show splash screen.
  -d, --debug           Print debug output to console.
  -f, --logfile         Print debug output to logfile.
  -c, --config <file>   File with QMapShack configuration.
  -l, --locale <code>   Application locale.
  --font-family <name>  Application font family.
  --font-size <size>    Application font size.
  --style <name>        Qt style.
                        Available: windows11, windowsvista, Windows, Fusion
                        Recommended: Fusion

Arguments:
  files                 Files for future use.
```

But I doubt that many users will call and check QMS help. Therefore, a notice in Wiki will be indispensable.

kiozen:
I will merge it now. There are probably more glitches. But most should be gone.

Issue #1169, opened 2026-08-01
Map drawing area normal/maximized/fullscreen viewport problem (Windows only!)


JFritzle:
### Describe the bug
[comment]: # (A clear and concise description of what the bug is)

To revive closed but never solved issue #251:

Wenn switching QMS window's size normal → maximized → fullscreen → maximized,
map drawing area viewport is **mostly** falsely shrinked to map viewport of QMS window's "normal" size before and becomes to small,
while overlay's viewport is correctly shrinked to  map/overlay viewport of QMS window's "maximized" size.

**Mostly** two problems arise:
1. Map drawing area does no longer cover whole canvas
2. Map elements and overlay elements no longer match

### What have you done to circle down the problem?
[comment]: # (Did you compare to previous versions? What versions? Did you check if others have the same problem?)

When saying **mostly**, it sometimes happens that map drawing area viewport is not shrinked to map viewport of QMS window's "normal" size before but correctly shrinked to map viewport of QMS window's "maximized" size before. Timing problem when querying map viewport's size?

### To Reproduce
[comment]: # (List the steps to reproduce the behavior)

See procedure described above and screenshosts below

### Expected behavior
[comment]: # (A clear and concise description of what you expected to happen)

I expect map drawing area viewport always to be resized back to "maximized" viewport after switching QMS window from "fullscreen" to "maximized".

### Screenshots
[comment]: # (If applicable, add screenshots to help explain your problem)

Screenshot 1:
QMS window has "normal" size
Map drawing area viewport covers whole available canvas size
Track on map has a highlighted overlay matching track on map
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/5eced4a8-c9dd-4256-b544-5f1420af927f" />

Screenshot 2:
After maximizing QMS window by "maximize" button at upper right corner, QMS has "maximized" size
Map drawing area viewport covers available canvas size
Track on map has a highlighted overlay still matching track on map
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/59d39c32-0b61-4b28-8684-d901c052c154" />

Screenshot 3:
After switching QMS window to "fullscreen" by F11 key, QMS has "fullscreen" size
Map drawing area viewport covers whole available canvas size
Track on map has a highlighted overlay still matching track on map
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/a527bebd-a4db-4070-b16c-76d6a118710b" />

[...]

PR #1203, opened 2026-08-14
[QMS-1169] Fix: Map viewport keeps the previous size after leaving fu…

kiozen:

QMS-#1169

### What you have done:

Bug: when a draw thread blocked a resize, CCanvas reposted the resize event with the size captured at that moment. That stale event lands after the real final geometry, so the draw contexts keep an intermediate size — and the partial per-context application let map and GIS overlay diverge.

Fix: single slotUpdateDrawContextViewport() that reads the current size/pixel ratio at apply time, applies all-or-nothing, retries via a 50 ms timer and each context's finished(); paintEvent() reconciles contexts against the canvas; GUI-thread wait(100) dropped; print() now waits for the threads on both size changes.

See ticket. 

Also test printing an area of the map.

### Does the code comply to the coding rules and naming conventions [Coding Guideline](https://github.com/Maproom/qmapshack/blob/dev/README_CODING.md):

- [x] yes

- [x] yes

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -487,6 +487,26 @@ divides the scale by `pixelRatio`.
 Test HiDPI paths on a normal screen with `QT_SCALE_FACTOR=2 build/bin/qmapshack`, adding
 `QT_SCALE_FACTOR_ROUNDING_POLICY=PassThrough` for fractional factors like 1.5.
 
+### Applying a new viewport size
+
+A draw thread holds a reference to its buffer with the mutex unlocked, so the buffers cannot be
+rebuilt while it runs and a resize has to be deferred. `CCanvas::slotUpdateDrawContextViewport()` is
+the only path that applies size and pixel ratio, and three rules keep it honest:
+
+- **Never pass a size captured earlier.** It reads `size()`/`devicePixelRatio()` at apply time. A
+  retry runs long after the event that scheduled it, and a queued resize event for an intermediate
+  geometry is delivered *after* the synchronously sent event for the final one — the Windows
+  fullscreen → maximized sequence does exactly that.
+- **All or nothing** (`canResize()` before `resize()`). A partially applied resize leaves the layers
+  with different viewports, which shows as map content no longer matching the GIS overlay.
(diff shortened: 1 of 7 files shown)


kiozen:
@JFritzle please test

JFritzle:
Switching from "maximized" to "fullscreen" and back to "maximized" by F11 key, map viewport now returns reliable to "maximized" viewport. 👍 

But main window is still shortly flickering to "normal" size when switching from "fullscreen" back to "maximized" by F11 key.
This is quite strange:
1. Resize QMS in "normal" mode, i.e. Qt::WindowNoState
2. Close QMS -> Window geometry and docks get saved
3. Open QMS -> Window geometry and docks get restored to "normal" state of step 1
4. Maximize QMS by "maximize" button on the upper right window border
5. Close QMS
6. Open QMS -> Window geometry and docks get restored to "maximized" state of step 4
7. Hit F11 -> QMS enters "fullscreen" state
8. Hit F11 again -> QMS leaves "fullscreen" state and enters shortly to "normal" state of step 1 before entering "maximized" state of step 4, now at least restoring correct viewport of step 4

I wonder about intermediate "normal" state of step 8 as within QMS session started with step 6 "normal" state has never been requested. It seems, window geometry and docks stored at step 2 gets shortly conjured out of the hatbox.
Perhaps saving of 2 sets of window geometry and docks may be necessary,
set "normal" keeping settings of "normal" state and set "maximized" keeping settings of "maximized" state
and after F11 key press to restore  the one corresponding to state preceeding "fullscreen" state?
Might in addition to functions `displayRegular()` and `displayFullscreen()` a third function `displayMaximized()` help which restores  saved "maximized" settings at transition from "fullscreen" to "maximized"?

I can live with solution of this PR but not flickering to "normal" state in between would be perfect.

kiozen:
I made a small fix. I am in doubt this will fix it but it's not the wrong thing to do either. If it does not fix it you have to tackle the issue on your PC.

Anyway, I will merge.

JFritzle:
> I am in doubt this will.

It will not.

